PR #14, opened 2016-04-15
remove temporary install dir

antoine-de:
else the binary could not be completly cleaned

---
This change is [Reviewable](https://reviewable.io/reviews/canaltp/mimirsbrunn/14)

### DIFF
--- a/build_packages.sh
+++ b/build_packages.sh
@@ -7,6 +7,9 @@ raw_version="`git describe`"
 # for debian version number, we don't want the leading 'v' of the git tag
 version=${raw_version#v}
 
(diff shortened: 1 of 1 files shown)


build_packages.sh
```diff
@@ -7,6 +7,9 @@ raw_version="`git describe`"
 # for debian version number, we don't want the leading 'v' of the git tag
 version=${raw_version#v}
 
+if [ -d $temporary_install_dir ]; then
+rm -rf $temporary_install_dir
```
TeXitoi: indentation

PR #27, opened 2016-05-04
add test on boundaries

kinnou02:


src/boundaries.rs
```diff
@@ -0,0 +1,337 @@
...
+}
+
+#[test]
+fn test_build_bounadry_not_closed() {
+    let mut builder = osm_builder::OsmBuilder::new();
+    let rel_id: osmpbfreader::OsmId;
+    {
+        let mut rel = builder.relation();
+        rel_id = rel.relation_id;
+        rel.outer(vec![(mimir::Coord {
```
antoine-de: I think it would be nice to have a builder for the tuple like:

`named_coord("start', 3.4, 5.2)` to add syntaxic sugar

PR #30, opened 2016-05-13
link the api to some params

antoine-de:
not proud of the moves....

I put the API in a struct because I though it would be more convenient in the futur.

same, wait a bit to merge, cannot test it well here

### DIFF
--- a/libs/skojig/src/query.rs
+++ b/libs/skojig/src/query.rs
@@ -35,7 +35,7 @@ use rs_es::operations::search::SearchResult;
 use mimir;
 use serde_json;
 
-fn build_rs_client(cnx: String) -> rs_es::Client {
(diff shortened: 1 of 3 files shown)


libs/skojig/src/query.rs
```diff
@@ -34,17 +34,17 @@ use rs_es::query::Query as rs_q;
 use rs_es::operations::search::SearchResult;
 use mimir;
 
-fn build_rs_client(args: &Args) -> rs_es::Client {
+fn build_rs_client(cnx: String) -> rs_es::Client {
```
kinnou02: why not taking a ```&String```?

TeXitoi: then better a `&str`

PR #43, opened 2016-05-27
refacto of the bulk_insert method

kinnou02:


libs/mimir/src/rubber.rs
```diff
@@ -240,30 +240,25 @@ impl Rubber {
...
-            chunk.clear();
-            let addr = match iter.next() {
+            //we remove an item for checking if the iterator is empty
+            let action = match actions.next() {
                 Some(a) => a,
                 None => break,
             };
-            //create the action and apply with_id to it if es_id returned Something
-            let action = addr.es_id().into_iter().fold(Action::index(addr), |action, id| action.with_id(id));
+            let mut chunk = actions.by_ref().take(1000).collect::<Vec<_>>();
```
antoine-de: Another way to implement this would be not to peek at a first action to get a stop condition, but check if chunk.len() is < to 1000 ad then stop

kinnou02:
Done

PR #48, opened 2016-06-01
add version and es url to /status

kinnou02:
also better dockerfile, we gain 10mo...


antoine-de:
it would be great to also add a `--version` to all executables

PR #52, opened 2016-06-08
bragi: A second search with match = name.ngram if no solution

kadhikari:


libs/bragi/src/query.rs
```diff
@@ -125,6 +126,13 @@ pub fn autocomplete(q: String,
     if let Some(ref coord) = coord {
         query_location(&q, coord)
     } else {
-        query(&q, cnx)
+    	//First search with match = "name.prefix".
+    	//If no result then another search with match = "name.ngram"  
+    	let results = try!(query(&q, cnx, "name.prefix"));
```
antoine-de: I think it's better to encapsulate a bit more and call there a `query_prefix(...)` and `query_ngram(...)` no ?

kadhikari:
@antoine-de 
I modified on your comment

PR #79, opened 2016-06-29
mimir: use of CoordWrapper instead of Coord in Street and Addr

kadhikari:


libs/osm_builder/src/lib.rs
```diff
@@ -107,14 +107,14 @@ impl OsmBuilder {
...
-    pub fn node(&mut self, coord: mimir::Coord, name: Option<String>) -> osmpbfreader::OsmId {
+    pub fn node(&mut self, coord: mimir::CoordWrapper, name: Option<String>) -> osmpbfreader::OsmId {
         if let Some(ref value) = name.clone().and_then(|n| self.named_nodes.get(&n)){
             return *value.clone()
         }
         let n = osmpbfreader::Node {
             id: self.node_id,
-            lat: coord.lat,
-            lon: coord.lon,
+            lat: coord.x,
```
TeXitoi: I'd prefer `lat: coord.lat()`

libs/osm_builder/src/lib.rs
```diff
@@ -107,14 +107,14 @@ impl OsmBuilder {
...
+            lon: coord.y,
```
TeXitoi: same

libs/osm_builder/src/lib.rs
```diff
@@ -33,8 +33,8 @@ extern crate osmpbfreader;
 extern crate mimir;
 use std::collections::BTreeMap;
 
-pub fn named_node(lat: f64, lon: f64, name: &'static str) -> (mimir::Coord, Option<String>) {
-    (mimir::Coord {lon: lon, lat: lat}, Some(name.to_string()))
+pub fn named_node(lat: f64, lon: f64, name: &'static str) -> (mimir::CoordWrapper, Option<String>) {
```
antoine-de: maybe we can name it Coord ?

libs/osm_builder/src/lib.rs
```diff
@@ -113,8 +113,8 @@ impl OsmBuilder {
         }
         let n = osmpbfreader::Node {
             id: self.node_id,
-            lat: coord.lat,
-            lon: coord.lon,
+            lat: coord.x,
+            lon: coord.y,
```
TeXitoi: writing a Coord::lat() and Coord::lon() may be a good idea.

src/bin/osm2mimir.rs
```diff
@@ -167,7 +167,7 @@ fn administrative_regions(pbf: &mut OsmPbfReader, levels: BTreeSet<u32>) -> Admi
                                            objects.get(&r.member).and_then(|value| {
                                                match value {
                                                    &osmpbfreader::OsmObj::Node(ref node) => {
-                                                       Some(mimir::CoordWrapper(geo::Coordinate {
+                                                       Some(mimir::Coord(geo::Coordinate {
```
TeXitoi: better to use Coord::new

PR #82, opened 2016-09-22
add the source of osm fixture

nlehuby:


TeXitoi:
What's the point of having .osm and .osm.pbf? can't we generate automatically the pbf according to the osm. We may have unsyncronized files if we keep the 2 in the repository.

prhod:
I need the OSM file to add some POIs. It can eventually replace the PBF file.

pbougue:
I vote for one file only... so we may add the pbf generation somewhere? Not in control on that subject...

kinnou02:
I don't like the idea to run osmosis during the build/test phase

pbougue:
maybe add a command on build by travis to run osmosis and warn if diffs in pbf?

kinnou02:
We will still require osmosis for our build system

pbougue:
Ok, so we just merge that as is, and hope every update is well done?
Then add so doc on how to manage osm fixtures would be nice.
Fine by me.

nlehuby:
Doc added.

tests/fixtures/readme.md
```diff
@@ -0,0 +1,21 @@
+## OSM fixtures
+
+The sample OSM files to use are the *.osm.pbf ones.
+
+For now, they contain a really small subset of real OSM data.
+
+To add data to the OSM fixture :
```
pbougue: It would be nice to have a preventive first step checking that when generating pbf from unmodified osm it doesn't change a thing (otherwise go break someone's knee :) )

pbougue: Binary + different implem possible, so not doable

PR #88, opened 2016-11-16
Added postcode to POI

datanel:


datanel:
When I load the fixtures 
```

pbougue:
I guess this has to do with the ES model : https://github.com/CanalTP/mimirsbrunn/blob/master/json/settings.json#L211
Maybe say it should be indexed, have a look at what's done on other zip_codes or other fields (I guess we want exact match).

PR #91, opened 2016-11-17
update osm tests and readme

prhod:
osmconvert has a bug, the name tag is not included in the output pbf file.


tests/fixtures/readme.md
```diff
@@ -10,6 +10,9 @@ To add data to the OSM fixture :
 * save the new source file (File > Save As...)
 * transform the *.osm file into an *.osm.pbf file using osmconvert or osmosis
 
+Example of osmosis command line :
```
pbougue: you may add that it is ricky to use osmconvert

PR #93, opened 2016-11-21
apply cargo fmt

datanel:


datanel:
cargo fmt applies modifications on src/ not libs.
I have to go to libs/bragi.... and aply cargo fmt.

what's why do_not_merge

PR #98, opened 2016-11-23
adding a complex poi and renaming osm.pbf file name

prhod:
The BANO csv file keep the old name because there is still 3 cities in it.


TeXitoi:
Did you remove the old  tests/fixtures/three_cities.osm ?

datanel:
@TeXitoi there is no three_cities.osm

PR #99, opened 2016-11-23
Optimization import pois

azime:


src/bin/osm2mimir.rs
```diff
@@ -480,32 +483,34 @@ fn main() {
...
-    }
-
-    if args.flag_import_way {
-        info!("importing streets into Mimir");
-        let nb_streets = rubber.index("way", &args.flag_dataset, streets.into_iter())
-            .unwrap();
-        info!("Nb of indexed street: {}", nb_streets);
-    }
-
+	}
```
TeXitoi: badly indented block.

PR #100, opened 2016-11-24
import pois optimization

azime:


src/bin/osm2mimir.rs
```diff
@@ -98,32 +100,28 @@ fn main() {
...
-    }
-
+    
+        if args.flag_import_way {
+            info!("importing streets into Mimir");
+            let nb_streets = rubber.index("way", &args.flag_dataset, streets.into_iter())
+                .unwrap();
+            info!("Nb of indexed street: {}", nb_streets);
+        }
+	}
```
TeXitoi: indentation problems.

src/bin/osm2mimir.rs
```diff
@@ -98,32 +100,28 @@ fn main() {
...
+        info!("Extracting streets from osm");
+        let mut streets = streets(&mut parsed_pbf, &admins_geofinder, city_level);
+
+    	info!("computing city, street weight");
+        for st in &mut streets {
+            for admin in &mut st.administrative_regions {
+                admin.weight.set(admin.weight.get() + 1);
+                if admin.level == city_level {
+                    st.weight = admin.weight.get();
+                }
```
TeXitoi: That will not work. You must first compute the admin weight, and then the street one.

Imagine there is streets A and B in admin C, with your implementation, A will have weight 1 and B 2. But the real solution is A and B have weight 2.

src/osm_reader/poi.rs
```diff
@@ -99,12 +98,16 @@ fn parse_poi(osmobj: &osmpbfreader::OsmObj,
 
     let name = osmobj.tags().get("name").map_or("", |name| name);
     let adms = admins_geofinder.get(&coord);
+    let zip_codes = match osmobj.tags().get("addr:postcode") {
+		Some(val) if ! val.is_empty() => vec![val.clone()],
+		_ => get_zip_codes_from_admins(&adms) 
+    };
```
TeXitoi: You can also use `if let Some(val) = ...` for such cases.

PR #104, opened 2016-11-25
Ignore poi without coord name

azime:


src/osm_reader/poi.rs
```diff
@@ -97,20 +97,31 @@ fn parse_poi(osmobj: &osmpbfreader::OsmObj,
     };
 
     let name = osmobj.tags().get("name").map_or("", |name| name);
+
+    if coord.is_default() {
+        warn!("The poi {} is rejected, cause: without coordinates.", id);
+        return None;
+    }
+
+    if name == "" {
```
antoine-de: `is_empty` seems a bit more explicit

src/osm_reader/poi.rs
```diff
@@ -127,6 +138,8 @@ pub fn pois(pbf: &mut OsmPbfReader,
     objects.iter()
         .filter(|&(_, obj)| matcher.is_poi(&obj))
         .map(|(_, obj)| parse_poi(obj, &objects, admins_geofinder, city_level))
+        .filter(|o| o.is_some())
```
antoine-de: `filter_map` no ?

tests/bragi_test.rs
```diff
@@ -376,14 +376,21 @@ pub fn bragi_tests(es_wrapper: ::ElasticSearchWrapper) {
     let geocodings = get_results(bragi_get("/autocomplete?q=77000 Hôtel de Ville (Melun)"));
     let types = get_types(&geocodings);
     assert_eq!(count_types(&types, "poi"), 1);
-}
 
-fn get_labels<'a>(r: &'a Vec<BTreeMap<String, serde_json::Value>>) -> Vec<&'a str> {
-    r.iter().map(|e| e.get("label").and_then(|l| l.as_str()).unwrap_or("")).collect()
+    //ID : poi:osm:node:663715752, Label : (Saint-Martin-d'Hères) name == ""
```
antoine-de: your comment is clear but I think only within this PR's context.

It would be better to be a bit more explicit:
`we search for a POI (id = ...) with a label but an empty name, it should be filtered)`

tests/bragi_test.rs
```diff
@@ -376,14 +376,21 @@ pub fn bragi_tests(es_wrapper: ::ElasticSearchWrapper) {
...
+    let geocodings = get_results(bragi_get("/autocomplete?q=(Saint-Martin-d'Hères)"));
+    let types = get_types(&geocodings);
+    assert_eq!(count_types(&types, "poi"), 0);
+
+    //ID : poi:osm:relation:2561223, Label : ENSE3 site Ampère, coord empty
```
antoine-de: same here

PR #105, opened 2016-12-02
upgrade rust-geo

azime:


libs/bragi/src/model.in.rs
```diff
@@ -88,7 +88,7 @@ impl ToGeom for mimir::Place {
     }
 }
 
-impl ToGeom for geo::Coordinate {
+impl ToGeom for geo::Coordinate<f64> {
```
antoine-de: isn't 32 enough for us ?

what do you think @TeXitoi , @kinnou02 ?

antoine-de: I think it's 2m at most

azime: osm data used : IDF
data loaded : poi, street, admins
Difference between f64 and f32 : ~3Mo

PR #107, opened 2016-12-06
mimirsbrunn: set empty poi name with type

patochectp:


src/osm_reader/poi.rs
```diff
@@ -96,19 +96,14 @@ fn parse_poi(osmobj: &osmpbfreader::OsmObj,
         }
     };
 
-    let name = osmobj.tags().get("name").map_or("", |name| name);
+    let name = osmobj.tags().get("name").map_or(osmobj.tags().get("amenity").unwrap(), |name| name );
```
TeXitoi: You don't want a hard unwrap. imagine a poi with `tags =[("leisure", "park")]`, kaboom.

patochectp: done

src/osm_reader/poi.rs
```diff
@@ -96,23 +96,28 @@ fn parse_poi(osmobj: &osmpbfreader::OsmObj,
...
     if coord.is_default() {
         info!("The poi {} is rejected, cause: could not compute coordinates.",
               id);
         return None;
     }
 
+    if name.is_none() {
+        info!("The poi {} is rejected, cause: could not compute name.", id);
+        return None;
+    }
```
TeXitoi:
```rust
let name = if let Some(s) = name {
    s
} else {
    info!("The poi {} is rejected, cause: could not compute name.", id);
    return None;
};
```
and remove the `unwrap()` after.

src/osm_reader/poi.rs
```diff
@@ -96,15 +96,15 @@ fn parse_poi(osmobj: &osmpbfreader::OsmObj,
         }
     };
 
-    let name = osmobj.tags().get("name").map_or("", |name| name);
+    let name = osmobj.tags().get("name").map_or(osmobj.tags().get("amenity"), |name| Some(name) );
```
TeXitoi: extra space at the end make rust-fmt test fail.

patochectp:
done

tests/bragi_test.rs
```diff
@@ -382,10 +382,10 @@ pub fn bragi_tests(es_wrapper: ::ElasticSearchWrapper) {
     let types = get_types(&geocodings);
     assert_eq!(count_types(&types, "poi"), 1);
 
-    // we search for a POI (id = 663715752) with a label but an empty name, it should be filtered)
-    let geocodings = get_results(bragi_get("/autocomplete?q=(Saint-Martin-d'Hères)"));
+    // we search for POI with a label but an empty name set with type)
```
antoine-de: proposal for a clearer comment:

// we search for POIs with a type but an empty name, we should have set the name with the type.
// for exemple there are parkings without name (but with the tag "anemity" = "Parking"), we should be able to query them

PR #109, opened 2016-12-21
Import stops

azime:


README.md
```diff
@@ -135,6 +135,20 @@ To import all those data into Mimir, you only have to do:
 
 The `--connection-string` argument refers to the ElasticSearch url
 
+
+### stops2mimir
+
+This component imports stops into Mimir.
+It is recommanded to run stops integration after osm integration so that stops are attached to admins.
```
antoine-de: maybe you can add a comment to tell that the stop.txt is the NTFS stop file (with a ntfs reference ?)

json/settings.json
```diff
@@ -227,6 +227,76 @@
...
+                },
+                "weight": {
+                  "type": "long"
+                }
+              }
+            },
+            "weight": {
+              "type": "long"
+            },
+            "zip_codes": {
```
antoine-de: we have 2 zip_codes ?

src/bin/stops2mimir.rs
```diff
@@ -0,0 +1,224 @@
...
+                None => Err(csv::Error::Decode(format!("Failed accessing record '{}'.", pos))),
+            }
+        }
+        fn parse_f64(s: &str) -> csv::Result<f64> {
+            s.parse()
+                .map_err(|_| csv::Error::Decode(format!("Failed converting '{}' from str.", s)))
+        }
+        fn is_valid_stop_area(location_type: &Option<u8>,
+                              visible: &Option<u8>)
+                              -> csv::Result<bool> {
```
antoine-de: you can return a Result<()>

src/bin/stops2mimir.rs
```diff
@@ -0,0 +1,224 @@
...
+
+impl<'a, R: std::io::Read + 'a> StopPointIter<'a, R> {
+    fn new(r: &'a mut csv::Reader<R>) -> Option<Self> {
+        let headers = if let Ok(hs) = r.headers() {
+            hs
+        } else {
+            return None;
+        };
+        let get = |name| headers.iter().position(|s| s == name);
+        let stop_id_pos = if let Some(pos) = get("stop_id") {
```
antoine-de: can't we do a function for this ?

src/bin/stops2mimir.rs
```diff
@@ -0,0 +1,224 @@
...
+            pos
+        } else {
+            return None;
```
antoine-de: wouldn't that make a bad error message ?

src/bin/stops2mimir.rs
```diff
@@ -0,0 +1,224 @@
...
+Options:
+    -h, --help              \
+     Show this message.
+    -i, --input=<file>      NTFS stops.txt file.
+    -c, \
+     --connection-string=<connection-string>
+                            Elasticsearch \
+     parameters, [default: http://localhost:9200/munin]
+    -d, --dataset=<dataset>
+";
```
TeXitoi: rust-fmt killed the message :(

src/bin/stops2mimir.rs
TeXitoi: Maybe something as
```rust
let get = |name| headers.iter().position(|s| s == name).ok_or(format!("no {} column", name));
let stop_id_pos = get("stop_id")?;
```
(and returning `Result<Self, Box<std::error::Error>>`) It'll do the trick for cheap about the error message.

src/bin/stops2mimir.rs
```diff
@@ -0,0 +1,224 @@
...
+}
+
+impl<'a, R: std::io::Read + 'a> Iterator for StopPointIter<'a, R> {
+    type Item = csv::Result<mimir::Stop>;
+    fn next(&mut self) -> Option<Self::Item> {
+        fn get(record: &[String], pos: usize) -> csv::Result<&str> {
+            match record.get(pos) {
+                Some(s) => Ok(&s),
+                None => Err(csv::Error::Decode(format!("Failed accessing record '{}'.", pos))),
+            }
```
TeXitoi:
```rust
record.get(pos)
    .map(|s| &s)
    .ok_or_else(|| csv::Error::Decode(format!("Failed accessing record '{}'.", pos)))
```

src/bin/stops2mimir.rs
```diff
@@ -0,0 +1,224 @@
...
+                              -> csv::Result<bool> {
+            if (*location_type == Some(1)) && (*visible != Some(0)) {
+                Ok(true)
+            } else {
+                Err(csv::Error::Decode("Not stop_area.".to_string()))
+            }
+        }
+        self.iter.next().map(|r| {
+            r.and_then(|r| {
+                let stop_id = try!(get(&r, self.stop_id_pos));
```
TeXitoi: if you prefer: https://blog.rust-lang.org/2016/11/10/Rust-1.13.html#the--operator

src/bin/stops2mimir.rs
```diff
@@ -0,0 +1,205 @@
...
+        let get_optional_pos = |name| headers.iter().position(|s| s == name);
+
+        let get_pos = |field| {
+            get_optional_pos(field)
+                .ok_or(csv::Error::Decode(format!("Invalid file, cannot find column '{}'", field)))
+        };
+
+        Ok(StopPointIter {
+            iter: r.records(),
+            stop_id_pos: try!(get_pos("stop_id")),
```
TeXitoi: if you prefer: https://blog.rust-lang.org/2016/11/10/Rust-1.13.html#the--operator

src/bin/stops2mimir.rs
```diff
@@ -0,0 +1,205 @@
...
+        .unwrap_or_else(|e| e.exit());
+
+    info!("creation of indexes");
+    let mut rubber = Rubber::new(&args.flag_connection_string);
+    let mut rdr = csv::Reader::from_file(args.flag_input)
+        .unwrap()
+        .double_quote(true);
+
+    let stops: Vec<mimir::Stop> = StopPointIter::new(&mut rdr)
+        .expect("Can't find needed fields in the header.")
```
TeXitoi: unwrap will be better here as it will explain the error.

src/bin/stops2mimir.rs
antoine-de: yes I wondered for this one since there is only one call to `try!`

do we want to always use `?` ?

src/bin/stops2mimir.rs
antoine-de: yes, missed this with the refacto, thanks!

src/bin/stops2mimir.rs
TeXitoi: We can stay as is for the moment.

antoine-de:
@TeXitoi changes pushed

PR #112, opened 2016-12-23
stops2mimir: Add 'stop_area:' to stop id

pbougue:


src/bin/stops2mimir.rs
```diff
@@ -131,7 +131,7 @@ impl<'a, R: std::io::Read + 'a> Iterator for StopPointIter<'a, R> {
                 let stop_lon = try!(parse_f64(stop_lon));
                 let stop_name = try!(get(&r, self.stop_name_pos));
                 Ok(mimir::Stop {
-                    id: stop_id.to_string(),
+                    id: format!("stop_area:{}", stop_id),
```
antoine-de: we need to add a comment to explain why we are doing this

PR #113, opened 2016-12-23
better error handling

antoine-de:
less logs and better handling of the stop_area and visible flags

### DIFF
--- a/src/bin/stops2mimir.rs
+++ b/src/bin/stops2mimir.rs
@@ -112,7 +112,7 @@ impl<'a, R: std::io::Read + 'a> Iterator for StopPointIter<'a, R> {
         }
 
         fn is_valid_stop_area(location_type: &Option<u8>, visible: &Option<u8>) -> csv::Result<()> {
-            if *location_type == Some(0) {
+            if *location_type != Some(1) {
                 Err(csv::Error::Decode("not a stop_area.".to_string()))
             } else if *visible == Some(0) {
                 Err(csv::Error::Decode("stop_area invisible.".to_string()))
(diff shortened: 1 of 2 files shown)


src/bin/stops2mimir.rs
```diff
@@ -112,10 +112,12 @@ impl<'a, R: std::io::Read + 'a> Iterator for StopPointIter<'a, R> {
         }
 
         fn is_valid_stop_area(location_type: &Option<u8>, visible: &Option<u8>) -> csv::Result<()> {
-            if *location_type == Some(1) && *visible != Some(0) {
-                Ok(())
+            if *location_type == Some(0) {
+                Err(csv::Error::Decode("not a stop_area.".to_string()))
```
TeXitoi: No, we want it equal to 1 to be a stop area.

tests/fixtures/stops.txt
```diff
@@ -1,9 +1,9 @@
 stop_id,visible,stop_name,stop_lat,stop_lon,zone_id,location_type,parent_station,stop_timezone,equipment_id,contributor_id,geometry_id,frame_id
 SA:main_station,1,"République",47.099758,2.491858,,1,,Europe/Paris,,BGT,,BGT:19
-SA:second_station,,"14 Juillet",47.123966,2.402522,,1,,Europe/Paris,,BGT,,BGT:19
+SA:second_station,,"14 Juillet",47.123966,2.402522,,,,Europe/Paris,,BGT,,BGT:19
```
TeXitoi: That's not what you want.

src/bin/stops2mimir.rs
antoine-de: since it's an optional field in GTFS we'll thus only have stop_points when we want to have stop areas no ?

antoine-de: ah yes sorry misread the doc, in this case we want all to be stoppoints, I'll change this

PR #117, opened 2017-01-09
Update osmpbfreader to 0.7.0 + style arround osmpbfreader

TeXitoi:


src/osm_reader/street.rs
```diff
@@ -155,14 +153,14 @@ pub fn streets(pbf: &mut OsmPbfReader,
...
-   	                    id: way.id.to_string(),
-   	                    street_name: way_name.to_string(),
-   	                    label: format_label(&admins, city_level, way_name),
-   	                    weight: 1,
-   	                    zip_codes: get_zip_codes_from_admins(&admins),
-   	                    administrative_regions: admins,
-   	                    coord: get_way_coord(objs_map, way),
+            ret ret(street_list.push(mimir::Street {
+   	        id: way.id.0.to_string(),
+                street_name: way_name.to_string(),
```
antoine-de: bad ident

src/osm_reader/admin.rs
```diff
@@ -100,21 +100,14 @@ pub fn administrative_regions(pbf: &mut OsmPbfReader, levels: BTreeSet<u32>) ->
...
-                        match value {
-                            &osmpbfreader::OsmObj::Node(ref node) => {
-                                Some(mimir::Coord::new(node.lat(), node.lon()))
-                            }
-                            _ => None,
-                        }
-                    })
-                });
+                .and_then(|r| objects.get(&r.member))
+                .and_then(osmpbfreader::OsmObj::node)
```
antoine-de: it's the same than
`and_then(|o| o.node())` no ? I think the latter is a bit clearer.

TeXitoi: That's the same, yes.

PR #119, opened 2017-01-11
add the possibility to trace the time in log

antoine-de:
you can now trace the timestamp in log with the env var `LOG_TIME`

### DIFF
--- a/libs/mimir/src/lib.rs
+++ b/libs/mimir/src/lib.rs
@@ -54,19 +54,14 @@ use chrono::Local;
 pub fn logger_init() -> Result<(), log::SetLoggerError> {
     let mut builder = env_logger::LogBuilder::new();
 
-    if let Ok(s) = std::env::var("LOG_TIME") {
-        if s == "1" {
-            builder.format(|record| {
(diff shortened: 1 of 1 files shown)


TeXitoi:
Why not just doing that by default?

antoine-de:
it adds verbosity to the log, but I'm ok with including this by default

antoine-de:
done

libs/mimir/src/lib.rs
```diff
@@ -48,12 +48,20 @@ extern crate geojson;
...
-    builder.filter(None, log::LogLevelFilter::Info);
+
+    builder.filter(None, log::LogLevelFilter::Info)
+        .format(|record| {
+            format!("[{time}]{lvl}:{loc}: {msg}",
+                    time = Local::now(),
+                    lvl = record.level(),
+                    loc = record.location().module_path(),
+                    msg = record.args())
+        });;
```
TeXitoi: double `;`?

antoine-de: oups mispaste :confused:

PR #120, opened 2017-01-16
bragi: add citycode on street, addr, poi and stop

patochectp:


libs/bragi/src/model.in.rs
```diff
@@ -66,7 +66,7 @@ pub struct GeocodingResponse {
     pub street: Option<String>,
     pub postcode: Option<String>,
     pub city: Option<String>,
-    pub city_code: Option<String>,
```
antoine-de: hum, we unfortunately can't rename a variable like this because bragi would stop working during deployment.

you need to add another field, add a comment on this one saying it's deprecated and after the next release, we'll be able to remove it

libs/bragi/src/model.in.rs
```diff
@@ -155,6 +155,13 @@ fn get_city_name(admins: &Vec<Rc<mimir::Admin>>) -> Option<String> {
         .map(|admin| admin.label.clone())
 }
 
+fn get_city_code(admins: &Vec<Rc<mimir::Admin>>) -> Option<String> {
+    admins.iter()
+        .find(|a| !a.zip_codes.is_empty())
+        .or_else(|| admins.iter().next())
+        .map(|admin| admin.insee.clone())
```
antoine-de: it's the insee, not the postcode ?

patochectp: In the query http://api-adresse.data.gouv.fr/search?q=Paris%2075 citycode is fill with insee

antoine-de: ok

libs/bragi/src/model.in.rs
patochectp: done

libs/bragi/src/model.in.rs
```diff
@@ -155,6 +157,13 @@ fn get_city_name(admins: &Vec<Rc<mimir::Admin>>) -> Option<String> {
         .map(|admin| admin.label.clone())
 }
 
+fn get_city_code(admins: &Vec<Rc<mimir::Admin>>) -> Option<String> {
```
pbougue: minor: I'd name it `get_citycode`

antoine-de:
rust fmt is not valid but it's because the latest rust fmt version is not valid, we merge this PR anyway

https://github.com/rust-lang-nursery/rustfmt/issues/1279

PR #123, opened 2017-01-20
relax the minimum match filter

antoine-de:
when we search [0, 2] fields, we need to match 100% of them
when we search [0, 4] fields, we need to match 90% of them
when we [5, +] fields, we only need to match 70% on them 

this way if the has more fields than the documents (additional information like country, the region, ...) they can be ignored

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-minimum-should-match.html

we this on go from 938 fails on geocoding tester (on idf) to 881

Note: the impact on performance seems negligeable

### DIFF
--- a/libs/bragi/src/query.rs
+++ b/libs/bragi/src/query.rs
@@ -72,14 +72,25 @@ pub fn make_place(doc_type: String, value: Option<Box<serde_json::Value>>) -> Op
     })
 }
 
+enum MatchType {
+    Prefix,
+    Fuzzy,
+}
+
 fn build_query(q: &str,
-               match_type: &str,
+               match_type: MatchType,
                coord: &Option<model::Coord>,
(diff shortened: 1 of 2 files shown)


TeXitoi:
It may be great to set this value depending of prefix/ngram request.

Nothing in common, but ngram request should also include prefix for better quality of results.

antoine-de:
you would lower the value for ngram ?

I fear this would add very strange results at the end (but since it's at the end it might not be very important)

For the second point, you want to add a prefix match and boost it ?
not a bad idea, I can test it, thanks!

TeXitoi:
Exactly for the 2 points.

For the first point, if you write chales de gaule, you have 3/6 corresponding ngrams, thus I think it's important to have something not too low (but maybe the ratio is not calculated like that).

antoine-de:
I pushed the modifications

PR #124, opened 2017-01-20
house number handling

antoine-de:
filter to handle house numbers

we either want:
* to exactly match the document house_number
* or that the document has no house_number

we were also matching the zip code in this, so it was not working

geocoding tester fails:
before 938
after 817


antoine-de:
I corrected the tests and added a way not to clean the test docker

`DONT_KILL_THE_WHALE=1 cargo test`
will leave the docker alive

after you can do
`BRAGI_ES=http://localhost:9242 ./target/debug/bragi`

 to get a bragi on the tests data

you will need to docker stop/rm mimirsbrun_test before the next tests

PR #127, opened 2017-01-30
Continuous deployment configuration

antoine-de:
here is the fist draft to use a jenkinsfile for mimirsbrunn

we'll need to rewrite all the jenkinsfile with the [declarative pipleline](https://github.com/jenkinsci/pipeline-model-definition-plugin) :scream:  ([here is the first attempt at this](http://jenkins.ci.canaltp.fr/job/test_mimir_declarative/)), but this can come in a second time


libs/docker_wrapper/src/lib.rs
```diff
@@ -38,27 +38,34 @@ use std::error::Error;
...
             return Err(format!("`docker run` failed {}", &status).into());
         }
 
+        // we need to get the ip of the container if the container has been run on another machine
+        let container_ip_cmd = try!(Command::new("docker")
+            .args(&["inspect", "--format={{.NetworkSettings.IPAddress}}", "mimirsbrunn_tests"])
+            .output());
+
+        let container_ip =
+            std::str::from_utf8(container_ip_cmd.stdout.as_slice())?.replace("\n", "");
```
TeXitoi:
```rust
std::str::from_utf8(container_ip_cmd.stdout.as_slice())?.trim();
```

PR #130, opened 2017-01-31
pt_dataset filter added

azime:


libs/bragi/src/query.rs
```diff
@@ -210,8 +240,11 @@ fn query(q: &str,
 
     let mut client = build_rs_client(&cnx.to_string());
 
+    let stops_index = make_stops_index(pt_dataset);
+    let indexes = make_indexes(&stops_index, &mut client);
```
antoine-de: wouldn't it be better if we don't get any pt_dataset to query on `munin` and else query on `stops_{dataset},munin_street_network` ?

libs/mimir/src/rubber.rs
```diff
@@ -166,6 +166,7 @@ impl Rubber {
                          dataset: &str,
                          index: String)
                          -> Result<(), String> {
+        let v = vec!["addr", "street", "poi", "admin", "way"];
```
antoine-de: hum, not convinced by this, wouldn't it be better to use a `is_street_network` trait trait in the mimir's type (a bit like the `doc_type`) ?

tests/bragi_stops_test.rs
```diff
@@ -106,3 +117,31 @@ fn stop_no_admin_test(bragi: &BragiHandler) {
     let admins = stop.get("administrative_regions").and_then(|a| a.as_array());
     assert_eq!(admins.map(|a| a.len()).unwrap_or(0), 0);
 }
+
+fn stop_filtred_by_dataset_test(bragi: &BragiHandler) {
```
pbougue: filt`e`red

tests/bragi_stops_test.rs
```diff
@@ -106,3 +117,31 @@ fn stop_no_admin_test(bragi: &BragiHandler) {
...
+    // All stops
```
antoine-de: can you add a bit of comment explaining more this ?

like there are 2 datasets imported, so we should be able to query stops from both datasets

tests/stops2mimir_test.rs
```diff
@@ -31,15 +31,15 @@
 extern crate serde_json;
 extern crate mimir;
 
-/// Simple call to a BANO load into ES base
+/// Simple call to a stops2mimir load into ES base
 /// Checks that we are able to find one object (a specific address)
 pub fn stops2mimir_sample_test(es_wrapper: ::ElasticSearchWrapper) {
-    let osm2mimir = concat!(env!("OUT_DIR"), "/../../../stops2mimir");
-    ::launch_and_assert(osm2mimir,
+    let stops2mimir = concat!(env!("OUT_DIR"), "/../../../stops2mimir");
```
antoine-de: :+1:

libs/bragi/src/query.rs
pbougue: might not be so long-term, as we will have proprietary stops... at least we will have to provide a "open" call (and the "full access" is optional to me)

libs/mimir/src/rubber.rs
```diff
@@ -175,7 +176,14 @@ impl Rubber {
         let type_index = get_main_type_index(doc_type);
         try!(self.alias(&type_index, &vec![dataset_index.clone()], &last_indexes));
 
-        try!(self.alias("munin", &vec![type_index.to_string()], &vec![]));
+        if v.contains(&doc_type) {
+            try!(self.alias("munin_street_network",
```
pbougue: Not fan of the name, I'd rather use `geo_data` (or open_geo_data) but street_network is redundant with fallback computing and confusing to me. (also not fan of an index on that, as we might have extra private geo_data for poi... even if it's not the goal)

tests/bragi_stops_test.rs
azime: OK

libs/bragi/src/query.rs
azime: @antoine-de OK

libs/bragi/src/query.rs
```diff
@@ -229,6 +262,7 @@ fn query(q: &str,
 }
 
 pub fn autocomplete(q: &str,
+                    pt_dataset: &str,
```
azime: OK

tests/bragi_stops_test.rs
azime: ok

libs/mimir/src/rubber.rs
azime: replace  street_network by geo_data

libs/bragi/src/query.rs
```diff
@@ -198,7 +198,37 @@ fn build_query(q: &str,
...
+            let mut result: Vec<&str> = vec![];
+            if is_existing_index(client, &dataset) {
+                result.push(dataset);
+            }
+            if is_existing_index(client, &"munin_geo_data".to_string()) {
+                result.push("munin_geo_data");
+            }
+            result
+        }
+        _ => vec!["munin"],
```
pbougue: more secure to call on "munin_geo_data" here as full_access may only be usefull for debug, and we don't want to do that by mistake for any lambda user

antoine-de: By thinking a bit more about it, I think you're right since with a default without stops we'll be able to deploy it to prod directly.

But if we do this we'll need a way to get all the stops for debug (a `_all_stops=true` param ?)

azime: _all_data=true

libs/bragi/src/query.rs
```diff
@@ -198,7 +198,40 @@ fn build_query(q: &str,
         .build()
 }
 
+fn is_existing_index(client: &mut rs_es::Client, index: &String) -> bool {
+    match client.open_index(index) {
```
antoine-de: since the log is not very important (I think we'll see the indexes we search on in the logs), you can rewrite this:
`client.open_index(index).is_ok()`

src/bin/bano2mimir.rs
```diff
@@ -129,7 +129,7 @@ fn index_bano<I>(cnx_string: &str, dataset: &str, files: I)
             Ok(nb) => info!("importing {:?}: {} addresses added.", &f, nb),
         }
     }
-    rubber.publish_index(doc_type, dataset, addr_index).unwrap();
+    rubber.publish_index(doc_type, dataset, addr_index, true).unwrap();
```
antoine-de: Hum In this I think it would be clearer to:
* not declare `let doc_type = "addr";` but use `Address::doc_type()`
* use `address::is_geo_data()`

src/bin/osm2mimir.rs
```diff
@@ -113,12 +113,12 @@ fn main() {
...
         if args.flag_import_way {
             info!("importing streets into Mimir");
-            let nb_streets = rubber.index("way", &args.flag_dataset, streets.into_iter())
+            let nb_streets = rubber.index(&args.flag_dataset, streets.into_iter())
                 .unwrap();
             info!("Nb of indexed street: {}", nb_streets);
         }
     }
-    let nb_admins = rubber.index("admin", &args.flag_dataset, admins.iter())
+    let nb_admins = rubber.index(&args.flag_dataset, admins.iter())
```
antoine-de: :+1:

libs/bragi/src/api.rs
```diff
@@ -147,6 +149,9 @@ impl ApiEndPoint {
                 let cnx = self.es_cnx_string.clone();
                 endpoint.handle(move |client, params| {
                     let q = params.find("q").and_then(|val| val.as_str()).unwrap_or("").to_string();
+                    let pt_dataset = params.find("pt_dataset")
+                        .and_then(|val| val.as_str());
+                    let all_data = params.find("_all_data").and_then(|val| val.as_bool());
```
antoine-de: I think it would be better to use a `bool` instead of a Option<bool>, just init it to false

libs/bragi/src/api.rs
```diff
@@ -224,8 +240,14 @@ impl ApiEndPoint {
                             lat: lat.unwrap(),
                         })
                     });
-                    let model_autocomplete =
-                        query::autocomplete(&q, offset, limit, coord, &cnx, None);
+                    let model_autocomplete = query::autocomplete(&q,
+                                                                 &pt_dataset,
+                                                                 &all_data,
```
antoine-de: I don't think you need to pass it as a ref

libs/bragi/src/query.rs
```diff
@@ -198,7 +198,40 @@ fn build_query(q: &str,
...
+        Ok(_) => true,
+        Err(_) => {
+            debug!("{} index not found", index);
+            false
+        }
+    }
+}
+
+fn make_indexes<'a>(all_data: &Option<bool>,
+                    pt_dataset_index: &'a String,
```
antoine-de: don't you want an Option<&'a String> here ?

libs/bragi/src/api.rs
TeXitoi: +1

libs/bragi/src/query.rs
azime: ok

libs/bragi/src/query.rs
TeXitoi: you never want a `&String`, you always want a `&str` instead

src/bin/bano2mimir.rs
azime: ok

libs/bragi/src/api.rs
azime: ok

libs/mimir/src/objects.in.rs
```diff
@@ -110,6 +110,10 @@ pub trait DocType {
     fn doc_type() -> &'static str; // provides the elasticsearch type name
 }
 
+pub trait IsGeoData {
+    fn is_geo_data() -> bool;
+}
```
TeXitoi: Maybe something like that is better?

```rust
pub trait MimirObject: serde::Serialize {
    fn doc_type() -> &'static str;
    fn is_geo_data() -> bool;
    fn es_id(&self) -> Option<String>;
}
```

libs/bragi/src/query.rs
azime: ok

libs/bragi/src/query.rs
```diff
@@ -199,39 +199,36 @@ fn build_query(q: &str,
...
     if is_existing_index(client, &"munin_geo_data".to_string()) {
         result.push("munin_geo_data");
     }
-    return result;
+    match *pt_dataset_index {
+        Some(ref dataset) => {
+            if is_existing_index(client, dataset) {
+                result.push(dataset);
+            }
+            result
```
antoine-de: hum I think it would be better for this return to be outside the match

libs/mimir/src/objects.in.rs
azime: OK

libs/bragi/src/query.rs
azime: ok

libs/bragi/src/query.rs
```diff
@@ -87,9 +88,9 @@ fn build_query(q: &str,
     // we order the type of object we want
     // Note: the addresses are boosted more because even if we don't want them first
     // because they are more severely filtered
-    let boost_addr = rs_q::build_term("_type", "addr").with_boost(5000).build();
-    let boost_admin = rs_q::build_term("_type", "admin").with_boost(3000).build();
-    let boost_stop = rs_q::build_term("_type", "stop").with_boost(2000).build();
+    let boost_addr = rs_q::build_term("_type", Addr::doc_type()).with_boost(5000).build();
```
antoine-de: :+1:

PR #131, opened 2017-02-01
add admin to stop label

antoine-de:
change stops label, add it's city in the label

`chatelet (Paris)`

like it is done for the other objects

also fill the postcode


tests/bragi_stops_test.rs
```diff
@@ -84,7 +84,6 @@ fn stop_attached_to_admin_test(bragi: &BragiHandler) {
     assert_eq!(get_value(stop, "id"), "stop_area:SA:second_station");
     assert_eq!(get_value(stop, "citycode"), "77487");
     assert_eq!(get_value(stop, "postcode"), "77000");
-    assert!(false);
```
azime: :+1

PR #142, opened 2017-02-27
manage when zip_code is present in the request

xlqian:
manage zip_codes in case where the zip_code is present in the request,
and zip_codes disturb the 100% label.prefix, it'll be a pity that we launch a fuzzy 
on the request  if the request is actually well spelt.  

Ex: 

q: 20 rue de thumesnil 59000

Solution: We use `multi_search` with `cross_fields` to manage this case.

Some notes regarding to the performance and the quality:

* Failed tests pass from 2083 tests to 1440 tests over 4863 tests (ile de france and npdc)
* Test time pass from 4min to 4min46s

### DIFF
--- a/libs/bragi/src/query.rs
+++ b/libs/bragi/src/query.rs
@@ -98,7 +98,7 @@ fn build_query(q: &str,
     };
 
     let boost_main_match_query = rs_q::build_match(main_match_type.to_string(), q.to_string())
-        .with_boost(1000)
+        .with_boost(500)
         .build();
 
     let boost_zipcode_match_query = rs_q::build_match("zip_codes.prefix".to_string(),
@@ -116,7 +116,7 @@ fn build_query(q: &str,
     // to have better results
     if match_type == MatchType::Fuzzy {
         should_query.push(rs_q::build_match("label.prefix", q.to_string())
-            .with_boost(2000)
+            .with_boost(1000)
             .build());
     }
 
@@ -125,7 +125,7 @@ fn build_query(q: &str,
         let boost_on_proximity =
             rs_q::build_function_score()
                 .with_boost_mode(rs_es::query::compound::BoostMode::Multiply)
-                .with_boost(500)
+                .with_boost(1500)
                 .with_function(Function::build_decay("coord",
                                            rs_u::Location::LatLon(c.lat, c.lon),
                                            rs_u::Distance::new(50f64,
@@ -137,7 +137,7 @@ fn build_query(q: &str,
         // if we don't have coords, we take the field `weight` into account
         let boost_on_weight = rs_q::build_function_score()
             .with_boost_mode(rs_es::query::compound::BoostMode::Multiply)
-            .with_boost(300)
+            .with_boost(500)
             .with_query(rs_q::build_match_all().build())
             .with_function(Function::build_field_value_factor("weight")
                 .with_factor(1)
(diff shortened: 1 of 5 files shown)


libs/bragi/src/query.rs
```diff
@@ -151,45 +151,54 @@ fn build_query(q: &str,
...
+    // A "must"" clause is used to manage zip_codes in case where the zip_code is present in the
+    // request and its appearance disturbs the 100% label.prefix, it'll be a pitty that we
+    // launch a fuzzy on the request if the request is actually well spelt.
+    // Ex:
+    //  q: 20 rue de thumesnil 59000
+    // WITH the following must clause, "59000" will match a zip_code, and "20 rue de thumesnil"
+    // will match the label.prefix
+    // WITHOUT the following must clause, since we don't have the zip codes in the labels
+    // the request will be rejected by 100% label.prefix and passed to label.ngram
+    let should = vec![rs_q::build_match(main_match_type.to_string(), q.to_string())
```
TeXitoi: It must only be done in the prefix case.

tests/bragi_osm_test.rs
```diff
@@ -61,15 +61,15 @@ pub fn bragi_osm_test(es_wrapper: ::ElasticSearchWrapper) {
 
 fn zip_code_test(bragi: &BragiHandler) {
     let all_20 = bragi.get("/autocomplete?q=77000");
-    assert_eq!(all_20.len(), 10);
+    assert_eq!(all_20.len(), 3);
```
TeXitoi: I don't like that, but we can live with.

xlqian: I rolled back that code

libs/bragi/src/query.rs
```diff
@@ -151,45 +151,44 @@ fn build_query(q: &str,
...
+    let second_condition = match match_type {
         MatchType::Prefix => {
-            // when we search [0, 2] fields, we need to match 100% of them
-            // when we search [3, 4] fields, we need to match 90% of them
-            // when we search [5, +] fields, we only need to match 70% on them
-            // this way if the input has more fields than the documents
-            // (additional information like the country, the region, ...)
-            // they can be ignored
-            MinimumShouldMatch::from(vec![min_match(2, 90f64), min_match(4, 70f64)])
+            rs_q::build_multi_match(vec!["label.prefix".to_string(),
```
antoine-de: can you add a comment explanning this ?

libs/bragi/src/query.rs
```diff
@@ -151,45 +151,44 @@ fn build_query(q: &str,
...
+                                         "zip_codes.prefix".to_string()],
+                                    q.to_string())
+                .with_type(rs_es::query::full_text::MatchQueryType::CrossFields)
+                .with_operator("and")
+                .build()
         }
         // for fuzzy search we lower our expectation and we accept 50% of token match
```
TeXitoi: unsyncronized comment

libs/bragi/src/query.rs
```diff
@@ -151,45 +151,44 @@ fn build_query(q: &str,
...
-            // this way if the input has more fields than the documents
-            // (additional information like the country, the region, ...)
-            // they can be ignored
-            MinimumShouldMatch::from(vec![min_match(2, 90f64), min_match(4, 70f64)])
+            rs_q::build_multi_match(vec!["label.prefix".to_string(),
+                                         "zip_codes.prefix".to_string()],
+                                    q.to_string())
+                .with_type(rs_es::query::full_text::MatchQueryType::CrossFields)
+                .with_operator("and")
+                .build()
```
TeXitoi: A small comment explaining that?

PR #143, opened 2017-03-08
 manage one-word misspelt request and very long request

xlqian:


libs/bragi/src/query.rs
```diff
@@ -183,12 +183,17 @@ fn build_query(q: &str,
                 .with_operator("and")
                 .build()
         }
-        // for fuzzy search we lower our expectation and we accept 45% of token match
+        // for fuzzy search we lower our expectation and we accept 40% of token match on label.ngram
+        // the "40%" is empirical, it's supposed to be able to manage cases like one-word missspelt
+        // requests and very long requets.
```
pbougue: reque`s`ts

libs/bragi/src/query.rs
```diff
@@ -183,12 +183,17 @@ fn build_query(q: &str,
...
+        // One-word misspelt request:
+        //     Vaureaaal (instead of Vaureal)
```
pbougue: Woulod be nice to add a test on that also (or on `buissson`)

xlqian:
an issue of rustfmt keep this PR from being checked... 
https://github.com/rust-lang-nursery/rustfmt/issues/1366

PR #144, opened 2017-03-13
feature api added

azime:
places by id, example:

`http://127.0.0.1:4000/features/admin:fr:77487?pt_dataset=d1`


antoine-de:
I'm more for a new API I find it clearer.

a `/detail` api ? `/object` ? `/uri` ?
should we return a list or just one object ? 
are we sure the id are unique even across indexes ? I don't think so especially for stops no ?
shouldn't we thus add a way to provide the pt_dataset ?

And I think we should not put this in `fn autocomplete` and `fn query`, the functions are already too complexes :stuck_out_tongue_winking_eye:

tests/bragi_stops_test.rs
```diff
@@ -162,3 +163,14 @@ fn stop_all_data_test(bragi: &BragiHandler) {
     let response = bragi.get("/autocomplete?q=14 juillet&_all_data=true");
     assert_eq!(response.len(), 2);
 }
+
+fn stop_by_id_test(bragi: &BragiHandler) {
+    // search with id
+    let response = bragi.get("/feature?id=stop_area:SA:second_station&pt_dataset=dataset1");
```
antoine-de: from a user perspective I would prefer `/feature/stop_area:SA:second_station?pt_dataset=dataset1`

and since we return a list (even if it's in most cases a one element list), wouldn't it be better to name it `features` ?
@TeXitoi @pbougue what do you think ?

libs/bragi/src/query.rs
```diff
@@ -301,15 +315,52 @@ fn query(q: &str,
...
+    collect(result)
+}
+
+pub fn feature(pt_dataset: &Option<&str>,
+               all_data: bool,
+               cnx: &str,
+               id: &Option<&serde_json::Value>)
+               -> Result<Vec<mimir::Place>, rs_es::error::EsError> {
+
+    let results = try!(query_feature(&pt_dataset, all_data, cnx, id));
```
antoine-de: you write this
```
query_feature(&pt_dataset, all_data, cnx, id)
``` 
but then is the function mandatory ?

libs/bragi/src/api.rs
```diff
@@ -142,11 +144,38 @@ impl ApiEndPoint {
...
+                });
+                let cnx = self.es_cnx_string.clone();
+                endpoint.handle(move |client, params| {
+                    let id = params.find("id");
+                    let pt_dataset = params.find("pt_dataset")
+                        .and_then(|val| val.as_str());
+                    let all_data =
+                        params.find("_all_data").and_then(|val| val.as_bool()).unwrap_or(false);
+
+                    let fearture = query::feature(&pt_dataset, all_data, &cnx, &id);
```
azime: ok !

libs/bragi/src/params.rs
```diff
@@ -0,0 +1,236 @@
...
+
+
+const MAX_LAT: f64 = 180f64;
+const MIN_LAT: f64 = -180f64;
+
+const MAX_LON: f64 = 90f64;
+const MIN_LON: f64 = -90f64;
+
+
+pub fn dataset_builder(params: &mut json_dsl::Builder) {
```
antoine-de: hum, maybe it would be better to call this dataset_param ?

antoine-de: yep, don't you think it's better ?

tests/bragi_features_test.rs
```diff
@@ -0,0 +1,84 @@
...
+pub fn bragi_features_test(es_wrapper: ::ElasticSearchWrapper) {
+    let bragi = BragiHandler::new(format!("{}/munin", es_wrapper.host()));
+
+    // *********************************
+    // We load the OSM dataset and three-cities bano dataset
+    // the current dataset are thus (load order matters):
+    // - osm_fixture.osm.pbf (including ways)
+    // - bano-three_cities
+    // *********************************
+    let osm2mimir = concat!(env!("OUT_DIR"), "/../../../osm2mimir");
```
azime: I prefer it to be in another file, a new API

tests/bragi_filter_types_test.rs
```diff
@@ -162,3 +164,42 @@ fn unvalid_type_test(bragi: &BragiHandler) {
...
+}
+
+fn stop_by_id_test(bragi: &BragiHandler) {
+    // search with id
+    let response = bragi.get("/features/stop_area:SA:second_station?pt_dataset=dataset1");
+    assert_eq!(response.len(), 1);
+    let stop = response.first().unwrap();
+    assert_eq!(get_value(stop, "id"), "stop_area:SA:second_station");
+
+}
```
azime: See: [NotFound](https://github.com/hyperium/hyper/blob/c42f18db05e47fc24e8a8ece76cbc782b7558e8b/src/status.rs#L324)

tests/bragi_filter_types_test.rs
```diff
@@ -202,4 +203,14 @@ fn stop_by_id_test(bragi: &BragiHandler) {
     let stop = response.first().unwrap();
     assert_eq!(get_value(stop, "id"), "stop_area:SA:second_station");
 
-}
\ No newline at end of file
+}
+
+fn stop_by_id_not_exist_test(bragi: &BragiHandler) {
+    // search with id
+    let response = bragi.raw_get("/features/stop_area:SA:second_station::AA?pt_dataset=dataset1").unwrap();
```
antoine-de: can you change the name to `stop_area_that_does_not_exists` ? I find it easier to read :wink:

PR #147, opened 2017-03-29
bragi: error handling

kadhikari:
Catch the error with following messages:
- "Elasticsearch Connection Error"
- "Index not found"

### DIFF
--- a/libs/bragi/src/query.rs
+++ b/libs/bragi/src/query.rs
@@ -495,4 +495,15 @@ fn test_make_indexes_impl() {
                    .unwrap(),
                Vec::<String>::new());
 
+    // dataset fr types poi, city, street, house without public_transport:stop_area
+    // and the function is_existing_index with an error in the result (Elasticsearch is absent..)
+    match make_indexes_impl(false,
+                            &Some("munin_stop_fr".to_string()),
+                            &Some(vec!["poi", "city", "street", "house"]),
+                            |_index| -> Result<bool, rs_es::error::EsError> {
+                                Err(rs_es::error::EsError::EsError("Elasticsearch".to_string()))
+                            }) {
+        Err(e) => assert_eq!(e.to_string(), "Elasticsearch".to_string()),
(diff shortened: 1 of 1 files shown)


libs/bragi/src/query.rs
```diff
@@ -214,8 +214,21 @@ fn build_query(q: &str,
...
-    !index.is_empty() && client.open_index(&index).is_ok()
+fn is_existing_index(client: &mut rs_es::Client,
+                     index: &str)
+                     -> Result<bool, rs_es::error::EsError> {
+    if index.is_empty() {
+        return Ok(false);
+    }
+    match client.open_index(&index) {
+        Err(::rs_es::error::EsError::EsServerError(_)) => {
+            Err(rs_es::error::EsError::EsServerError("Index not found ".to_string()))
```
pbougue: you probably can reuse original error and do a message like `format!("Index not found: {}", e.to_string())`
probably not working, but you have the idea (same below).

pbougue: And you have an extra space (in case you don't change message).
Same below.

libs/bragi/src/query.rs
```diff
@@ -214,8 +214,17 @@ fn build_query(q: &str,
...
-fn is_existing_index(client: &mut rs_es::Client, index: &str) -> bool {
-    !index.is_empty() && client.open_index(&index).is_ok()
+fn is_existing_index(client: &mut rs_es::Client,
+                     index: &str)
+                     -> Result<bool, rs_es::error::EsError> {
+    if index.is_empty() {
+        return Ok(false);
+    }
+    match client.open_index(&index) {
+        Err(::rs_es::error::EsError::EsServerError(_)) => Ok(false),
```
antoine-de: I don't understand why this is valid. We want an error if no ES can be found no ?

antoine-de: Ok I get what this error is, but I think a comment explaing what is this error would be great

PR #149, opened 2017-04-14
Add a reverse geocoding API to bragi

TeXitoi:
See https://github.com/benashford/rs-es/pull/107 for the modifications in rs-es

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,4 +50,3 @@ name = "tests"
 [replace]
 # remove this line once queryst > 1.0.1 is released
 "queryst:1.0.1" = { git = "https://github.com/rustless/queryst", rev = "b29604af7bfb05c2b9d98d726ab1bafd0533cc10" }
-"rs-es:0.9.0" = { git = "https://github.com/TeXitoi/rs-es", rev = "4e77efdfb7a54300035d5cbe217e6e866c56e227" }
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -50,3 +50,4 @@ name = "tests"
 [replace]
 # remove this line once queryst > 1.0.1 is released
 "queryst:1.0.1" = { git = "https://github.com/rustless/queryst", rev = "b29604af7bfb05c2b9d98d726ab1bafd0533cc10" }
+"rs-es:0.9.0" = { git = "https://github.com/TeXitoi/rs-es", rev = "4e77efdfb7a54300035d5cbe217e6e866c56e227" }
```
TeXitoi: To-do: update to rs-es 0.9.1

libs/bragi/src/query.rs
```diff
@@ -70,6 +70,19 @@ enum MatchType {
     Fuzzy,
 }
 
+fn build_proximity(coord: &model::Coord) -> rs_q {
```
antoine-de: can you add a comment ?

libs/bragi/src/query.rs
```diff
@@ -362,6 +365,24 @@ pub fn features(pt_dataset: &Option<&str>,
     }
 }
 
+pub fn reverse(coord: &model::Coord, cnx: &str) -> Result<Vec<mimir::Place>, EsError> {
```
antoine-de: can you add a comment ?

PR #150, opened 2017-05-05
Search with city and street name

azime:
* Query example : "201 avenue Jean Jaures Lievin"
> without correction :  "201 avenue Jean Jaurès Sentinelle"
> With correction : "avenue Jean Jaures Liévin" in position 10
* Query example : "201 avenue Jean Jaures Liévin"
> With correction : "avenue Jean Jaures Liévin" in position 1

* "geocoder-tester" on "iledefrance" data: 16 more tests passed

Jira: https://jira.canaltp.fr/browse/HOR-142


libs/bragi/src/query.rs
```diff
@@ -113,11 +113,26 @@ fn build_query(q: &str,
...
-    let mut should_query = vec![boost_addr,
-                                boost_admin,
-                                boost_stop,
-                                boost_main_match_query,
-                                boost_zipcode_match_query];
+    let mut should_query =
+        vec![boost_addr,
+             boost_admin,
+             boost_stop,
+             rs_q::build_bool()
```
antoine-de: I think we should add an OR between the field non existence  and the boost

libs/bragi/src/query.rs
```diff
@@ -142,6 +141,16 @@ fn build_query(q: &str,
                 .build())
             .build();
         should_query.push(boost_on_weight);
+
+        should_query.push(
+            rs_q::build_match("administrative_regions.label".to_string(), q.to_string())
+            .with_boost(50)
```
pbougue: maybe we should put that value in a variable to change both if needed?

libs/bragi/src/query.rs
```diff
@@ -92,6 +92,7 @@ fn build_query(q: &str,
                -> rs_es::query::Query {
 
     use rs_es::query::functions::Function;
+    let boost_ad_label = 50;
```
antoine-de: isn't it a bit low ? did you try several values ? (and did you rerun geocoder tester after @TeXitoi changes ?)

pbougue: minor: i find `boost_admin_label` more readable

azime: @antoine-de "16 more tests passed" after @TeXitoi changes.
I modified  boost_admin_label = 100 = boost_zipecode, now "17 more tests passed"

TeXitoi:
Don't work correctly :confused:

PR #155, opened 2017-05-23
Weight and tunnig

TeXitoi:


TeXitoi:
I can't give numbers for bench because I have france imported and bench only have Île-de-France.

between dev and my machine (should be about the same data), I have  488 -> 411

PR #156, opened 2017-06-01
Manage multiple admins with the same insee

TeXitoi:


src/osm_reader/admin.rs
```diff
@@ -101,9 +103,20 @@ pub fn administrative_regions(pbf: &mut OsmPbfReader, levels: BTreeSet<u32>) ->
...
+                .get("ref:INSEE")
+                .map(|v| v.trim_left_matches('0')) {
+                Some(val) if !insee_inserted.contains(val) => {
+                    insee_inserted.insert(val.to_string());
+                    (format!("admin:fr:{}", val), val)
+                }
                 Some(val) => {
-                    (format!("admin:fr:{}", val.trim_left_matches('0')), val.trim_left_matches('0'))
+                    let id = format!("admin:osm:{}", relation.id.0);
+                    warn!("relation/{}: have the INSEE {} that is already used, using {} as id",
```
antoine-de: minor: is this really a warning ?

are there very few admin in this case ?

TeXitoi: I've only found that on freshly merged admins, or with the INSEE reported to the upper levels.

PR #159, opened 2017-08-02
Cargo update

TeXitoi:


src/bin/stops2mimir.rs
```diff
@@ -68,7 +71,7 @@ struct Args {
 }
 
 struct StopPointIter<'a, R: std::io::Read + 'a> {
```
datanel: we can't remove this iterator with csv 1.0.0-beta.4 ?

TeXitoi: Work in progress

PR #161, opened 2017-08-02
Migrate to structopt

TeXitoi:
To review and merge after #160 is merged.

### DIFF
--- a/src/bin/osm2mimir.rs
+++ b/src/bin/osm2mimir.rs
@@ -29,14 +29,14 @@
 // www.navitia.io
 
 #[macro_use]
-extern crate serde_derive;
-extern crate serde;
-extern crate docopt;
+extern crate structopt_derive;
+extern crate structopt;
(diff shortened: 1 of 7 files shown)


src/bin/osm2mimir.rs
```diff
@@ -29,64 +29,62 @@
...
+    /// City level to  calculate weight.
+    #[structopt(short = "C", long = "city-level", default_value = "8")]
+    city_level: u32,
+    /// Elasticsearch parameters.
+    #[structopt(short = "c", long = "connection-string",
+                default_value = "http://localhost:9200/munin")]
+    connection_string: String,
+    /// Import ways.
+    #[structopt(short = "w", long = "import-way")]
+    import_way: bool,
```
antoine-de: It's for another PR, but I think we should cleanup those `import_*` parameters, I'm not convinced by their uses

PR #162, opened 2017-08-10
Add a global stops index

antoine-de:
As a step to have an auto-completion working cross navitia's coverages, when we import a stop we aggregate all the stops in a unique index.

For the moment the stops are 'merged' only if they got the same ID (that will likely change) and we retain the data only from one of the merged stop.

An additional stop field has been added, `coverages` with the list of the stop's coverage.

This will make it possible for bragi to filter a query on a set of coverage.

There is not the bragi part in this PR

### DIFF
--- a/libs/mimir/src/rubber.rs
+++ b/libs/mimir/src/rubber.rs
@@ -132,7 +132,7 @@ impl Rubber {
         )
     }
 
-    fn create_index(&self, name: &String) -> Result<(), String> {
+    pub fn create_index(&self, name: &String) -> Result<(), String> {
         debug!("creating index");
         // Note: in rs_es it can be done with MappingOperation but for the moment I think
         // storing the mapping in json is more convenient
--- a/libs/mimir/src/rubber.rs
+++ b/libs/mimir/src/rubber.rs
@@ -29,7 +29,7 @@
 // www.navitia.io
 
 
-use super::objects::{MimirObject, Admin};
+use super::objects::{Admin, MimirObject};
 use chrono;
 use hyper;
 use hyper::status::StatusCode;
@@ -38,8 +38,10 @@ use rs_es;
 use rs_es::EsResponse;
 use serde_json;
 use rs_es::operations::search::ScanResult;
-
-use super::objects::{AliasOperations, AliasOperation, AliasParameter};
+use serde;
+use std;
+use std::collections::BTreeMap;
+use super::objects::{AliasOperation, AliasOperations, AliasParameter};
 use rs_es::units::Duration;
 
 const SYNONYMS: [&'static str; 17] = [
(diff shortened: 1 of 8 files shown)


antoine-de:
Note: I get 0 dupplicate when merging the datasets `sncf`, `fr-nw`, and `fr-idf`

when adding `fr-se` I got 63...

Note: the query to get the dupplicate is (you should activate the groovy script on ES, it's only for test purpose)

```json
{
  "query": {
    "bool": {
      "filter": {
        "script": {
          "script": "_source.coverages.size() > 1"
        }
      }
    }
  }
}
```

libs/mimir/src/rubber.rs
```diff
@@ -257,7 +280,12 @@ impl Rubber {
 
     /// add a list of new indexes to the alias
     /// remove a list of indexes from the alias
-    fn alias(&self, alias: &str, add: &Vec<String>, remove: &Vec<String>) -> Result<(), String> {
+    pub fn alias(
+        &self,
+        alias: &str,
+        add: &Vec<String>,
+        remove: &Vec<String>,
```
TeXitoi: `s/&Vec<String>/&[String]/g`

libs/mimir/src/rubber.rs
```diff
@@ -384,14 +412,20 @@ impl Rubber {
...
-        let mut scan: ScanResult<Admin> = try!(
+    pub fn get_all_objects_from_index<T>(
+        &mut self,
+        index: &str,
+    ) -> Result<Vec<T>, rs_es::error::EsError>
+    where
+        for<'de> T: MimirObject + serde::de::Deserialize<'de> + std::fmt::Debug,
+    {
+        let mut result: Vec<T> = vec![];
+        let mut scan: ScanResult<T> = try!(
```
TeXitoi: EXTERMINATE! EXTERMINATE! `try!` EXTERMINATE!

libs/mimir/src/rubber.rs
```diff
@@ -410,6 +444,10 @@ impl Rubber {
         try!(scan.close(&mut self.es_client));
         Ok(result)
     }
+
+    fn get_admins_from_index(&mut self, index: &str) -> Result<Vec<Admin>, rs_es::error::EsError> {
+        self.get_all_objects_from_index(index)
+    }
```
TeXitoi: What the point?

tests/stops2mimir_test.rs
```diff
@@ -61,4 +62,53 @@ pub fn stops2mimir_sample_test(es_wrapper: ::ElasticSearchWrapper) {
...
+                    "stop_area:SA:second_station:dataset2" => {
+                        assert_eq!(stop.coverages, vec!["dataset2"])
+                    }
+                    _ => {
+                        print!("{}: {:?}", stop.id, stop);
+                        assert_eq!(stop.coverages, vec!["dataset1"])
+                    }
+                }
+            }
+            _ => assert!(false),
```
TeXitoi: `unreachable!()`

tests/stops2mimir_test.rs
```diff
@@ -61,4 +62,53 @@ pub fn stops2mimir_sample_test(es_wrapper: ::ElasticSearchWrapper) {
...
+                        assert!(
+                            vec!["All known stop", "All known stop, but different name"]
+                                .contains(&stop.label.as_ref())
+                        );
+                    }
+                    "stop_area:SA:second_station:dataset2" => {
+                        assert_eq!(stop.coverages, vec!["dataset2"])
+                    }
+                    _ => {
+                        print!("{}: {:?}", stop.id, stop);
```
TeXitoi: trace to remove ?

PR #166, opened 2017-11-22
set different MinimumShouldMatch per size of the query

kinnou02:
This should greatly improve the speed of bragi, chosen values are still a bit experimental, with these one there is 8 more failure on geocoder-tester but overall test are down to 104sec while they were taking 180 seconds before.

TODO:
  - [x] update comment

```
before: ============= 380 failed, 1712 passed, 1 skipped in 185.53 seconds =============
after: ============= 368 failed, 1724 passed, 1 skipped in 94.98 seconds ==============

```

poke @antoine-de  

This PR adds a new index, so bragi will require that all data have been re-indexed before working correctly. The previous version of bragi will still works even with the new indexes.

### DIFF
--- a/json/settings.json
+++ b/json/settings.json
@@ -55,6 +55,7 @@
                         "type": "string",
                         "index_options": "docs",
                         "analyzer": "word",
+                        "copy_to": "full_label",
                         "fields": {
                             "prefix": {
                                 "type": "string",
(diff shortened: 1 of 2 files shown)


TeXitoi:
Mmm, prefix match on zip code. That may be messy with house number.

PR #167, opened 2017-12-06
Usage of index template for settings and mapping

kinnou02:
Synonyms are still defined at the creation of the index and will
override those defined in the templates. But at least one synonym need to
be defined on the template.

Elasticsearch templates has to be initialized before importing data,
this is done with the `mimir_init` binary.

This PR may change the result of geocoder-tester since a lot of indexes
will have less shard than previously and that scoring is done per shard.
Having less shard reduce load on query and recent version of
Elasticsearch have a limit on the number of shard that can be queried.

Recommended values are the following:
  - less 3M documents: 1 shard
  - between 3M and 5M documents with an expected growth over 5M: 2 shards.
  - More than 5M: int (number of expected documents / 5M +1)

With this PR each index has mapping defined for only one type, this is
required by recent version of Elasticsearch.

Require: https://github.com/CanalTP/fab_mimir/pull/20


libs/mimir/src/rubber.rs
```diff
@@ -164,6 +164,44 @@ impl Rubber {
             })
     }
 
+    pub fn create_template(&self, name: &str, settings: &str) -> Result<(), String> {
```
antoine-de: could be private no ?

src/bin/mimir_init.rs
```diff
@@ -0,0 +1,60 @@
...
+extern crate mimirsbrunn;
+#[macro_use]
+extern crate log;
+
+use mimir::rubber::Rubber;
+use structopt::StructOpt;
+
+
+
+#[derive(StructOpt, Debug)]
```
antoine-de: so this is a new exe to be run the first time ?

how are updates handled ?

kinnou02: It is, as stated in the PR comment, it is already integrated to fab_mimir

src/bin/mimir_init.rs
```diff
@@ -0,0 +1,60 @@
...
+
+fn main() {
+    mimir::logger_init().unwrap();
+    info!("creating templates");
+
+    let args = Args::from_args();
+
+    let rubber = Rubber::new(&args.connection_string);
+    rubber.initialize_templates().unwrap();
+
```
pbougue: minor: extra line

src/bin/mimir_init.rs
TeXitoi: I don't understand why we can't do that lazily at each creation of a index.

TeXitoi: OK for me with the commit "update template before each data integration"

PR #169, opened 2017-12-19
properties for poi object

azime:


libs/bragi/src/model.rs
```diff
@@ -78,8 +79,11 @@ pub struct GeocodingResponse {
     pub administrative_regions: Vec<Rc<mimir::Admin>>,
     #[serde(skip_serializing_if = "Vec::is_empty", default)]
     pub poi_types: Vec<mimir::PoiType>,
+    #[serde(skip_serializing_if = "BTreeMap::is_empty")]
```
pbougue: (@TeXitoi) might more secure with a `default` here too, no? https://serde.rs/field-attrs.html#serdedefault

src/osm_reader/poi.rs
```diff
@@ -191,6 +191,13 @@ const DEFAULT_JSON_POI_TYPES: &'static str = r#"
 }
 "#;
 
+fn make_properties(tags: &osmpbfreader::Tags) -> BTreeMap<String, String>{
+    let mut result = BTreeMap::new();
+    for tag in tags.iter(){
+        result.insert(tag.0.clone(), tag.1.clone());
+    }
+    result
+}
```
pbougue: fmt: probably missing newline

antoine-de: isn't there a more rustic way to do this ? it seems a simple conversion between a flat_map and a BTreeMap that could be done with `into_iter` no ?

tests/bragi_poi_test.rs
```diff
@@ -89,6 +89,8 @@ fn poi_admin_address_test(bragi: &BragiHandler) {
 
     // the first element returned should be the poi 'Le-Mée-sur-Seine Courtilleraies'
     let poi = geocodings.first().unwrap();
+	let properties = poi.get("properties").and_then(|json| json.as_array()).unwrap();
+	assert_eq!(properties.len(), 9);
```
pbougue: It could be nice to have a test on key/value fields

libs/mimir/src/objects.rs
```diff
@@ -145,7 +145,11 @@ impl<T: MimirObject> MimirObject for Rc<T> {
         T::es_id(self)
     }
 }
-
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Propertie {
```
pbougue: typo: Property

src/osm_reader/poi.rs
```diff
@@ -191,6 +191,13 @@ const DEFAULT_JSON_POI_TYPES: &'static str = r#"
 }
 "#;
 
+fn make_properties(tags: &osmpbfreader::Tags) -> Vec<mimir::Propertie>{
+	tags.iter().map(|propretie| mimir::Propertie {
```
pbougue: typo: property, not propretie

pbougue:
And for rustfmt, we can do something like what's mentionned on that reddit : https://www.reddit.com/r/rust/comments/7l26y1/compliment_for_rustfmt_contributors/drj500o/
The first example looks simple enough (even if we would use it a little bit differently) : https://github.com/killercup/cargo-edit/blob/7a3964c08f363e4865fca531660414c2627ef901/.travis.yml#L9-L15

src/osm_reader/poi.rs
```diff
@@ -191,6 +191,13 @@ const DEFAULT_JSON_POI_TYPES: &'static str = r#"
 }
 "#;
 
+fn make_properties(tags: &osmpbfreader::Tags) -> Vec<mimir::Property>{
+	tags.iter().map(|propretie| mimir::Property {
```
pbougue: typo: propretie > property

PR #174, opened 2018-01-08
Import OpenAddresses

antoine-de:
connector to import data from https://openaddresses.io/, so now you can have addresses from all around the world :world_map: 

you can use this to import french data too if you want, but beware that it's the [ban](https://adresse.data.gouv.fr/), not [bano](http://openstreetmap.fr/bano).

(and it's a bit cumbersome since openaddresse split the files by regions)

### DIFF
--- a/src/bin/openaddresses2mimir.rs
+++ b/src/bin/openaddresses2mimir.rs
@@ -58,7 +58,7 @@ pub struct OpenAddresse {
     pub city: String,
     pub number: String,
     pub unit: String,
-    pub hash: String,
+    pub hash: Option<String>,
     pub lat: f64,
     pub lon: f64,
 }
--- a/src/bin/openaddresses2mimir.rs
+++ b/src/bin/openaddresses2mimir.rs
@@ -28,17 +28,17 @@
 // https://groups.google.com/d/forum/navitia
 // www.navitia.io
 
-#[macro_use]
-extern crate serde_derive;
-#[macro_use]
-extern crate structopt_derive;
-extern crate structopt;
 extern crate csv;
+extern crate geo;
+#[macro_use]
+extern crate log;
 extern crate mimir;
 extern crate mimirsbrunn;
 #[macro_use]
-extern crate log;
-extern crate geo;
+extern crate serde_derive;
+extern crate structopt;
+#[macro_use]
+extern crate structopt_derive;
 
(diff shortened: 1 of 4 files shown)


src/bin/openaddresses2mimir.rs
```diff
@@ -0,0 +1,178 @@
...
+        let addr_label = format!("{} ({})", addr_name, self.city);
+        let street_id = format!("street:{}", self.id); // TODO check if thats ok
+        let admins = admins_geofinder.get(&geo::Coordinate {
+            x: self.lat,
+            y: self.lon,
+        });
+
+        let weight = admins
+            .iter()
+            .find(|a| a.level == 8)
```
antoine-de: I don't like this (it has been copied from the bano import)

but I see 2 ways to better handle thoses:
* add a `city_level` in the cli args
* better handle the `city_level` in `osm2mimir`.
  we could tag the admin at import time, so only  `osm2mimir` has this logic and `bano2mimir` and `openaddresses2mimir` only check `a.level == AdminLevel::City`.
  this raise some questions, if you are interested I can create an issue

kinnou02: I think this should be handled in osm2mimir.
Later we could use another data source for admin/city

antoine-de: I totally agree with this (and I got ideas for the other admin data source)

I'll create an issue for this then

TeXitoi: That's OK for bano as city will always be 8.

I think short term can be an argument with default to 8. And open an issue to think about this problem.

src/bin/openaddresses2mimir.rs
```diff
@@ -0,0 +1,178 @@
...
+    let admins = rubber
+        .get_admins_from_dataset(dataset)
+        .unwrap_or_else(|err| {
+            info!(
+                "Administratives regions not found in es db for dataset {}. (error: {})",
+                dataset, err
+            );
+            vec![]
+        });
+    let admins_geofinder = admins.iter().cloned().collect();
```
TeXitoi:
```rust
let admins_geofinder = admins.into_iter().collect();
```

TeXitoi:
You need to rebase, the merged version doesn't compile.

PR #175, opened 2018-01-15
Add address in poi for autocomplete

kadhikari:
Add attribut address in poi. It can be of the type Addr or Street.
Ticket concerned: https://jira.canaltp.fr/browse/NAVP-721

### DIFF
--- a/libs/mimir/src/objects.rs
+++ b/libs/mimir/src/objects.rs
@@ -54,6 +54,7 @@ pub enum Place {
 
 /// Object stored in elastic search
 #[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(rename_all = "snake_case", tag = "type")]
 pub enum Address {
     Street(Street),
     Addr(Addr),
--- a/libs/mimir/src/objects.rs
+++ b/libs/mimir/src/objects.rs
@@ -52,6 +52,13 @@ pub enum Place {
     Stop(Stop),
 }
 
+/// Object stored in elastic search
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub enum Address {
+    Street(Street),
+    Addr(Addr),
+}
+
 impl Place {
     pub fn is_admin(&self) -> bool {
         match *self {
@@ -109,6 +116,16 @@ impl Place {
             Place::Stop(ref o) => o.admins(),
         }
     }
+    
+    pub fn address(&self) -> Option<Address> {
+        match *self {
+            Place::Admin(_) => None,
+            Place::Street(ref o) =>Some(Address::Street(o.clone())),
+            Place::Addr(ref o) => Some(Address::Addr(o.clone())),
+            Place::Poi(_) => None,
+            Place::Stop(_) => None,
+        }
+    }
 }
 
 pub trait MimirObject: serde::Serialize {
@@ -161,6 +178,7 @@ pub struct Poi {
     pub zip_codes: Vec<String>,
     pub poi_type: PoiType,
     pub properties: Vec<Property>,
+    pub address : Option<Address>,
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
--- a/libs/mimir/src/objects.rs
+++ b/libs/mimir/src/objects.rs
@@ -116,11 +116,11 @@ impl Place {
             Place::Stop(ref o) => o.admins(),
         }
     }
-    
+
     pub fn address(&self) -> Option<Address> {
         match *self {
             Place::Admin(_) => None,
-            Place::Street(ref o) =>Some(Address::Street(o.clone())),
+            Place::Street(ref o) => Some(Address::Street(o.clone())),
             Place::Addr(ref o) => Some(Address::Addr(o.clone())),
             Place::Poi(_) => None,
             Place::Stop(_) => None,
@@ -178,7 +178,7 @@ pub struct Poi {
     pub zip_codes: Vec<String>,
     pub poi_type: PoiType,
     pub properties: Vec<Property>,
-    pub address : Option<Address>,
+    pub address: Option<Address>,
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
--- a/libs/mimir/src/rubber.rs
+++ b/libs/mimir/src/rubber.rs
@@ -40,7 +40,7 @@ use rs_es::operations::search::ScanResult;
 use serde;
 use std;
 use std::collections::BTreeMap;
-use super::objects::{AliasOperation, AliasOperations, AliasParameter, Place, Coord};
+use super::objects::{AliasOperation, AliasOperations, AliasParameter, Coord, Place};
 use rs_es::units::Duration;
 use rs_es::units as rs_u;
 use rs_es::query::Query;
@@ -107,19 +107,16 @@ pub fn get_indexes_by_type(a_type: &str) -> String {
 pub fn collect(result: SearchResult<serde_json::Value>) -> Result<Vec<Place>, EsError> {
     debug!(
         "{} documents found in {} ms",
-        result.hits.total,
-        result.took
+        result.hits.total, result.took
     );
     // for the moment rs-es does not handle enum Document,
     // so we need to convert the ES glob to a Place
-    Ok(
-        result
-            .hits
-            .hits
-            .into_iter()
-            .filter_map(|hit| make_place(hit.doc_type, hit.source))
-            .collect(),
-    )
+    Ok(result
+        .hits
+        .hits
+        .into_iter()
+        .filter_map(|hit| make_place(hit.doc_type, hit.source))
+        .collect())
 }
 
 /// takes a ES json blob and build a Place from it
@@ -208,14 +205,12 @@ pub fn make_indexes_impl<F: FnMut(&str) -> Result<bool, EsError>>(
     let mut pt_dataset_indexes: Vec<String> = vec![];
     match pt_datasets.len() {
         0 => (),
-        1 => {
-            for pt_dataset in pt_datasets.iter() {
-                try!(push(
-                    &mut pt_dataset_indexes,
-                    format!("munin_stop_{}", pt_dataset).as_str(),
-                ))
-            }
-        }
+        1 => for pt_dataset in pt_datasets.iter() {
+            try!(push(
+                &mut pt_dataset_indexes,
+                format!("munin_stop_{}", pt_dataset).as_str(),
+            ))
+        },
         _ => try!(push(&mut pt_dataset_indexes, "munin_global_stops")),
     };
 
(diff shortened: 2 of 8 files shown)


TeXitoi:
rustfmt fails

tests/bragi_poi_test.rs
```diff
@@ -50,24 +50,24 @@ pub fn bragi_poi_test(es_wrapper: ::ElasticSearchWrapper) {
     // - osm_fixture.osm.pbf (including ways and pois)
     // - bano-three_cities
     // ******************************************
-    let osm2mimir = concat!(env!("OUT_DIR"), "/../../../osm2mimir");
+    let bano2mimir = concat!(env!("OUT_DIR"), "/../../../bano2mimir");
```
antoine-de: the comment above is not synchronized

tests/bragi_poi_test.rs
```diff
@@ -225,3 +227,31 @@ fn poi_misspelt_one_word_admin_test(bragi: &BragiHandler) {
...
+    assert!(count_types(&types, Poi::doc_type()) >= 1);
+    let poi = geocodings.first().unwrap();
+    assert_eq!(get_value(poi, "type"), Poi::doc_type());
+    assert_eq!(get_value(poi, "id"), "poi:osm:way:112361498");
+    assert_eq!(get_value(poi, "label"), "Hôtel de Ville (Melun)");
+    assert_eq!(get_poi_type_ids(poi), &["poi_type:amenity:townhall"]);
+    
+    assert_eq!(poi.get("address").and_then(|a| a.pointer("/Addr/house_number")), Some(&json!("2")));
+    assert_eq!(poi.get("address").and_then(|a| a.pointer("/Addr/id")), Some(&json!("addr:2.65801;48.53685")));
+    assert_eq!(poi.get("address").and_then(|a| a.pointer("/Addr/label")), Some(&json!("2 Rue de la Reine Blanche (Melun)")));
```
antoine-de: I don't understand how the address has the city name in it, since there was no OSM at reading time :thinking:

libs/mimir/src/objects.rs
```diff
@@ -52,6 +52,13 @@ pub enum Place {
     Stop(Stop),
 }
 
+/// Object stored in elastic search
+#[derive(Serialize, Deserialize, Debug, Clone)]
```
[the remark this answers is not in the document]
antoine-de: hum I don't really see the implication of this

libs/mimir/src/rubber.rs
```diff
@@ -90,6 +93,150 @@ pub fn get_date_index_name(base_index_name: &str) -> String {
...
+    pt_datasets: &[&str],
+    types: &[&str],
+    client: &mut rs_es::Client,
+) -> Result<Vec<String>, EsError> {
+    make_indexes_impl(all_data, pt_datasets, types, |index| {
+        is_existing_index(client, index)
+    })
+}
+
+pub fn make_indexes_impl<F: FnMut(&str) -> Result<bool, EsError>>(
```
antoine-de: I don't understand this code has been moved from somewhere no ?
I don't see where from in the PR

libs/mimir/src/rubber.rs
```diff
@@ -90,6 +93,150 @@ pub fn get_date_index_name(base_index_name: &str) -> String {
...
+    }
+    match client.open_index(&index) {
+        //This error indicates that the search index is absent in ElasticSearch.
+        Err(EsError::EsError(_)) => Ok(false),
+        Err(e) => Err(e),
+        Ok(_) => Ok(true),
+    }
+}
+
+pub fn make_indexes(
```
antoine-de: it's not really an index creation there, isn't it `get_indexes` ?

libs/mimir/src/objects.rs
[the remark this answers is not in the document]
TeXitoi: Small, only on the elasticsearch database. May be relevant if we want to use that in the search later.

tests/bragi_poi_test.rs
TeXitoi: BANO has the city name, that's the list of administravite_regions that may be inexistent.

libs/mimir/src/rubber.rs
TeXitoi: Looks like there is code duplication.

libs/mimir/src/rubber.rs
```diff
@@ -90,6 +93,150 @@ pub fn get_date_index_name(base_index_name: &str) -> String {
...
+        "public_transport:stop_area" => "stop",
+        "city" => "admin",
+        "house" => "addr",
+        _ => a_type,
+    };
+
+    format!("munin_{}", doc_type)
+}
+
+pub fn collect(result: SearchResult<serde_json::Value>) -> Result<Vec<Place>, EsError> {
```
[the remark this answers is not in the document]
TeXitoi: To returns the same thing as an autocompletion.

libs/mimir/src/rubber.rs
kadhikari: done

src/osm_reader/poi.rs
```diff
@@ -288,6 +289,24 @@ pub fn compute_poi_weight(pois_vec: &mut [Poi], city_level: u32) {
...
+                    addr[0].address()
+                } else {
+                    None
+                }
+            }
+            _ => {
+                warn!("The poi {:?} {:?} doesn't have address", poi.id, poi.name);
+                None
+            }
+        };
```
TeXitoi:
```rust
poi.address = rubber
    .get_address(&poi.coord)
    .ok()
    .and_then(|addrs| addrs.into_iter().next())
    .map(|addr| addr.address());
if poi.address.is_none() {
    warn!("The poi {:?} {:?} doesn't have address", poi.id, poi.name);
}
```

tests/bragi_poi_test.rs
TeXitoi: still relevant: unsynchronized comment.

tests/bragi_poi_test.rs
```diff
@@ -225,3 +226,39 @@ fn poi_misspelt_one_word_admin_test(bragi: &BragiHandler) {
...
+    // search poi: Poi as a relation in osm data
+    let geocodings = bragi.get("/autocomplete?q=Parking (Le Coudray-Montceaux)");
+    let types = get_types(&geocodings);
+    assert_eq!(count_types(&types, Poi::doc_type()), 1);
+    let first_poi = geocodings
+        .iter()
+        .find(|e| get_value(e, "type") == Poi::doc_type())
+        .unwrap();
+    assert_eq!(get_value(first_poi, "citycode"), "91179");
+    assert_eq!(get_poi_type_ids(first_poi), &["poi_type:amenity:parking"]);
```
TeXitoi: That's just a copy and paste from another test? seems irrelevant here.

tests/bragi_poi_test.rs
```diff
@@ -225,3 +226,39 @@ fn poi_misspelt_one_word_admin_test(bragi: &BragiHandler) {
...
+
+    // search poi: Poi as a way in osm data
```
TeXitoi: We don't care of this detail here.

tests/bragi_poi_test.rs
```diff
@@ -225,3 +226,39 @@ fn poi_misspelt_one_word_admin_test(bragi: &BragiHandler) {
...
+    let geocodings = bragi.get("/autocomplete?q=77000 Hôtel de Ville (Melun)");
+    let types = get_types(&geocodings);
+    assert!(count_types(&types, Poi::doc_type()) >= 1);
+    let poi = geocodings.first().unwrap();
+    assert_eq!(get_value(poi, "type"), Poi::doc_type());
+    assert_eq!(get_value(poi, "id"), "poi:osm:way:112361498");
+    assert_eq!(get_value(poi, "label"), "Hôtel de Ville (Melun)");
+    assert_eq!(get_poi_type_ids(poi), &["poi_type:amenity:townhall"]);
```
TeXitoi: The two last asserts are not relevant here.

libs/mimir/src/rubber.rs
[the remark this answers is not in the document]
antoine-de: I do not agree, it's nice that the API stays the same, but this is a helper, and it's not classy to do a `addrs.into_iter().next()` (or worse a check then `[0]` :stuck_out_tongue_winking_eye: ) to use the field.

I think this should return an `Option` and it should be wrapped in a `Vec` in the `GeocodingResponse`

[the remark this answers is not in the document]
TeXitoi: `fn collect` is also used in `fn features`. It get a raw elasticsearch response and gives you the corresponding `Vec<Place>`.

libs/mimir/src/rubber.rs
```diff
@@ -258,6 +400,29 @@ impl Rubber {
             .collect())
     }
 
+    pub fn get_address(&mut self, coord: &Coord) -> Result<Vec<Place>, EsError> {
```
TeXitoi: @antoine-de this one can return an Option instead of a Vec, but not collect.

TeXitoi: It was like that before and changing this type will obscurate reverse. I don't know which one is better, but that's true that an Option reflect more the intent of this method.

libs/mimir/src/rubber.rs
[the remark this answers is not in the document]
antoine-de: why does `features` returns a `Vec` ? (and collect is really not very explicit)

libs/mimir/src/rubber.rs
antoine-de: ah yes maybe that's enough

libs/mimir/src/rubber.rs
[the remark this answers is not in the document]
TeXitoi: features is used for autocomplete, thus a vector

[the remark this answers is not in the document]
TeXitoi: for the naming, that's Abder's fault! https://github.com/CanalTP/mimirsbrunn/commit/4dc9883d28429146454a980f08324d017663e16e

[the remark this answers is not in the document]
TeXitoi: my bad, feature is also a one response api. But that's not code from this PR.

@kadhikari feel free to refactor or not as this is not your responsability.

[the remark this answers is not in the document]
antoine-de: for the naming, that's not the same as at the naming time the function was private :stuck_out_tongue_winking_eye:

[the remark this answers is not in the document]
kadhikari: @TeXitoi for the refactor, i will wait monday since i won't  be present tomorrow

kadhikari:
@TeXitoi could you merge as it is and i will midify in the next PR.

PR #182, opened 2018-01-17
Add street in street

antoine-de:
small modification filling the "street" field in the api response if the object is a "street".
In this case "street" will be a duplicate with "name" but I think it's more coherent, and it enable tools like geocoder tester to test the "street" field for addresses and streets


TeXitoi:
for the rustfmt crash: https://github.com/rust-lang-nursery/rustfmt/issues/2370

PR #185, opened 2018-01-19
Adding AdminType (City or Unknown) to Admins.

### LINKED ISSUE #176: More centralized admin handling
Hi all,

In osm2mimir there is a `city_level` to be able to determine which admin is a city.

We also need to know if and admin is a city in openaddresses2mimir (and in bano too even if we can hardcode the french osm `admin_level` to 8).

I think we can just add a `type` to the admin.

To be able to have finer grained types i think the best approach is an enum:

```rust
enum AdminType {
    City,
    Unkown
};
```

In the future we could add [libpostal hierarchy](https://github.com/openvenues/libpostal#parser-labels) or [Wof hierarchy](https://github.com/whosonfirst/whosonfirst-placetypes), but that's another story :wink: 

For the naming, i'm not convinced by `type`, do you have better alternative ? `level` is not available since it would break retro compatibility.

If you agree on this, we're willing to implement it.

jbgriesner:
The code in "./bin" (i.e. osm2mimir, bano2mimir,  bragi, openaddresses2mimir etc) require a "city_level" parameter to control which admin is a city (the default value is level 8).

To handle in a close future more sophisticated admin hierarchies, we have added a "type" to the Admin thanks to the enum "AdminType" that we defined in "mimir/src/objects". 

Instead of comparing the current admin level to the city_level, we have access to the information directly inside the Admin.

Fixes #176


libs/mimir/src/objects.rs
```diff
@@ -253,6 +259,29 @@ pub struct Admin {
             deserialize_with = "custom_multi_polygon_deserialize",
             skip_serializing_if = "Option::is_none", default)]
     pub boundary: Option<geo::MultiPolygon<f64>>,
+    pub Type: AdminType,
```
TeXitoi: fields must be snake_case, not PascalCase. I would name that `admin_type`.

libs/mimir/src/objects.rs
```diff
@@ -253,6 +259,29 @@ pub struct Admin {
...
+impl Admin {
+    pub fn is_city(&self) -> bool {
+        match *self {
+            Admin {
+                Type: AdminType::City,
+                ..
+            } => true,
+            _ => false,
+        }
+    }
```
TeXitoi:
```rust
pub fn is_city(&self) -> bool {
    match self.admin_type {
        AdminType::City => true,
        _ => false,
    }
}
```

src/bin/openaddresses2mimir.rs
```diff
@@ -151,9 +151,6 @@ struct Args {
     /// Name of the dataset.
     #[structopt(short = "d", long = "dataset", default_value = "fr")]
     dataset: String,
-    /// City level to  calculate weight.
-    #[structopt(short = "C", long = "city-level", default_value = "8")]
-    city_level: u32,
```
TeXitoi: You can't remove the option now, it will fail in our integration. For the moment, just let the option, and log a warn if it is used to say that's deprecated.

libs/mimir/src/objects.rs
```diff
@@ -253,6 +259,29 @@ pub struct Admin {
...
+}
+
+impl fmt::Display for Admin {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        if self.is_city() {
+            write!(f, "{:?}", AdminType::City)
+        } else {
+            write!(f, "{:?}", AdminType::Unknown)
+        }
+    }
```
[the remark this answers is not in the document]
TeXitoi: Still relevant

libs/mimir/src/objects.rs
```diff
@@ -253,6 +259,25 @@ pub struct Admin {
...
+impl Admin {
+    pub fn is_city(&self) -> bool {
+        match self.admin_type {
+            AdminType::City => true,
+            _ => false,
+        }
+    }
+}
+
+impl fmt::Display for Admin {
```
TeXitoi: Display is for user-facing output. You don't implement user-facing output for Admin here.

TeXitoi:
Thanks for the contribution!

src/bin/openaddresses2mimir.rs
TeXitoi:
```rust
/// Deprecated option.
#[structopt(short = "C", long = "city-level")]
city_level: Option<String>,
```

Somewhere
```rust
if args.city_level.is_some() {
    warn!("city-level option is deprecated, it now has no effect.");
}
```

jbgriesner: My bad, I correct it asap.

jbgriesner:
Tests code remain for future work

libs/mimir/src/objects.rs
jbgriesner: Which trait do you think should be implemented ?
I just trusted the doc : https://doc.rust-lang.org/stable/std/string/trait.ToString.html

jbgriesner: I'll fix it in my next commit:)

src/bin/ntfs2mimir.rs
```diff
@@ -55,9 +55,9 @@ struct Args {
     #[structopt(short = "c", long = "connection-string",
                 default_value = "http://localhost:9200/munin")]
     connection_string: String,
-    /// City level to calculate weight.
+    /// Deprecated option.
     #[structopt(short = "C", long = "city-level", default_value = "8")]
-    city_level: u32,
+    city_level: Option<u32>,
```
TeXitoi: Did you checked that, if the option is not provided, the warning is not emitted?

libs/mimir/src/objects.rs
```diff
@@ -253,6 +259,25 @@ pub struct Admin {
...
+impl Admin {
+    pub fn is_city(&self) -> bool {
+        match self.admin_type {
+            AdminType::City => true,
+            _ => false,
+        }
+    }
+}
+
+impl fmt::Display for AdminType {
```
TeXitoi: Please put the impl near the definition of the corresponding type.

tests/bragi_osm_test.rs
```diff
@@ -122,3 +124,10 @@ fn zip_code_admin_test(bragi: &BragiHandler) {
     let count = count_types(&types, "house");
     assert_eq!(count, 0);
 }
+
+fn city_admin_test(bragi: &BragiHandler) {
+    let all_melun = bragi.get("/autocomplete?q=Melun Rp");
+    let cities_all_melun = filter_by_type(&all_melun, "city");
+    let types = get_types(&cities_all_melun);
+    assert!(types.iter().all(|r| *r == "city",));
```
antoine-de: this test is useless since the type is filtered by `filter_by_type`

you should  just check that `cities_all_melun` has at least one element.

and you should check an `unknown` admin if there is one in the dataset.

jbgriesner: You're right: I fixed it.
Concerning the test for `unknown` there is no admin with unknown type in the imported data.
So we cannot test this feature here.

libs/mimir/src/objects.rs
```diff
@@ -253,6 +268,16 @@ pub struct Admin {
             deserialize_with = "custom_multi_polygon_deserialize",
             skip_serializing_if = "Option::is_none", default)]
     pub boundary: Option<geo::MultiPolygon<f64>>,
+    pub admin_type: AdminType,
```
TeXitoi: The new version will be unable to read data from the old version. Please add `#[serde(default = "admin_type_city")]` to this field. The needed steps to make that compilable is left as an exercise.

jbgriesner:
I checked backward compatibility of the new version (with the "serde(default = "admin_type_city")" annotation) and it looks fine.
As @antoine-de suggested it, you are free to squash all commits & merge.

PR #186, opened 2018-01-29
Fix lat lon in ES

antoine-de:
## invert coordinate to have lon/lat in the right order in ElasticSearch

we handle coordinate as `x`/`y` internally, but it the end we want `lon`/`lat` and there was a classical swap between `lon` and `lat`.

The swap was not detected before because bragi swap again the coordinates so in bragi the coordinates are ok even if they are inversed in ES 🤣 

to clarify, now we always do `lon`=`x` and `lat`=`y`, and we always give first `lon` then `lat` in the coordinate constructor

WARNING, this is a breaking change, I don't see any (easy) way to make this retrocompatible, It will need a full reindex !

### DIFF
--- a/libs/bragi/src/api.rs
+++ b/libs/bragi/src/api.rs
@@ -130,8 +130,8 @@ impl ApiEndPoint {
                 let cnx = self.es_cnx_string.clone();
                 endpoint.handle(move |client, params| {
                     let coord = ::mimir::Coord::new(
-                        params.find("lat").and_then(|p| p.as_f64()).unwrap(),
                         params.find("lon").and_then(|p| p.as_f64()).unwrap(),
+                        params.find("lat").and_then(|p| p.as_f64()).unwrap(),
                     );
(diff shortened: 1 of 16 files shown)


kinnou02:
It's worst than that, even with a full reindex there is a time where the coordinate will be wrong in the response :sob: 
The only ways to prevent this would be to:
  - use new fields for the correct X Y
  - switch the alias on the new indexes when bragi start (and even like this a multinode environnement will encounters some errors)

antoine-de:
ah we could indeed copy all the boundaries/coord to other fields and thus have:
* `coord`: wrongly represented coord
* `good_coord`: good coord

and in bragi add a test that if we get a `good_coord` we return it, else we switch the `coord` lon/lat

but do you think we need this ? you are not yet in production no ? the breaking change way would be way easier 😜

antoine-de:
any thoughts on this PR?

PR #187, opened 2018-02-06
stricter types in Rubber

antoine-de:
a common pattern while using Rubber is (the example is taken from `bano2mimir`):
```rust
let addr_index = rubber.make_index(Addr::doc_type(), dataset)?;
rubber.bulk_index(&addr_index, iter)?;
rubber.publish_index(Addr::doc_type(), dataset, addr_index, Addr::is_geo_data())?
```

it feels error prone to write it like this.

This PR add a `TypedIndex<T: MimirObject>` for more compil time check

The same code is now written
```rust
let addr_index = rubber.make_index(dataset)?; // <- since we bulk insert addresses, the compiler knows it's a TypedIndex<Addr>
rubber.bulk_index(&addr_index, iter)?;
rubber.publish_index(dataset, addr_index)?
```

with explicit types it is
```rust
let addr_index = rubber.make_index::<Addr>(dataset)?;
rubber.bulk_index::<Addr>(&addr_index, iter)?;
rubber.publish_index::<Addr>(dataset, addr_index)?
```

I think this has even more meaning since the next version of elastic search will be able to store only one type per index (as mimir does) [well the types are even prone to disapear](https://www.elastic.co/guide/en/elasticsearch/reference/6.x/removal-of-types.html)

### DIFF
--- a/libs/mimir/src/rubber.rs
+++ b/libs/mimir/src/rubber.rs
@@ -75,7 +75,7 @@ pub struct Rubber {
 }
 
 pub struct TypedIndex<T> {
-    pub name: String,
+    name: String,
     _type: PhantomData<T>,
 }
 
@@ -579,7 +579,7 @@ impl Rubber {
         // TODO better error handling
         let index = self.make_index(dataset)?;
         let nb_elements = self.bulk_index(&index, iter).map_err(|e| e.to_string())?;
-        self.publish_index::<T>(dataset, index)?;
+        self.publish_index(dataset, index)?;
         Ok(nb_elements)
     }
 
--- a/libs/mimir/src/rubber.rs
+++ b/libs/mimir/src/rubber.rs
@@ -76,14 +76,14 @@ pub struct Rubber {
 
 pub struct TypedIndex<T> {
     pub name: String,
-    _type: PhantomData<T>
+    _type: PhantomData<T>,
 }
 
 impl<T> TypedIndex<T> {
     pub fn new(name: String) -> TypedIndex<T> {
         TypedIndex {
             name: name,
-            _type: PhantomData
+            _type: PhantomData,
         }
     }
 }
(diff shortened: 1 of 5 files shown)


libs/mimir/src/rubber.rs
```diff
@@ -565,24 +577,21 @@ impl Rubber {
         I: Iterator<Item = T>,
     {
         // TODO better error handling
-        let index = try!(self.make_index(T::doc_type(), dataset));
-        let nb_elements = try!(self.bulk_index(&index, iter).map_err(|e| e.to_string()));
-        try!(self.publish_index(T::doc_type(), dataset, index, T::is_geo_data(),));
+        let index = self.make_index(dataset)?;
+        let nb_elements = self.bulk_index(&index, iter).map_err(|e| e.to_string())?;
+        self.publish_index::<T>(dataset, index)?;
```
TeXitoi: that's surprising to givr the type hint here and not on make_index

libs/mimir/src/rubber.rs
```diff
@@ -73,16 +74,30 @@ pub struct Rubber {
     http_client: hyper::client::Client,
 }
 
+pub struct TypedIndex<T> {
+    pub name: String,
```
TeXitoi: I think you don't need this pub

TeXitoi:
tests are not compiling

src/bin/bano2mimir.rs
```diff
@@ -42,7 +42,7 @@ extern crate structopt_derive;
 
 use std::path::Path;
 use mimir::rubber::Rubber;
-use mimir::objects::{Addr, Admin};
+use mimir::objects::{Admin};
```
TeXitoi: useless {}

antoine-de:
I really need to find a way to always run rust-fmt I always forget 🤣 

corrected, thanks for the review!

PR #188, opened 2018-02-08
move the boundary building into another repository

antoine-de:
the boundary building is needed either by mimir and by cosmogony, so we
move it into another crate.

DO NOT merge this PR yet, it depends on https://github.com/QwantResearch/osm_boundaries_utils_rs/pull/1

once the previous PR has been merged, I will update the Cargo.toml to point to the master branch


Cargo.toml
```diff
@@ -24,6 +24,7 @@ ordered-float = "0.0.2"
 mdo = "*"
 itertools = "*"
 navitia_model = { git = "https://github.com/CanalTP/navitia_model" }
+osm_boundaries_utils = { git = "https://github.com/QwantResearch/osm_boundaries_utils_rs", branch = "init" }
```
pbougue: So it's allright, now, no?

antoine-de: yep, updated, thanks

PR #193, opened 2018-02-23
add timezone, codes and properties in stop_area

xlqian:
https://jira.canaltp.fr/browse/NAVP-726

### DIFF
--- a/libs/mimir/src/objects.rs
+++ b/libs/mimir/src/objects.rs
@@ -222,6 +222,12 @@ pub struct PhysicalMode {
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Code {
+    pub name: String,
+    pub value: String,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct Stop {
     pub id: String,
     pub label: String,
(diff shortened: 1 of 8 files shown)


libs/mimir/src/objects.rs
```diff
@@ -236,6 +242,12 @@ pub struct Stop {
     pub physical_modes: Vec<PhysicalMode>,
     #[serde(default)]
     pub coverages: Vec<String>,
+    #[serde(default)]
```
xlqian: @TeXitoi  is this necessaryN

libs/mimir/src/objects.rs
```diff
@@ -236,6 +242,12 @@ pub struct Stop {
...
+    pub timezone: String,
+    #[serde(default)]
```
xlqian: @TeXitoi is this necessary?

libs/mimir/src/objects.rs
```diff
@@ -236,6 +242,12 @@ pub struct Stop {
...
+    pub codes: Vec<Code>,
+    #[serde(default)]
```
xlqian: @TeXitoi is this necessary?

libs/mimir/src/objects.rs
TeXitoi: If new bragi real old database, without that it will crash.

libs/mimir/src/objects.rs
TeXitoi: same

libs/mimir/src/objects.rs
TeXitoi: same

PR #196, opened 2018-02-28
add basic prometheus metrics to bragi

kinnou02:
the endpoint `/metrics` return them


libs/bragi/src/api.rs
```diff
@@ -121,6 +165,21 @@ impl ApiEndPoint {
...
+    fn prometheus(&self) -> rustless::Api {
+        Api::build(|api| {
+            api.get("metrics", |endpoint| {
+                endpoint.handle(move |mut client, _params| {
+                    let encoder = prometheus::TextEncoder::new();
+                    let metric_familys = prometheus::gather();
+                    let mut buffer = vec![];
+                    encoder.encode(&metric_familys, &mut buffer).unwrap();
+                    client.set_content_type(encoder.format_type().parse::<Mime>().unwrap());
+                    client.text(String::from_utf8(buffer).unwrap())
```
TeXitoi: A bit verbose, but I didn't found anything better. The crate would need a bit of polish.

Cargo.toml
```diff
@@ -28,6 +28,8 @@ mdo = "*"
 itertools = "*"
 navitia_model = { git = "https://github.com/CanalTP/navitia_model" }
 osm_boundaries_utils = { git = "https://github.com/QwantResearch/osm_boundaries_utils_rs", version = "0.1" }
+prometheus = "0.3"
+lazy_static = "1.0"
```
TeXitoi: Why do you touch to the global Cargo.toml if you only modify the bragi crate?

kinnou02: I made some tests in the main of bragi at first.
I have removed them.

libs/bragi/src/api.rs
```diff
@@ -92,6 +116,57 @@ impl ApiEndPoint {
...
+                    })
+                    .and_then(|counter| {
+                        counter.inc();
+                        Ok(())
+                    })
+                    .or_else(|err| -> Result<(), ()> {
+                        error!("impossible to get HTTP_COUNTER metrics"; "err" => err.to_string());
+                        Ok(())
+                    })
+                    .unwrap();
```
TeXitoi: `.unwrap_or_else`

libs/bragi/src/api.rs
```diff
@@ -92,6 +116,57 @@ impl ApiEndPoint {
...
+
+                HTTP_REQ_HISTOGRAM
+                    .get_metric_with(&labels!{
+                        "handler" => client.endpoint.path.path.as_str(),
+                        "method" => method.as_str(),
+                    })
+                    .and_then(|timer| {
+                        client.ext.insert::<Timer>(timer.start_timer());
+                        Ok(())
+                    })
```
TeXitoi: you want to `map` here

libs/bragi/src/api.rs
```diff
@@ -92,6 +116,57 @@ impl ApiEndPoint {
...
+                HTTP_COUNTER
+                    .get_metric_with(&labels!{
+                        "handler" => client.endpoint.path.path.as_str(),
+                        "method" => method.as_str(),
+                        "status" => code.as_str(),
+                    })
+                    .and_then(|counter| {
+                        counter.inc();
+                        Ok(())
+                    })
```
TeXitoi: You want to `map` here

libs/bragi/src/api.rs
```diff
@@ -92,6 +116,57 @@ impl ApiEndPoint {
...
+                    .unwrap();
+                client
+                    .ext
+                    .remove::<Timer>()
+                    .ok_or("")
+                    .and_then(|timer| {
+                        timer.observe_duration();
+                        Ok(())
+                    })
```
TeXitoi: `map`

libs/bragi/src/api.rs
```diff
@@ -92,6 +116,57 @@ impl ApiEndPoint {
...
+                    .or_else(|err| -> Result<(), ()> {
+                        error!("impossible to get HTTP_COUNTER metrics"; "err" => err.to_string());
+                        Ok(())
+                    })
+                    .unwrap();
+                client
+                    .ext
+                    .remove::<Timer>()
+                    .ok_or("")
```
TeXitoi: useless, just keep the option

libs/bragi/src/api.rs
```diff
@@ -92,6 +116,57 @@ impl ApiEndPoint {
...
+                    .and_then(|timer| {
+                        timer.observe_duration();
+                        Ok(())
+                    })
+                    .or_else(|_err| {
```
TeXitoi: `unwrap_or_else`

antoine-de:
seems nice!

do you know if it's costly to record all those events?

and how does this work, all is stored in bragi or is it send to another service?

kinnou02:
it's pretty cheap, we only store a few int into some maps, something like this: `map<string, map<string, map<float, int>>>`
and these integer are increased at each call.
prometheus call every few secs (15 by default) the `/metrics` endpoint and store them, this doesn't reset our counters, this only happen at restart.

libs/bragi/src/query.rs
```diff
@@ -238,6 +260,13 @@ fn query(
         // but we want to return empty response in this case.
         return Ok(vec![]);
     }
+    let timer = ES_REQ_HISTOGRAM
+        .get_metric_with_label_values(&[query_type.as_str()])
+        .map(|h| Some(h.start_timer()))
+        .unwrap_or_else(|err| {
+            error!("impossible to get ES_REQ_HISTOGRAM metrics"; "err" => err.to_string());
+            None
+        });
```
TeXitoi:
```rust
let timer = ES_REQ_HISTOGRAM
    .get_metric_with_label_values(&[query_type.as_str()])
    .map(|h| h.start_timer())
    .map_err(|err| error!("impossible to get ES_REQ_HISTOGRAM metrics"; "err" => err.to_string()))
    .ok()
```

kinnou02: fmt added a few lines :(

PR #202, opened 2018-03-20
Better error management

xlqian:
use the crate failure(https://boats.gitlab.io/failure/intro.html) for a better error management


src/bin/bano2mimir.rs
```diff
@@ -151,21 +153,18 @@ where
...
-            Err(e) => panic!("failed to bulk insert file {:?} because: {}", &f, e),
-            Ok(nb) => info!("importing {:?}: {} addresses added.", &f, nb),
-        }
+        let nb = rubber
+            .bulk_index(&addr_index, iter)
+            .context(format!("failed to bulk insert file {:?}", &f))?;
+        info!("importing {:?}: {} addresses added.", &f, nb);
     }
-    rubber.publish_index(dataset, addr_index).unwrap();
+    rubber.publish_index(dataset, addr_index)
```
xlqian: how can i add context here?

src/bin/mimir_init.rs
```diff
@@ -46,12 +47,21 @@ struct Args {
     connection_string: String,
 }
 
+pub fn run() -> Result<(), failure::Error> {
+    let args = Args::from_args();
+
+    let rubber = Rubber::new(&args.connection_string);
+    rubber.initialize_templates()
```
xlqian: same here

libs/mimir/src/rubber.rs
```diff
@@ -280,21 +282,28 @@ impl Rubber {
...
+        dataset: &str,
+    ) -> Result<TypedIndex<T>, failure::Error> {
         let index_name = get_date_index_name(&get_main_type_and_dataset_index::<T>(dataset));
         info!("creating index {}", index_name);
         self.create_index(&index_name.to_string())?;
         Ok(TypedIndex::new(index_name))
     }
 
-    pub fn create_index(&self, name: &String) -> Result<(), String> {
+    pub fn create_index(&self, name: &String) -> Result<(), failure::Error> {
```
antoine-de: wouldn't it be better to have a custom rubber error here ?

src/bin/bano2mimir.rs
antoine-de: what do you mean by context ?
https://docs.rs/failure/0.1.1/failure/struct.Error.html#method.context ?

antoine-de: if you mean add more detail to the error, typing it might help (but I don't find this perfeect)

you can do something like https://github.com/osm-without-borders/cosmogony/blob/master/src/zone_typer.rs#L23 for example

src/bin/stops2mimir.rs
```diff
@@ -166,7 +159,23 @@ fn main() {
         })
         .collect();
     set_weights(stops.iter_mut(), &nb_stop_points);
-    import_stops(stops, &args.connection_string, &args.dataset);
+    import_stops(stops, &args.connection_string, &args.dataset)
```
xlqian: same here

src/stops.rs
```diff
@@ -52,24 +53,28 @@ where
...
-    let global_index = update_global_stop_index(&mut rubber, stops.iter(), dataset).unwrap();
+    let global_index = update_global_stop_index(&mut rubber, stops.iter(), dataset)?;
 
     info!("Importing {} stops into Mimir", stops.len());
-    let nb_stops = rubber.index(dataset, stops.iter()).unwrap();
+    let nb_stops = rubber.index(dataset, stops.iter())?;
     info!("Nb of indexed stops: {}", nb_stops);
 
-    publish_global_index(&mut rubber, &global_index).unwrap();
+    publish_global_index(&mut rubber, &global_index)
```
xlqian: same here

src/bin/bano2mimir.rs
xlqian: @antoine-de Hi, antoine, good to see you here again :heart:  
I've been looking for the right way, so hard,  to do this 
```
 rubber.publish_index(dataset, addr_index).context("blabla")
```
the compiler won't let me pass :( 

thx for the doc, i'll read them

libs/mimir/src/rubber.rs
TeXitoi: If we need to match on the error kind, we will do the refactor, but for now, that's not really useful.

src/bin/mimir_init.rs
TeXitoi: initialize_template is quite explicit, thus context is not really needed.

src/bin/stops2mimir.rs
TeXitoi: added

src/stops.rs
TeXitoi: added

src/bin/openaddresses2mimir.rs
```diff
@@ -154,27 +155,36 @@ struct Args {
...
+}
+
+fn main() {
+    let _guard = mimir::logger_init();
+    info!("importing open addresses into Mimir");
+
+    let args = Args::from_args();
+    if let Err(err) = run(args) {
+        for cause in err.causes() {
+            eprintln!("{}", cause);
```
antoine-de: why not use the logger `error!` here ?

src/bin/openaddresses2mimir.rs
```diff
@@ -154,27 +155,36 @@ struct Args {
...
+        )
+    }
+}
+
+fn main() {
+    let _guard = mimir::logger_init();
+    info!("importing open addresses into Mimir");
+
+    let args = Args::from_args();
+    if let Err(err) = run(args) {
```
antoine-de: this since pattern is found in every mimir binary, wouldn't it be nice to factorize it ?

src/bin/openaddresses2mimir.rs
antoine-de: really minor, but in cosmogony I also output the `backtrace`s contained in the `cause`. Normally it should add some more info when run with `RUST_BACKTRACE=1`, but I never managed to have some :confused:  I think it's because it depends on the error creation. Have you experiences with those @TeXitoi ?

libs/mimir/src/rubber.rs
antoine-de: since rubber is used a a library, I though it would match more the [Failure guide](https://boats.gitlab.io/failure/error-errorkind.html), but I aggre that it's more work, so I think it's fine to postpone it

src/bin/openaddresses2mimir.rs
TeXitoi: To output something even if all the logs are deactivated?

src/bin/openaddresses2mimir.rs
TeXitoi: To ouput something even if all the logs are desactivated?

src/bin/openaddresses2mimir.rs
TeXitoi: It would be a tricky homework for a rust newbie! @benoit-bst @nonifier any volunteer?

From l183 to l189 in a function that takes `run` by argument, in mimirsbrunn, and used in every binary.

src/bin/openaddresses2mimir.rs
TeXitoi: Never tried the backtrace, I had some difficulties to find something printing all the causes.

antoine-de: ok fair enough. but if the logs are handled by `slog` (and I think @kinnou02 was working on something like this) wouldn't this become a problem ? 
(and the way the logger is initialized It's quite difficult not to have logs no ?)

TeXitoi: @kinnou02 you chose between eprintln! and error!

kinnou02: I prefer `error!` this way we will have all log in a unified format

PR #203, opened 2018-03-22
Add comments into Mimirsbrunn

benoit-bst:


src/bin/ntfs2mimir.rs
```diff
@@ -82,6 +82,13 @@ fn to_mimir(
             name: navitia.physical_modes[pm_idx].name.clone(),
         })
         .collect();
+    let mut comments: Vec<mimir::Comment> = vec![];
```
antoine-de: I think it would be more rustic to write this with `map`/`collect`, it would then be possible to remove the  `mut`.
And one `mut` removed = one kitty saved 😻

benoit-bst: Thank @antoine-de for this tip. I am a new Rustacean. 
In addition, I love kittens...

datanel:
@TeXitoi rustfmt add `| Ok(_) => Ok(()),` (libs/docker_wrapper/src/lib.rs:93:13)
but tests fail with that

```
error: Use of a '|' at the beginning of a match arm is experimental (see issue #44101)
  --> libs/docker_wrapper/src/lib.rs:93:13
   |
93 |             | Ok(_) => Ok(()),
   |             ^

```

TeXitoi:
https://github.com/CanalTP/mimirsbrunn/pull/202/files#diff-08632cd6f9eda58d4950903756775a7cR82

Could be great to make a small reproducing example and fill a bug.

TeXitoi:
https://github.com/rust-lang-nursery/rustfmt/issues/2554

src/bin/ntfs2mimir.rs
```diff
@@ -82,7 +82,14 @@ fn to_mimir(
             name: navitia.physical_modes[pm_idx].name.clone(),
         })
         .collect();
-
+    let comments = navitia
+        .comments
+        .iter_from(&stop_area.comment_links)
+        .map(|comment| mimir::Comment {
+            id: format!("comments:{}", comment.id),
```
TeXitoi: http://vippriv-jorm-ws.mutu.prod.canaltp.prod/v1/coverage/fr-pdl/stop_points/stop_point%3ANAZ%3ASP%3A20551/?

id is not needed.

PR #204, opened 2018-04-03
Mimir: Treat admin objects with a higher level than a city as an AdminRegion

nonifier:
Administrative regions were reported with an unknown type that made Jormung unable to treat them as 'admin'. They are now reported as `AdminRegion`. Further changes to Jormung need to be done to serialize the object accordingly.


src/osm_reader/admin.rs
```diff
@@ -200,3 +207,15 @@ fn format_zip_codes(zip_codes: &[String]) -> String {
         ),
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn should_return_correct_admin_type()
```
antoine-de: you should run `rustfmt`, I think it will complain on those brackets after a new line :stuck_out_tongue_winking_eye:

src/osm_reader/admin.rs
```diff
@@ -175,6 +175,13 @@ pub fn administrative_regions(
     administrative_regions
 }
 
+pub fn get_admin_type(level: u32, city_lvl: u32) -> mimir::AdminType {
+    if level == city_lvl {
+        return mimir::AdminType::City
```
antoine-de: you can remove this `return`

antoine-de:
Ss a side note I started [working again](https://github.com/antoine-de/mimirsbrunn/tree/cosmogony2mimir) on a [cosmogony](https://github.com/osm-without-borders/cosmogony) importer and I need to extend the `AdminType`

would it be ok for you to use the [Cosmogony `ZoneType`](https://github.com/osm-without-borders/cosmogony/blob/master/src/zone.rs#L20-L29) (an `Option<ZoneType>` I guess if we want to handle the `Unkown` but I'm not sure we need this) or do you want your own type ? 

We'll need a fine grained type if we want to use [OpenCage](https://github.com/OpenCageData/address-formatting) for i18n labels

libs/mimir/src/objects.rs
```diff
@@ -287,17 +287,17 @@ impl Members for Stop {
     }
 }
 
-#[derive(Serialize, Deserialize, Debug, Clone)]
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
 pub enum AdminType {
     City,
-    Unknown,
+    AdminRegion,
```
antoine-de: even if it's a bit twisted I prefer the `Unknown` name as before an `AdminType::AdminRegion` doesn't feel very meaningful

But I think with my previous comment we could only implement `Display` for `Option<Cosmogony::ZoneType`

and not very important, but maybe  it would be better to use a custom function not the 'official `Display` trait because it's something specific to bragi.

TeXitoi: OK for me: not necessary to debate about the name of something that will be gone shortly.

src/osm_reader/admin.rs
```diff
@@ -175,6 +173,14 @@ pub fn administrative_regions(
     administrative_regions
 }
 
+pub fn get_admin_type(level: u32, city_lvl: u32) -> mimir::AdminType {
```
TeXitoi: you don't need pub here

libs/mimir/src/objects.rs
TeXitoi: No, we just let the unknown as before to not change the elasticsearch database.

Issue #206, opened 2018-04-06
Handle admin weight with cosmogony


antoine-de:
For the moment the admin weight (used to priorize the administrative regions in the search) is the number of street in an admin

If the admin comes from [cosmogony](https://github.com/osm-without-borders/cosmogony) it's more difficult to do this (since we don't know the number of street while importing the admin)

How do you think we have compute a relevant weigh now ?

One idea would be to take the `population` from osm but the tag is not often filled.

We could also use the `wikidata` id and look for the population in wikidata (or other stats), but either cosmogony or mimir will become dependent of a wikidata dump :cry: 

Do you have better ideas ?

amatissart:
Actually we could stick to the current strategy and compute the weights in osm2mimir (based on the number of streets) . All admins would then be updated into elasticsearch with their new weights.

The drawback here is that it would require osm2mimir to be the first importer to run after cosmogony. (Unless other importers such as openaddresses2mimir can implement their own weights ?)

TeXitoi:
That would be a circular dep, as osm2mimir would depends on the cosmogony admins.

I think that using the osm population would be OK, with no information setted to 0. We can assume that if the field is not provided, this admin is not so important. I think it would do the job for a first try.

antoine-de:
I don't think it creates a circular dependency, `osm2mimir` would fetch the admins in `ES` and update/publish back their weight

the problem of the population is that the information would be very scares, I fear that the results will not be very good :confused:

TeXitoi:
Ha, you mean osm2mimir would *modify* the admins? OK, why not.

TeXitoi:
The weight is not so important.

TeXitoi:
You'll have a weight for Paris and Lyon. The important thing is that the good solution is in the 3 first results. I think we need not overestimate the importance of that before having real data.

antoine-de:
the admin weight seems ok as they are.

Can we close this ?

antoine-de:
it seems everybody is ok with the closing of this issue :stuck_out_tongue_winking_eye:

Issue #177, opened 2018-01-16
Manage depedencies between imports


TeXitoi:
From now, we have OSM and Bano/OpenAddresses. OSM import admins, streets and POIs. Bano/OpenAddresses imports addresses. Addresses depends on admins, streets depends on admins, and POIs depend on admins and addresses.

Thus, we need to import OSM => Bano/OpenAddresses => OSM, that's not so clean.

I don't know how to fix that. Maybe cutting osm in admin, street and poi import (it's quite buggy now) and force the  order of imports by type and not by data.

antoine-de:
the OSM => OA dependency might be fixed by the 2nd solution of #178

antoine-de:
with cosmogony2mimir we can close this, the canonical import process is:
cosmogony -> bano/oa -> osm (only for streets and sometimes POI)

PR #207, opened 2018-04-06
add cosmogony2mimir

### LINKED ISSUE #178: Worldwide administrative regions
Hi all,

One of the biggest problem for mimir to go worldwide :world_map: is the admin handling :sob: 

For the moment mimir has a very basic admin handling based on a cli configured `admin_level` that can be configured once per `dataset`.

The [admin level](https://wiki.openstreetmap.org/wiki/Tag:boundary%3Dadministrative) can be quite complex since they are different for each countries.

## Current limitations

The current approach has several downsides:
* we cannot import the world as one, we need at least `dataset`s with countries with the same admin level
* This `city` stuff is used only for label formating, but this is a larger problem than just a city detection (in some countries we might want the `state`, or the `region` or whatever is relevant in the country

## Back to the roots, why do we need this ?

In my understanding we need the administrative regions for:
 1. disambiguation: 
    a bad term, but it's essential to handle the admins because objects (street, address, pois and even admins themselves) are unique only in their admin context.
    So it's very important to be able the search `20 rue hector malot paris` and not have the `20 rue hector malot` in lyon. 
    For the admins it's the same, the difference between [paris france](http://www.openstreetmap.org/relation/7444) and [paris us](http://www.openstreetmap.org/relation/6678712) is their admin context (at least the countries)
 2. the response formatting: 
    In mimir the `admin_level` (or admin's `type` in #176) is used for the moment only for label formatting.
    It's nice for the `20 rue hector malot` to have a label like `20 rue hector malot (Paris)` at least because the `label` must be unique in the API's response for the user to be able to choose between `20 rue hector malot paris` and `20 rue hector malot in lyon`
 3. better search:
    It's not done for the moment, but the admin hierarchy might help if one day mimir use [libpostal](https://github.com/openvenues/libpostal) as we could do structured search. I will not detail this as it's not very important and quite a long shot.

For `1.` (and even maybe `3.`) we need the administrative hierarchy, but I don't think the admin's `type` is important. For disambiguation, linking an object to all it's admins without knowing which is the city, which is the region might be enough.

For `2.` though we need to know which one is the city (and to do correct i10n, we should also know which is the `region`, which is the `country`, ..., but i10n label formatting is a story for another issue :wink:)

## What the others do

The subject is quite difficult and has been tackled quite differently other places:
* [pelias](https://github.com/pelias/pelias/) use [WhosOnFirst](https://www.whosonfirst.org/) to handle this complexity
* [osmnames](http://osmnames.org/) has quite an [interesting approach](https://github.com/OSMNames/OSMNames/pull/82), but I don't think it works everywhere (especially in Germany with their [twisted admin hierarchy](https://de.wikipedia.org/wiki/Datei:Administrative_Gliederung_Deutschlands.svg) )
* [libpostal](https://github.com/openvenues/libpostal) has a [lot of rules](https://github.com/openvenues/libpostal/blob/master/resources/boundaries/osm/fr.yaml)

### Keep the cli based osm2mimir configuration

Instead of the `city_level`, we could give a more complex configuration associating a `admin_level` to an `AdminType`.

:white_check_mark: easy to implement
❌ impossible to make a world import, we need to split into dataset with the same configuration
❌ hard to maintain all the configuration

### Export the complexity to another place

A bit like the [osmnames's export](http://osmnames.org/download/) or [WhosOnFirst](https://www.whosonfirst.org/) approach, we could make another project that aim is to eat an osm file (and maybe something like [the countries shapes](https://raw.githubusercontent.com/OSMNames/OSMNames/master/data/sql/country_osm_grid.sql)) and output a file (csv ?) with all the admins we want, correctly typed (and maybe even linked to each others).

The implementation of such project is left as an exercise to the reader :laughing: but as for the approach below I think something based on [libpostal's rules](https://github.com/openvenues/libpostal/blob/master/resources/boundaries/osm/) is the best way to go, even if it might not be up to date.

:white_check_mark: hide the complexity outside of mimir
:white_check_mark: the output file could be useful to other projects (having a structured dataset with admins hierarchy could be quite nice !)
:white_check_mark: could make the mimir import pipeline a bit more straightforward since we should import *first* the admins, but then we could import either OSM or bano/OA in any order
❌ Would be a bit difficult to keep close to the `dataset` notion, or we would to have a way to partition the output file for example to extract the admins for a country, a continent, ... But anyway, I'm not sure the `dataset` notion scale well
❌ not trivial to implement

### More clever osm2mimir

osm2mimir could use [libpostal's rules](https://github.com/openvenues/libpostal/blob/master/resources/boundaries/osm/). The big problem of this approach is that we need to know the country of an object, so for smaller than country import, we'll need to either give the country in cli, or use something like a [the countries shapes database](https://raw.githubusercontent.com/OSMNames/OSMNames/master/data/sql/country_osm_grid.sql) to be able to know if an object is in a country.

:white_check_mark: fit more in the mimir architecture
:white_check_mark: less executable to import the data, so easier to use
❌ would be mimir centric, cannot be useful to other projects
❌ not trivial to implement

## Conclusion
What do you think ?
Do you see stuff I have missed ? which solution seems the best for you ?

I hope we can work together toward the greatest world wide :world_map:  open source geocoder :sparkling_heart:

### LINKED ISSUE #189: Handle osm2mimir CLI parameters correctly
Hi,

for the moment:
* `--import-admin` is useless
* less important but I feel that `--import-street` is mandatory (or at least should be defaulted to `true`)

The second point is just an ergonomic issue, but the first one is a pain since I'd like to be able to *not* import the admin (and use the new cosmogony for it), but fixing the parameter handling will break some integration (people forgetting to pass `--import-admin` won't have anything anymore.

What is the right way to handle this ?

Clap bool argument only seems to work as flag, we can't say `--import-admins=false`, so should we reverse and add `--no-*` parameters ?
or move the string parameters and check `true`/`false`

can we change the default ?

antoine-de:
add a [cosmogony](https://github.com/osm-without-borders/cosmogony) importer.

# what do it do
should close #178 (and #177 as a side effect), with this mimir is a step closer to go worlwide :tada: 

This changes the mimir import pipeline as now we can import the administrative regions through cosmogony (we can still directly import them from osm though)

:warning: this PR breaks the cli retro compatibility as the parameter `--import-admin` does not default to true anymore (closes #189 ) :warning: 

# TODOs
This PR is ready for review but I still need to do:
* [x] as integration test combining cosmogony/osm/OA and bragi
* [x] find a clever way to compute an Admin weight (linked to #206 )
* [x] use cosmogony::ZoneType instead of `AdminType` without breaking ES
* [x] update cosmogony when it's available on crates.io

# Admin associations when cosmogony is used
For the adresses, streets, pois and stops, all the admins from Cosmogony will be used. We do not use Cosmogony hierarchy to do so, but check by geometric inclusion(behaviour unchanged)

for example in the test we have a poi (the melun's cityhall):

```javascript
{
  "id": "poi:osm:way:112361498",
  "type": "poi",
  "label": "Hôtel de Ville (Melun)",
  "name": "Hôtel de Ville",
  "postcode": "77000",
  "city": "Melun",
  "citycode": "77288",
  "administrative_regions": [ // <-- we can find the associated admins of the poi
    {
      "id": "admin:osm:relation:80071",
      "insee": "77288",
      "level": 8,
      "label": "Melun (77000-CP77001), Fausse Seine-et-Marne, France hexagonale",
      "name": "Melun",
      "zip_codes": [
        "77000",
        "77003",
        "77008",
        "CP77001"
      ],
      "weight": 0.14285714285714285,
      "coord": {
        "lon": 2.6608169,
        "lat": 48.539927
      },
      "zone_type": "city"
    },
    {
      "id": "admin:osm:relation:424253843",
      "insee": "77",
      "level": 6,
      "label": "Fausse Seine-et-Marne, France hexagonale",
      "name": "Fausse Seine-et-Marne",
      "zip_codes": [],
      "weight": 1,
      "coord": {
        "lon": 2.7190708886837056,
        "lat": 48.52116192196974
      },
      "zone_type": "state_district"
    },
    {
      "id": "admin:osm:relation:424256272",
      "insee": "",
      "level": 2,
      "label": "France hexagonale",
      "name": "France hexagonale",
      "zip_codes": [],
      "weight": 1,
      "coord": {
        "lon": 2.6936051723115395,
        "lat": 46.65468308592214
      },
      "zone_type": "country"
    }
  ],

```
The admin do not have associated admin (behavior unchanged). So Paris is not associated to France in the response (but the query paris france will match the Paris object because of the paris label: `Paris (75000-75116), Île-de-France, France`

```javascript
{
  "id": "admin:osm:relation:80071",
  "type": "city",
  "label": "Melun (77000-CP77001), Fausse Seine-et-Marne, France hexagonale",
  "name": "Melun",
  "postcode": "77000;77003;77008;CP77001",
  "city": null, // <--- the city of a city is null :/
  "citycode": "77288",
  "level": 8,
  "administrative_regions": [] // <-- there are no associated admins to an admin, we just have the label to know the state/country/...
}
```

# Note
Note: As a side note @nlehuby changed the osm file to be able to have some countries.
the dataset now looks like:
awesome_hexagone png


antoine-de:
I added a complex integration test to show the api's response and changed the `AdminType` to the cosmogony `ZoneType` (to have `State`, `Country`,...)

I think this PR is ready for review.

I still have some remark on it:

1. I added a new field in the Admins, `ZoneType` that deprecates `AdminType` 
I added a new field because I did not want to do serde magic not to break retro compatibility (but I can if you insist).
This field will need to be cleaned afterward.
2. I don't like the fact that a `GeocodingResponse` contains directly a `mimir::Admin` ([here](https://github.com/CanalTP/mimirsbrunn/blob/master/libs/bragi/src/model.rs#L78))
This makes it more difficult to control which fields are serialized. eg for examples we output for the moment the `admin_type` AND the `zone_type` in the associated admins. I think that need to be changed in this PR since we don't want to have to maintain the retrocompatibility on the `admin_type` field.
I think we should add another type (`AssociatedAdmin` ?) with the fields we want to output
3. for the moment the link betweens admins are quite poor.
`Melun` city is not associated to it's state or it's country appart from in it's label. I don't think it's a problem but I think it's important enough to be highlighted.
What can be more problematic is thus that the field `city` of an admin is never filled ([melun's city is null](https://github.com/antoine-de/mimirsbrunn/blob/90162bb3378c1927f4cd007b36f6ba42d1124994/tests/canonical_import_process_test.rs#L98)
it might become even more important since cosmogony can import city district and even not official distrcit (like [le marais](https://www.openstreetmap.org/relation/6438571#map=15/48.8582/2.3612) that might be nice to associate to a city)
what do you think ?

tests/cosmogony2mimir_test.rs
```diff
@@ -0,0 +1,126 @@
...
+    assert!(all_objects.iter().any(|r| r.is_admin()));
+    // all cosmogony admins have boundaries
+    assert!(all_objects.iter().all(|r| r.admins()[0].boundary.is_some()));
+    assert!(all_objects.iter().all(|r| r.admins()[0].coord.is_valid()));
+
+    // check Livry-sur-Seine
+    let res: Vec<_> = es_wrapper
+        .search_and_filter("label:Livry-sur-Seine", |_| true)
+        .collect();
+    assert!(res.len() >= 1);
```
TeXitoi:
```rust
assert_eq!(res.len(), 1);
```

TeXitoi: The thing is that you then test on res[0], thus if you have more than 1 result, that may not be the one you want to test.

antoine-de: It's just that I want to enforce that the first to be returned has certain properties but I don't care about the others no?

Honestly I can change it without problem it's just that I find those integration tests difficult to maintain and I prefer to only test the important stuff

libs/bragi/src/model.rs
```diff
@@ -163,6 +163,13 @@ impl From<mimir::Admin> for GeocodingResponse {
     }
 }
 
+fn get_admin_type(adm: &mimir::Admin) -> String {
+    match adm.zone_type {
+        Some(t) => format!("{:?}", t).to_snake_case(),
```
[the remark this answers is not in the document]
TeXitoi: I like strum, but sadly they don't provide rename_all functionality.

src/bin/cosmogony2mimir.rs
```diff
@@ -0,0 +1,126 @@
...
+trait IntoAdmin {
+    fn into_admin(self) -> Admin;
+}
+
+impl IntoAdmin for Zone {
+    fn into_admin(self) -> Admin {
+        let insee = admin::read_insee(&self.tags).unwrap_or("");
+        let zip_codes = admin::read_zip_codes(&self.tags);
+        let label = self.label;
+        let weight = Cell::new(0.); //TODO, what do we want ?
```
TeXitoi: Let's use the OSM population value here, we may change that later, but having something for the big cities now is better than just always 0.

antoine-de: ok, I changed this, I'll post in the concerned issue why it's not perfect though

tests/cosmogony2mimir_test.rs
```diff
@@ -72,7 +72,7 @@ pub fn cosmogony2mimir_test(es_wrapper: ::ElasticSearchWrapper) {
             assert_eq!(livry_sur_seine.insee, "77255");
             assert_eq!(livry_sur_seine.level, 8);
             assert_eq!(livry_sur_seine.zip_codes, vec!["77000"]);
-            assert_eq!(livry_sur_seine.weight.get(), 0f64);
+            assert_eq!(livry_sur_seine.weight.get(), 1919f64);
```
TeXitoi: weights must be between 0 and 1

TeXitoi: https://github.com/CanalTP/mimirsbrunn/blob/master/src/osm_reader/admin.rs#L184-L196

antoine-de: oups my bad, forgot about this, it's fixed!

PR #209, opened 2018-04-11
fix the cargo.lock

antoine-de:
I think the cargo.lock had been manually edidted, so removing it and
building was not working because of the iron logger dependency.

logger depends on iron 6 and we depend on iron 4 and this was causing
conflicts.

did you need the git dependency for something special ?

I also removed the slog from the docker_wrapper.
I must admit that this was the easy way out, I had some strange built error 
(
```
114 |                 warn!("docker {a:?} failed", a = args)
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `slog::Value` is not implemented for `[&str]`
    |
    = note: required because of the requirements on the impl of `slog::Value` for `&[&str]`
    = note: required by `slog::SingleKV`
```
)
and since I don't see the point of using slog for the integration tests, I just removed it instead of understanding the problem :man_facepalming: 

if that's important for you I can dig on the problem.

As a side note, the navitia-model git dependency prevent the use of `git bisect` :cry: 
wouldn't it better to use a pinned version (either on crates.io or a tag) ?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -20,14 +20,17 @@ version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "approx 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "num-complex 0.1.42 (registry+https://github.com/rust-lang/crates.io-index)",
+ "num-complex 0.1.43 (registry+https://github.com/rust-lang/crates.io-index)",
  "num-traits 0.1.43 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
 name = "ansi_term"
-version = "0.10.2"
+version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "winapi 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
+]
 
 [[package]]
 name = "approx"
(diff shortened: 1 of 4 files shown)


TeXitoi:
The file is not edited by hand, but the deps are updated one by one.

logger 0.4 don't depend on ion 0.6, it depend on  iron >= 0.4, < 0.7.0. When you do a cargo update, it will import the latest ion version on logger, but we need logger to be compiled with ion 0.4 to work with rustless. not updating ion let the good depedency. We need logger 0.4 for a working slog.

I can't find a way to force ion 0.4 to be the only ion version.

I can't find why we use a git version on logger as seen in #194 @kinnou02 do you remember why? It seems to work well with 0.4 and not using the git repo.

antoine-de:
arf logger 0.2 breaks slog ? no unit tests break so there could be no breakage no :rofl: 

else maybe the good option would be to update rustless for iron 0.6, but it's a shame to fix a dead body :cry:

antoine-de:
Note: the build fail because of a `fmt`  update, I'll run it if you want this PR (hopefuly not creating conflicts with #207 :roll_eyes: )

TeXitoi:
```
cargo update
cargo update -p iron:0.6.0 --precise 0.4.0
```
make that working. Using logger 0.4 seems the good thing to do. I don't have any better solution, and using logger 0.2 is a good idea, using zombie code is enough with rustless.

No way to touch to rustless, using hyper directly would be better than that.

antoine-de:
what is the status on this pull request ? do we close it ?

it's a bit a pain not to be able to import the docker wrapper externally (but it's not blocking on our side)

TeXitoi:
I don't get it, docker_wrapper doesn't depend on iron.

antoine-de:
yes I think it's a separate problem, there seems to be a problem with the slog version in the wrapper when regenerating the cargo.lock

TeXitoi:
```
cargo update -p slog --precise 2.1.1
```
Fix the problem. I suspect slog 2.2 broke semver.

TeXitoi:
https://github.com/slog-rs/slog/issues/166

TeXitoi:
So the solutions are:
 * pin to slog 2.1.1
 * stop using named args for things that doesn't implement Display

In this case, the second choice seems to be the right one as we don't really care of the semantic thing in this case.

TeXitoi:
https://github.com/CanalTP/mimirsbrunn/pull/210

@antoine-de I think we can close this if you are OK.

antoine-de:
yes seems fine for me, thanks!

PR #210, opened 2018-04-25
Cargo update

TeXitoi:


antoine-de:
you need to rebase the PR, there is a small cargo.lock conflict

PR #211, opened 2018-05-15
Changes in ES query

antoine-de:
We made 2 changes in the ES query:

#### add boost on name 
the label can contains additional information (especially for admin from cosmogony), thus we provide an additional boost if we match the name for a better priorization
#### a bit of black magic to tune the weight on the type boosting
we feel that the admins are not enough prioritized compared to other objects (the problem is especially bad for pois as pois's label are quite diverse and contain the admin's name)
the values have been manually changes to fine better results on some cases and geocodeur-tester have been run before/after

### Tests
Note: it's not exactly [geocoder-tester](https://github.com/geocoders/geocoder-tester) it's [our fork ](https://github.com/QwantResearch/geocoder-tester) with a bit more tests (especially on pois and admins)

the tests are run only for france (as it's what is loaded in our instance):
* fuzzy: all the tests with `fuzzy` in the file name, made to test the resistance to errors
* admins: [only admins](https://github.com/QwantResearch/geocoder-tester/blob/master/geocoder_tester/world/france/test_admin.csv)
* poi: [pois](https://github.com/QwantResearch/geocoder-tester/blob/master/geocoder_tester/world/france/test_poi.csv) and [pois](https://github.com/QwantResearch/geocoder-tester/blob/master/geocoder_tester/world/france/iledefrance/test_pois.yml)
* main: all the other tests in the france directory :stuck_out_tongue_winking_eye: 

test | main | fuzzy | admins | pois
-----|---------|---------|----------|-----
before|1978 failed, 9038 passed (1085.21)|305 failed, 647 passed (202.50 s)|35 failed, 15 passed,(5.16 s)|29 failed, 35 passed (8.77 s)
after|1951 failed, 9065 passed, 1 skipped (1038.61 s)|228 failed, 724 passed (210.15 s)|15 failed, 35 passed (4.48 s)|28 failed, 36 passed, (8.94 s)

the overall quality is a tad better and the admins are better matched so I think it's ok for us.

If you're interested in the PR, I think it would be nice to test in also on your dataset as those kind of changes can have unexpected side effects. :roll_eyes:

### DIFF
--- a/libs/bragi/src/query.rs
+++ b/libs/bragi/src/query.rs
@@ -143,13 +143,13 @@ fn build_query(
     }
     let type_query = Query::build_bool()
         .with_should(vec![
-            match_type_with_boost::<Addr>(12.),
-            match_type_with_boost::<Admin>(11.),
-            match_type_with_boost::<Stop>(10.),
-            match_type_with_boost::<Poi>(2.),
+            match_type_with_boost::<Addr>(20.),
+            match_type_with_boost::<Admin>(19.),
+            match_type_with_boost::<Stop>(18.),
+            match_type_with_boost::<Poi>(1.5),
             match_type_with_boost::<Street>(1.),
         ])
-        .with_boost(20.)
+        .with_boost(30.)
         .build();
 
     // Priorization by query string
--- a/libs/bragi/src/query.rs
+++ b/libs/bragi/src/query.rs
@@ -154,6 +154,7 @@ fn build_query(
 
     // Priorization by query string
     let mut string_should = vec![
+        Query::build_match("name", q).with_boost(1.).build(),
         Query::build_match("label", q).with_boost(1.).build(),
         Query::build_match("label.prefix", q).with_boost(1.).build(),
         Query::build_match("zip_codes", q).with_boost(1.).build(),


kinnou02:
Master: 2111 failed, 9761 passed, 7 skipped in 371.45 seconds                                                                                                                                                  PR: 2163 failed, 9709 passed, 7 skipped in 378.34 seconds
A bit more failures :(

kinnou02:
I tried to apply #213, but it's has only farther decrease the number of success:  
```
2172 failed, 9700 passed, 7 skipped in 399.06 seconds
```

libs/bragi/src/query.rs
```diff
@@ -143,17 +143,18 @@ fn build_query(
...
+            match_type_with_boost::<Poi>(1.5),
             match_type_with_boost::<Street>(1.),
         ])
-        .with_boost(20.)
+        .with_boost(30.)
         .build();
 
     // Priorization by query string
     let mut string_should = vec![
+        Query::build_match("name", q).with_boost(1.).build(),
```
TeXitoi: I don't get it, you have something in the name that is not in the label?

TeXitoi: you mean more weight on the name part?

antoine-de: yes the name get double points.

it's especially important now that cosmogony create quite a big label for the admins

nlehuby:
I think that the regression can be explained by the fact that the addresses do not have an indexed `name` field either. We have many tests about addresses with house numbers, and now the streets are more boosted than the addresses (using both PR #211 and #212)

We need to work a little more on this and also add a boost on the name for the address objects.

kinnou02:
Also, there is no prefix nor ngram on the name, as most our new failures are fuzzy queries it might help to add these.

antoine-de:
close, superseded by #229

Issue #212, opened 2018-05-16
rename Street's `street_name` to `name`


antoine-de:
for the moment the `Street`s have a field `street_name` whereas other objects have a `name` attribute.

the field being separated means that #211 cannot boost the query.

this results in subperformant query.

for example the query
`"Rue de Bouxwiller Bouxwiller"` returns all the streets of the `Bouxwiller` city and the street `Bouxwiller` of other cities (with or without #211, the 2 first results on navitia are `Rue de Bouxwiller (Strasbourg)` and `Rue de Niedersoultzbach (Bouxwiller)`).

renaming the field `street_name` to `name` fixes this, but It's a breaking change.

Is it fine for you if we duplicate the field for the migration ? do you see a better way to do this ? is it fine for you ?

kinnou02:
It's fine to duplicate field for migration, it's the recommended way to do it.

PR #213, opened 2018-05-16
rename street_name to name

antoine-de:
fixes https://github.com/CanalTP/mimirsbrunn/issues/212

rename Street's `street_name` to `name`

since it's a breaking change, for the moment the fields are dupplicated,
and `street_name` is deprecated

Note: the `street_name` field has been removed from the mapping as we don't care how ES indexes it


antoine-de:
closed, superseded by  #229

PR #217, opened 2018-05-17
Add libgeos to dockerfile

pbougue:
libgeos is required now for bragi to run
also a bit of refactor on CMD and dockerignore pattern

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,3 @@
 *
-!target/*/bragi
+!target/release/bragi
 !build_packages/*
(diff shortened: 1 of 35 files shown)


antoine-de:
I think you should update you rustfmt 😜 

why is libgeos needed ?

pbougue:
I'll check tomorrow, but it's the write-mode option that's not recognized on rustfmt...

pbougue:
@antoine-de libgeos is dragged by cosmogony ;)

PR #222, opened 2018-05-27
POC on actix

TeXitoi:


libs/bragi/src/actix.rs
```diff
@@ -0,0 +1,41 @@
...
+use std::sync::Arc;
+
+// TODO: pretty errors, async es
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+struct Params {
+    q: String,
+}
+
+fn autocomplete(req: HttpRequest<Arc<::Args>>) -> Result<Json<model::v1::AutocompleteResponse>> {
```
antoine-de: why can't you use directly `Query<Params>` ?

edit: ahh to be able to get the state ?
if it's for this I think you can have both params a bit like [this](https://github.com/actix/examples/blob/master/diesel/src/main.rs#L42)

antoine-de:
superseeded by #286 (started from the work on this PR)

PR #224, opened 2018-06-05
Thread safety

### LINKED ISSUE #218: Make mimir objects thread-safe
One more issue on our path to let mimirsbrunn go worldwide :earth_africa: !  
We did some experiments to optimize large imports using multiple threads.

As far as I can see, there are currently two blocking issues to make that possible:
 * `<Rc<mimir::Admin>>`, that could be replaced with `<Arc<mimir::Admin>>` in all structs. What are the drawbacks of such a change?
 * Admin.weight uses `Cell<f64>` to update weights during some imports. I am not sure about this one. But something like `Arc<RwLock<f64>>` could do the trick, unless we find another way to handle weights and avoid interior mutability when it's not necessary?

Here is an implementation of these changes (usable, but needs a better error handling):
https://github.com/CanalTP/mimirsbrunn/compare/master...amatissart:threadSafe

As an example, I tried to run an importer (fetching rows from postgres) on multiple threads using [**rayon**](https://github.com/rayon-rs/rayon). 
See https://github.com/amatissart/fafnir/blob/multithread/src/lib.rs#L167
That works, but is still a bit cumbersome: [`par_chunks`](https://docs.rs/rayon/1.0.1/rayon/slice/trait.ParallelSlice.html#method.par_chunks) is used to initialize `Rubber` in each thread. Maybe there is a better way.

In the end, that could help to optimize all importers such as "osm2mimir", that reads data in parallel, and then imports objects in a single thread.

What do you think ? Do I overlook something?

antoine-de:
Makes mimir thread-safe.

close #218 

This does not multithread mimir as is, it just makes the object model thread-safe for a future PR (the first step for us is to use it in [fafnir](https://github.com/QwantResearch/fafnir), the PR is [here](https://github.com/QwantResearch/fafnir/pull/13))

Now, there is quite a big change for the object's weight. Following the discussion on #218 as added an abstract layer to be able to store the weight with an `AtomicUSize`. I'm not happy with the current implementation, I find it way to complex for what it does. Feel free to propose a better implementation.


libs/mimir/src/weight.rs
```diff
@@ -0,0 +1,174 @@
...
+    pub fn new(val: f64) -> Weight {
+        Weight {
+            val: AtomicUsize::new(val as usize),
+            normalized: AtomicBool::new(false),
+        }
+    }
+
+    fn from_normalized_float(val: f64) -> Weight {
+        Weight {
+            val: AtomicUsize::new((val * MAX_VAL) as usize),
```
kinnou02: You might want to assert than val is between 0 and 1

src/osm_reader/admin.rs
```diff
@@ -177,13 +177,14 @@ pub fn compute_admin_weight(streets: &StreetsVec, admins_geofinder: &AdminGeoFin
     let mut max = 1.;
     for st in streets {
         for admin in &st.administrative_regions {
-            admin.weight.set(admin.weight.get() + 1.);
-            max = f64::max(max, admin.weight.get());
+            let w = admin.weight.unnormalized_value();
+            //PR: removal of the +1, question this in the PR
```
antoine-de: Note: I removed the +1 here, I can't remember why we need it

src/osm_reader/poi.rs
```diff
@@ -282,7 +282,8 @@ pub fn compute_poi_weight(pois_vec: &mut [Poi]) {
     for poi in pois_vec {
         for admin in &mut poi.administrative_regions {
             if admin.is_city() {
-                poi.weight = admin.weight.get();
+                // Note: we can unwrap there because the admin should always have a computed weight there
+                poi.weight = admin.weight.value().unwrap();
```
TeXitoi: I don't like this unwrap, but no strong opinion for the moment. Anyone having an opinion on that?

src/osm_reader/street.rs
```diff
@@ -195,7 +195,8 @@ pub fn compute_street_weight(streets: &mut StreetsVec) {
     for st in streets {
         for admin in &mut st.administrative_regions {
             if admin.is_city() {
-                st.weight = admin.weight.get();
+                // Note: we can unwrap there because the admin should always have a computed weight there
+                st.weight = admin.weight.value().unwrap();
```
TeXitoi: Same here.

tests/rubber_test.rs
```diff
@@ -148,7 +148,7 @@ pub fn rubber_custom_id(mut es: ::ElasticSearchWrapper) {
         name: "my admin".to_string(),
         label: "my admin (zip_code)".to_string(),
         zip_codes: vec!["zip_code".to_string()],
-        weight: Cell::new(0.42),
+        weight: Weight::new(42u32),
```
TeXitoi: Or you can Weight::from_normalized_value?

tests/rubber_test.rs
```diff
@@ -238,12 +239,13 @@ pub fn rubber_ghost_index_cleanup(mut es: ::ElasticSearchWrapper) {
...
```
TeXitoi: Same here

src/osm_reader/admin.rs
antoine-de: oups 🤦‍♂️

libs/mimir/src/weight.rs
```diff
@@ -0,0 +1,177 @@
...
+    }
+}
+
+impl Serialize for Weight {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        // unwrap() here to ensure that the weight is always normalized before serialization
+        serializer.serialize_f64(self.value().unwrap())
```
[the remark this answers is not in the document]
antoine-de: it's the same for there and below, I would have prefered to have some stronger type around the weight (Normalized/Unormalized), but since it's deeply nested in the model I don't see a nice way to do it 😕 

Thus I prefered to use `unwrap` everywhere I would use a NormalizedWeight to ensure that there was no silent errors. It was at least pretty usefull during development, but If you prefer I can return an Err here an `unwrap_or` everywhere

tests/rubber_test.rs
antoine-de: the method is private since I wanted to hide a bit this complexity.

you think it's better to make it public than to have to call `normalize` ?

antoine-de:
I removed the weight struct since the Admins can now be immutable (thanks to #227)

The code is thus way cleaner!

antoine-de:
did you had time to review this ?

PR #226, opened 2018-06-11
Address must have street name

jbgriesner:
Some addresses have a void street.name field.
They shouldn't be imported since they decrease the reverse geocoding quality and furthermore are not 'searchable' in the autocomplete.

This PR filters these addresses + adds the corresponding tests.

PS: commit 4998fcd32ddfd95269d25f1eb06f6658bc3b5bc0 fails the `cargo fmt --all -- --check` so I format it the correct way

### DIFF
--- a/src/bin/bano2mimir.rs
+++ b/src/bin/bano2mimir.rs
@@ -156,13 +156,14 @@ where
         info!("importing {:?}...", &f);
         let mut rdr = csv::ReaderBuilder::new().has_headers(false).from_path(&f)?;
 
-        let iter = rdr.deserialize().filter_map(|r| {
-            r.map_err(|e| info!("impossible to read line, error: {}", e))
-                .ok()
-                .map(|b: Bano|
-                    b.into_addr(&admins_by_insee, &admins_geofinder)
-                )
-        }).filter(|a| !a.street.street_name.is_empty());
+        let iter = rdr
+            .deserialize()
+            .filter_map(|r| {
+                r.map_err(|e| info!("impossible to read line, error: {}", e))
+                    .ok()
+                    .map(|b: Bano| b.into_addr(&admins_by_insee, &admins_geofinder))
+            })
+            .filter(|a| !a.street.street_name.is_empty());
         let nb = rubber
             .bulk_index(&addr_index, iter)
             .with_context(|_| format!("failed to bulk insert file {:?}", &f))?;
--- a/src/bin/bano2mimir.rs
+++ b/src/bin/bano2mimir.rs
@@ -43,10 +43,10 @@ extern crate slog_scope;
 extern crate structopt;
 
 use failure::ResultExt;
+use mimir::MimirObject;
 use mimir::objects::Admin;
 use mimir::rubber::Rubber;
 use mimirsbrunn::admin_geofinder::AdminGeoFinder;
-use std::cell::Cell;
 use std::collections::BTreeMap;
 use std::fs;
 use std::path::PathBuf;
@@ -156,7 +156,6 @@ where
     for f in files {
         info!("importing {:?}...", &f);
         let mut rdr = csv::ReaderBuilder::new().has_headers(false).from_path(&f)?;
-        let skipped_addresses = Cell::new(0);
         let iter = rdr
             .deserialize()
             .filter_map(|r| {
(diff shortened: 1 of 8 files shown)


tests/fixtures/sample-bano.csv
```diff
@@ -30,3 +30,5 @@
 751124517P-8Z,8Z,Rue Hector Malot,75012,Paris,O+O,48.845692,2.375752
 751124517P-9,9,Rue Hector Malot,75012,Paris,OSM,48.845801,2.375630
 420424517P-20,20,Rue Hector Malot,42042,Trifouilli-les-Oies,OSM,50.0,2.0
+420424517P-20,20,,42042,Fake-City,OSM,50.0,2.0
```
antoine-de: you should change the id to be sure it's not just dedupplication

tests/bano2mimir_test.rs
```diff
@@ -33,6 +33,12 @@ use hyper;
 use hyper::client::Client;
 use mdo::option::{bind, ret};
 
+/// Returns the total number of results in the ES
+fn get_total(es_wrapper: &::ElasticSearchWrapper) -> u64 {
```
antoine-de: this function name is way too generic.

maybe `get_nb_elements` ?

tests/fixtures/sample-bano.csv
jbgriesner: done.

tests/bano2mimir_test.rs
jbgriesner: done

TeXitoi:
Does it happen in BANO? Because it should not, and I prefer a failure that a silent skip.

TeXitoi:
Can I have some samples of the data generating the problem?

jbgriesner:
Sure, here are two samples:

| Query X (/autocomplete?q="X") | Address street | Label |
| :---: | :---: | :---: | 
| Square Bénés (Saint-Laurent-du-Var) | "  " | "990  (Saint-Laurent-du-Var)" |
| caisse d'épargne saint laurent du var |  "   " | "35  (Saint-Laurent-du-Var)" |

As you can see we get only the street number (and the city name) in the label without the street name, which is not great.

We could maybe log the skipped streets ?

jbgriesner:
PS: here are the BANO entries corresponding to the two samples above:

061230396B-990,990,,06700,Saint-Laurent-du-Var,O+O,43.671073,7.190625

061230040P-35,35,,06700,Saint-Laurent-du-Var,O+O,43.673543,7.190509
061230172H-35,35,,06700,Saint-Laurent-du-Var,O+O,43.696871,7.183807
061230405L-35,35,,06700,Saint-Laurent-du-Var,O+O,43.660001,7.181865
061230574V-35,35,,06700,Saint-Laurent-du-Var,O+O,43.663139,7.188011
061230640S-35,35,,06700,Saint-Laurent-du-Var,O+O,43.669865,7.191818
061230741B-35,35,,06700,Saint-Laurent-du-Var,O+O,43.669710,7.181086
061230970A-35,35,,06700,Saint-Laurent-du-Var,O+O,43.675786,7.188957

nlehuby:
These addresses in BANO are from an opendata source. This may be fixable by adding to OSM every address where the opendata source has no street name.
It seems quite time-consuming and tedious, but if you want to fix the BANO data, feel free to do it :yum:

prhod:
Agreed with @nlehuby: the best way to fix those would be to add the addresses in OSM, but the number of errors may be ... interesting.
I think skipping them is a good idea, with a warning indicating the the count of ignored values.

jbgriesner:
Following @prhod suggestion, I added a warning of ignored addresses.

jbgriesner:
But what do we prefer:
1. a counter of the number of ignored addresses ?
2. or just a log each time an address is skipped ?
Note that problems will occur with solution 1 in case of multi threading.

src/bin/bano2mimir.rs
```diff
@@ -166,19 +165,17 @@ where
             })
             .filter(|a| {
                 !a.street.street_name.is_empty() || {
-                    skipped_addresses.set(skipped_addresses.get() + 1);
+                    info!(
+                        "Address {}:{:?}:{:?} has no street name and has been ignored.",
```
antoine-de: the street will not add any meaningfull information no ?

jbgriesner: It contains the city name, the zip code, the admins which I thought could save debug time.

antoine-de:
did you had time to review this PR ?

antoine-de:
If it's approved, can you merge it ? :roll_eyes: :smile:

Issue #230, opened 2018-06-19
filtering problem


cservan:
Hello,
it seems the response filtering by using the keyword "type" does not work. e.g.: 
`http://localhost/autocomplete?q=Paris&type=city`
give as first answer the state_district of Paris, not the city of Paris.

Best.

antoine-de:
It's not really a bug, it's more a bad api problem and a missing feature:

the `type` parameter in the API can filter the main `mimir` types:
* Street
* Address
* Poi
* Stops
* Administrative regions

for historical reason the filter parameter on the administrative regions is `city` (bad design, but at first there were only cities :stuck_out_tongue_winking_eye:  ) thus leading to this misunderstanding.

We can't stop supporting this because of retro-compatibility, but we should add a filter value on `administrative_region` (or maybe `zone` ?) (and remove `city` from the non existent documentation :rofl: ).

We should also add another parameter (`zone_type` ?) to be able to ask only for cities, or countries.

jbgriesner:
@cservan : we are adding the possibility to filter more explicitly the "type of zones" you are looking for as follows : 
* only cities: <bragi IP>/autocomplete?q=Paris&zone_type=city" 
* only cities: <bragi IP>/autocomplete?q=Paris&zone_type=state_district" 

The possible variants for this parameter are : 
* suburb
* city_district                                                                      
* city
* state_district
* state
* country_region
* country
* non_administrative

For more details about the changes just check PR https://github.com/CanalTP/mimirsbrunn/pull/235

antoine-de:
superseeded by #243

PR #235, opened 2018-06-26
Zone_type filter in Bragi

jbgriesner:
This PR handles the issue https://github.com/CanalTP/mimirsbrunn/issues/230

As suggested by @antoine-de the PR proposes 2 things:
* add "zone" as a match variant for the type parameter which is more explicit than "city" e.g. : 
"/autocomplete?q=Paris&type=zone"
* add a new parameter "zone_type" to filter the admins, e.g. : 
"/autocomplete?q=Paris&zone_type=city"

Some tests come along the changes.


libs/bragi/src/query.rs
```diff
@@ -132,6 +132,7 @@ fn build_query(
     shape: Option<Vec<rs_es::units::Location>>,
     pt_datasets: &[&str],
     all_data: bool,
+    zone_type: &str,
```
antoine-de: an Option would be better

libs/bragi/src/api.rs
```diff
@@ -351,6 +357,11 @@ impl ApiEndPoint {
                     });
 
                     let types = get_param_array(params, "type");
+                    let zone_type = params
+                        .find("zone_type")
```
antoine-de: using an enum like [Type](https://github.com/CanalTP/mimirsbrunn/blob/master/libs/bragi/src/params.rs#L255) would help early feedback to the api user

(should we use the CosmogonyType ? or maybe a separate enum would add additional flexibility)

libs/bragi/src/params.rs
```diff
@@ -265,6 +265,7 @@ impl FromStr for Type {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s {
             "city" => Ok(Type::City),
+            "zone" => Ok(Type::City),
```
antoine-de: renaming the `Type::City` would be nice too

tests/cosmogony2mimir_test.rs
```diff
@@ -123,4 +126,20 @@ pub fn cosmogony2mimir_test(es_wrapper: ::ElasticSearchWrapper) {
...
+        bragi.get("/autocomplete?q=Saint-Martin-d'Hères&zone_type=state_district");
+
+    let types = get_types(&response_without_filter);
+    let types_filtered = get_types(&response_with_filter);
+
+    assert_eq!(count_types(&types, "city"), 1);
+    assert_eq!(count_types(&types, "state_district"), 1);
+
+    assert_eq!(count_types(&types_filtered, "city"), 0);
+    assert_eq!(count_types(&types_filtered, "state_district"), 1);
```
antoine-de: you should also add a test of `type=zone`

libs/bragi/src/params.rs
```diff
@@ -273,3 +305,22 @@ impl FromStr for Type {
         }
     }
 }
+
+struct ZoneTypeWrapped(ZoneType);
```
jbgriesner: Wrapper to avoid the error : "only traits defined in the current crate can be implemented for arbitrary types"

tests/cosmogony2mimir_test.rs
```diff
@@ -123,4 +127,39 @@ pub fn cosmogony2mimir_test(es_wrapper: ::ElasticSearchWrapper) {
...
+
+    let response_with_filter_state_district =
+        bragi.get("/autocomplete?q=Saint-Martin-d'Hères&zone_type=state_district");
+    let types_filtered = get_types(&response_with_filter_state_district);
+    assert_eq!(count_types(&types_filtered, "city"), 0);
+    assert_eq!(count_types(&types_filtered, "state_district"), 1);
+
+    let response_zone = bragi.get("/autocomplete?q=Saint-Martin-d'Hères&type=zone");
+    let types_zone = get_types(&response_zone);
+    assert_eq!(count_types(&types_zone, "street"), 0);
```
antoine-de: did the test without the `type` filter had any street/addr/poi ? if so, you should add an  assert on the line below to make it explicit, if not, you should change the query for this test :stuck_out_tongue_winking_eye:

libs/bragi/src/params.rs
```diff
@@ -273,3 +305,22 @@ impl FromStr for Type {
...
+        match s {
+            "suburb" => Ok(ZoneTypeWrapped(ZoneType::Suburb)),
+            "city_district" => Ok(ZoneTypeWrapped(ZoneType::CityDistrict)),
+            "city" => Ok(ZoneTypeWrapped(ZoneType::City)),
+            "state_district" => Ok(ZoneTypeWrapped(ZoneType::StateDistrict)),
+            "state" => Ok(ZoneTypeWrapped(ZoneType::State)),
+            "country_region" => Ok(ZoneTypeWrapped(ZoneType::CountryRegion)),
+            "country" => Ok(ZoneTypeWrapped(ZoneType::Country)),
+            "non_admin" => Ok(ZoneTypeWrapped(ZoneType::NonAdministrative)),
+            _ => Err(format!("{} is not a valid zone_type", s)),
```
antoine-de: adding the list of possible value might help the user

libs/bragi/src/params.rs
```diff
@@ -178,6 +198,17 @@ fn check_type(types: &[JsonValue], path: &str) -> Result<(), valico_error::Valic
     Ok(())
 }
 
+fn check_zone_type(zone_type: &str, path: &str) -> Result<(), valico_error::ValicoErrors> {
+    if let Err(e) = ZoneTypeWrapped::from_str(zone_type) {
```
antoine-de: I think it would be more rustic to use pattern matching here

```rust
match ZoneTypeWrapped::from_str(zone_type) {
    Ok(_) => Ok(()),
    Err(e) => Err(vec![Box::new(json_dsl::errors::WrongValue {
            path: path.to_string(),
           detail: Some(e),
        })])
}

```

libs/bragi/src/params.rs
```diff
@@ -273,3 +305,22 @@ impl FromStr for Type {
         }
     }
 }
+
+struct ZoneTypeWrapped(ZoneType);
+
+impl FromStr for ZoneTypeWrapped {
+    type Err = String;
+    fn from_str(s: &str) -> Result<ZoneTypeWrapped, Self::Err> {
```
antoine-de: hum it seems we never use the return value of this, we use only the string, and use this function only to validate the input.

I find it a bit confusing.
If it's really only useful for this I'd prefer a simple list match (with a static array of possible values).
Else the reader can think the type is more useful.

I'm not againt more typing on this too though (str -> ZoneTypeWrapped and only transform ZoneTypeWrapped  at query time). If so we could also implement this trait for ZoneType in cosmogony.

As it stand I'm ok not to over engineer this and only validate the input against a static array.

Do others have opinion on this ? @TeXitoi @kinnou02 ?

libs/bragi/Cargo.toml
```diff
@@ -22,6 +22,7 @@ prometheus = {version= "0.3", features = ["process"]}
 hyper = "0.10"
 heck = "0.3"
 failure = "0.1"
+cosmogony = "0.2"
```
antoine-de: the center tag PR move the requirement on cosmogony to "0.3", we need to take care to have something coherent. (I'd say cosmogony "0.3" everywhere is find (so you need to also update the main Cargo.toml in this PR

tests/cosmogony2mimir_test.rs
```diff
@@ -123,4 +127,30 @@ pub fn cosmogony2mimir_test(es_wrapper: ::ElasticSearchWrapper) {
...
+    let bragi = BragiHandler::new(format!("{}/munin", es_wrapper.host()));
+
+    let response_wrong_zone_type =
+        bragi.raw_get("/autocomplete?q=Saint-Martin-d'Hères&zone_type=invalid_zt");
+    assert!(response_wrong_zone_type.is_err());
+    let iron_error = response_wrong_zone_type.unwrap_err();
+    assert_eq!(iron_error.response.status.unwrap(), BadRequest);
+    let json = to_json(iron_error.response);
+    let error_msg = json.pointer("/long").unwrap().as_str().unwrap();
+    assert!(error_msg.contains("invalid_zt does not belong to the valid zone types list"));
```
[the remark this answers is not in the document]
jbgriesner: Actually this is directly inspired by: 
* https://github.com/CanalTP/mimirsbrunn/blob/4bce5de176578d7064c973610de3d02cd34ff2f9/tests/bragi_filter_types_test.rs#L174
* https://github.com/CanalTP/mimirsbrunn/blob/4bce5de176578d7064c973610de3d02cd34ff2f9/tests/bragi_filter_types_test.rs#L224
* https://github.com/CanalTP/mimirsbrunn/blob/4bce5de176578d7064c973610de3d02cd34ff2f9/tests/bragi_filter_types_test.rs#L236
I guess it's just easier to test since the error message is nested idk.

libs/bragi/src/params.rs
```diff
@@ -320,6 +320,6 @@ fn is_valid(s: &str) -> Result<(), Err> {
     if ZONE_TYPES.contains(&s) {
         Ok(())
     } else {
-        Err(format!("{} does not belong to the valid zone types list: suburb, city, state_district, state, country_region, country, non_admin", s))
+        Err(format!("{} does not belong to the valid zone types list: {:?}", s, ZONE_TYPES))
```
antoine-de: we don't want debug message to user

tests/cosmogony2mimir_test.rs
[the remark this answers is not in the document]
antoine-de: you need to test your own error message

since you output an error with the list of possible values you need to add a test on this

libs/bragi/src/params.rs
jbgriesner: done

tests/cosmogony2mimir_test.rs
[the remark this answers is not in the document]
jbgriesner: done

libs/bragi/src/params.rs
```diff
@@ -306,21 +315,11 @@ impl FromStr for Type {
...
-            "country" => Ok(ZoneTypeWrapped(ZoneType::Country)),
-            "non_admin" => Ok(ZoneTypeWrapped(ZoneType::NonAdministrative)),
-            _ => Err(format!("{} is not a valid zone_type", s)),
-        }
+type Err = String;
+fn is_valid(s: &str) -> Result<(), Err> {
+    if ZONE_TYPES.contains(&s) {
+        Ok(())
+    } else {
+        Err(format!("{} does not belong to the valid zone types list: suburb, city, state_district, state, country_region, country, non_admin", s))
```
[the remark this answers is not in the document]
jbgriesner: done

libs/bragi/src/params.rs
jbgriesner: done

libs/bragi/src/params.rs
jbgriesner: done.

libs/bragi/src/params.rs
```diff
@@ -28,7 +28,9 @@
 // https://groups.google.com/d/forum/navitia
 // www.navitia.io
 
+use cosmogony::ZoneType;
```
antoine-de: why do you need this ?

libs/bragi/src/params.rs
```diff
@@ -28,7 +28,9 @@
...
 use rustless::json::JsonValue;
+use std::fmt;
```
antoine-de: why do you need this ?

libs/bragi/src/params.rs
```diff
@@ -144,6 +163,18 @@ pub fn get_param_array<'a>(params: &'a JsonValue, param_name: &str) -> Vec<&'a s
         .unwrap_or(vec![])
 }
 
+pub fn get_zone_type<'a>(params: &'a JsonValue, param_name: &str) -> Option<&'a str> {
```
antoine-de: this parameter in useless no ?

in general look at the compiler warning, they are really helpful

libs/bragi/src/params.rs
```diff
@@ -273,3 +315,16 @@ impl FromStr for Type {
         }
     }
 }
+
+type Err = String;
+fn is_valid(s: &str) -> Result<(), Err> {
```
antoine-de: the name is too generic.

maybe you can put this directly in `check_zone_type`

libs/bragi/src/params.rs
jbgriesner: just a relict from an obsolet change

libs/bragi/src/params.rs
jbgriesner: also a relict from an obsolet change^^

libs/bragi/src/params.rs
jbgriesner: true!

libs/bragi/src/params.rs
jbgriesner: done

jbgriesner:
I didn't change the cosmogony.json, but my vim config automatically removed the last cr from the file.
It's ok now

antoine-de:
did you had time to review this ?

buildbotctp:

TeXitoi:
that's not very flexible, I can't say "I'm searching for addresses, cities or countries".

antoine-de:
ah I see what you mean, so it would be:
* `/autocomplete?q=bob&type[]=zone` to get all zones with "bob"
* `/autocomplete?q=bob&type[]=zone&type[]=house` to get all addr and zone with "bob" ? 
* `/autocomplete?q=bob&type[]=zone&type[]=house&zone_type=city` to get all addr and city with "bob" ? 
* `/autocomplete?q=bob&type[]=house&zone_type=city` would return an error or won't take `zone_type` into account because there is no `type[]=zone` filter ?

antoine-de:
and what about:
`/autocomplete?q=bob&zone_type=city` so no type and a zone_type. Do we want:

* not to take `&zone_type` into account
* an error
* return all kind of objects, but if it's a zone filter it by zonetype ?

I'd vote for the third option but it might seems a bit weird

tests/cosmogony2mimir_test.rs
```diff
@@ -145,4 +149,44 @@ pub fn cosmogony2mimir_test(es_wrapper: ::ElasticSearchWrapper) {
...
+
+    // (i)
+    let response_wrong_zone_type =
+        bragi.raw_get("/autocomplete?q=Saint-Martin-d'Hères&zone_type=invalid_zt");
+    assert!(response_wrong_zone_type.is_err());
+    let iron_error = response_wrong_zone_type.unwrap_err();
+    assert_eq!(iron_error.response.status.unwrap(), BadRequest);
+    let json = to_json(iron_error.response);
+    let error_msg = json.pointer("/long").unwrap().as_str().unwrap();
+    assert!(error_msg.contains("invalid_zt does not belong to the valid zone types list: suburb, city, state_district, state, country_region, country, non_admin"));
```
pbougue: You can probably shorten the string contained and remove the list, as it may evolve or the sorting could change.
We'd win readability and maintenance.

antoine-de: I don't really agree there, at first there way a bug in the eror message creation and the list of values was not there. As it's very useful for the user I think it should be tested :stuck_out_tongue_winking_eye:

PR #237, opened 2018-07-02
Read bbox from cosmogony and osm admins, add bbox in bragi response

amatissart:
This PR implements a bbox field in the geocoder response: very useful to move and zoom in/out on a map when a user selects a result. 

* Add a new field `bbox: Option<geojson::Bbox>` in mimir Admin 
* Fill in this new field
  * read  `bbox` from the cosmogony admin in cosmogony2mimir, 
  * when importing admins using osm_reader (in osm2mimir), calculate `bbox` from the boundary
* Add a new field `bbox` in bragi GeocodingResponse, that will be output for admins with bbox

All `bbox` are instances of `geojson::Bbox` which is an alias for `Vec<f64>`.   
Another solution would be to use `geo::Bbox` (already used by Cosmogony), but that would require to implement specific serialization/deserialization. There is space for discussion here.

---

Note that this PR replaces `cosmogony.json` fixture with an updated pretty-printed version (easier to maintain and update on conflicts).


TeXitoi:
I prefer a rust::geo in the model. Updating the geo crate is great and should not be so difficult.

amatissart:
I pushed corrections to use `geo::Bbox` in the models with custom serialization/deserialization methods. Ideally we could expect that such serialization would be implemented in rust-geojson, but it is not feasible now as `Bbox` is not even published in the new "geo-types" crate (and it would require to upgrade "geo" in multiple dependencies).  

Also, I removed the `Deserialize` trait from structs in bragi, as I couldn't find any good reason to make them depend on deserialization details.

PR #239, opened 2018-07-05
Do not index useless fields

antoine-de:
by default elasticsearch index all unknown field (not in the mapping).

this means that it will take some time and some space to index some useless field. Like for example the big admins name (thrice in the poi index :wink: )

https://www.elastic.co/guide/en/elasticsearch/guide/current/dynamic-mapping.html

adding `"dynamic": "false"` to the mapping tells ES not to index the unknown field (but they are still stored).

Note: do not merge this PR yet, I opened it for review, and I'm waiting for the data import process to give figures about the impact on the index time/space

### DIFF
--- a/json/addr_settings.json
+++ b/json/addr_settings.json
@@ -49,6 +49,7 @@
     },
     "mappings": {
         "addr": {
+            "dynamic": "false",
             "properties": {
                 "id": { "type": "string", "index": "no" },
                 "house_number": { "type": "string", "analyzer": "word"},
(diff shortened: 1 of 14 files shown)


nonifier:
Is there any impact performance wise ? is it any faster ?

antoine-de:
I'm currently running some benches to give more figures but I hope it will be faster since ES will have way less work during bulk index :stuck_out_tongue_winking_eye:

buildbotctp:

TeXitoi:
@antoine-de can you give a performance value for this PR?

TeXitoi:
We may, in the future, need indexing the shape of the admin for a cities-like service, or a "give me the country of this point" service.

antoine-de:
yes it might be nice, but I think we'll add the index when needed (and anyway the current indexing is useless as we index it as a list a floats, not as a shape)

For the performance, I'm not very comfortable with my own benchmarks as it has be quite a tedious process to bench all 3 branches separately so I would really appreciate an external bench (but the gitlab runner stuff does not seems to move on our side...).

what I got for france is (but again it's a my laptop while doing other stuff):

branch | cosmo2mimir | osm2mimir
----| ---|----
master|00:20|1:37
useless_indexes|00:16|1:17

and I have not kept the addresses part which is the most important :confused: 

Note: from my point of view I think you can remove the "do_not_merge" tag

PR #240, opened 2018-07-05
Admin lookup speedup

antoine-de:
# Speed up the association to an admin

Our data import process to import the addresses is very long, even on a quite small dataset like France.

The major time (80% of the run time in our case) is taken to associate an admin to an address.
To do this, mimir load all the admin in a rtree and check the admin boundaries.

Since we use [cosmogony](https://github.com/osm-without-borders/cosmogony) to import our admins, we got lots of them (compared to the basic CanalTP use case that read only the admin 8 from osm).

This PR speed up the admin association by using the cosmogony hierarchy.
Cosmogony already do lots of geographic inclusion to produce a hierachy. The idea is to store this hierarchy in ES (only storing the parent's ID will not overload the db) and skip some `contains` based on this hierarchy.

We also consider that we associate a point to only one zone of a kind (at most 1 city, 1 state, 1 country, ...). This is clearly a bias since there is nothing stopping 2 cities to overlap, but we think we don't want to associate a point to 2 city. 
What do you think of this ?

This speeds up our import process by more than 2 (17h :scream: hours for the 28M addresses of OA drops to 7h) since we now only check 1 or 2 zones (and we especially tests very few big zones that have lots a points and are long to check)

I still need to run some bench to check that your import process is not impacted by this change (I think the only overhead is that we now sort the rtree results + the creation of the admin map by id, I need to check the impact)

### DIFF
--- a/src/admin_geofinder.rs
+++ b/src/admin_geofinder.rs
@@ -66,7 +66,7 @@ impl AdminGeoFinder {
 
         let rect = {
             let mut coords = match admin.boundary {
-                Some(ref b) => (*b).0.iter().flat_map(|poly| (poly.exterior).0.iter()),
+                Some(ref b) => b.0.iter().flat_map(|poly| (poly.exterior).0.iter()),
                 None => return,
             };
             let first_coord = match coords.next() {
@@ -88,7 +88,7 @@ impl AdminGeoFinder {
         };
         let bound_admin = BoundaryAndAdmin::new(admin);
         self.admin_by_id
-            .insert((*bound_admin.1).id.clone(), bound_admin.1.clone());
+            .insert(bound_admin.1.id.clone(), bound_admin.1.clone());
         self.admins.insert(rect, bound_admin);
     }
 
@@ -98,7 +98,7 @@ impl AdminGeoFinder {
         let search = Rect::from_float(down(x), up(x), down(y), up(y));
         let mut rtree_results = self.admins.get(&search);
 
-        rtree_results.sort_by_key(|(_, a)| (*a.1).zone_type);
+        rtree_results.sort_by_key(|(_, a)| a.1.zone_type);
 
         let mut tested_hierarchy = BTreeSet::<String>::new();
         let mut added_zone_types = BTreeSet::new();
@@ -107,9 +107,9 @@ impl AdminGeoFinder {
         for boundary_and_admin in rtree_results.into_iter().map(|(_, a)| a) {
             let boundary = &boundary_and_admin.0;
             let admin = &boundary_and_admin.1;
-            if tested_hierarchy.contains(&(*admin).id) {
+            if tested_hierarchy.contains(&admin.id) {
                 res.push(admin.clone());
-            } else if (*admin)
+            } else if admin
                 .zone_type
                 .as_ref()
                 .map_or(false, |zt| added_zone_types.contains(zt))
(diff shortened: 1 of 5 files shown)


kinnou02:
Having only one object type associated to each address doesn't cause any problems in France, but not all countries have a nice administrative "mille-feuille" like us. I will be surprised if there isn't any change in the results for European countries like UK or Germany.

antoine-de:
I haven't run geocoder on other countries than france :confounded:  (it is already quite a tedious process :sob: )

I don't really see why this would cause problem in germany or UK since it's something that should quite be taken care of by cosmogony no ?

src/admin_geofinder.rs
```diff
@@ -84,24 +86,57 @@ impl AdminGeoFinder {
...
+            let boundary = &boundary_and_admin.0;
+            let admin = &boundary_and_admin.1;
+            if tested_hierarchy.contains(&(*admin).id) {
+                res.push(admin.clone());
+            } else if (*admin)
+                .zone_type
+                .as_ref()
+                .map_or(false, |zt| added_zone_types.contains(zt))
+            {
+                // we don't want it, we already have this kind of ZoneType
```
TeXitoi: What if 2 zone type overlap?

src/admin_geofinder.rs
```diff
@@ -84,24 +86,57 @@ impl AdminGeoFinder {
...
+        rtree_results.sort_by_key(|(_, a)| (*a.1).zone_type);
+
+        let mut tested_hierarchy = BTreeSet::<String>::new();
+        let mut added_zone_types = BTreeSet::new();
+        let mut res = vec![];
+
+        for boundary_and_admin in rtree_results.into_iter().map(|(_, a)| a) {
+            let boundary = &boundary_and_admin.0;
+            let admin = &boundary_and_admin.1;
+            if tested_hierarchy.contains(&(*admin).id) {
```
TeXitoi: auto deref doesn't work?

src/admin_geofinder.rs
```diff
@@ -84,24 +86,57 @@ impl AdminGeoFinder {
...
+                res.push(admin.clone());
+            } else if (*admin)
```
TeXitoi: auto deref doesn't work?

src/admin_geofinder.rs
```diff
@@ -84,24 +86,57 @@ impl AdminGeoFinder {
...
-            .filter(|a| {
-                a.0
-                    .as_ref()
-                    .map_or(false, |b| (*b).contains(&geo::Point(*coord)))
-            })
-            .map(|admin_and_boundary| admin_and_boundary.1.clone())
-            .collect()
+        let mut rtree_results = self.admins.get(&search);
+
+        rtree_results.sort_by_key(|(_, a)| (*a.1).zone_type);
```
TeXitoi: auto deref doesn't work?

src/admin_geofinder.rs
```diff
@@ -84,24 +86,57 @@ impl AdminGeoFinder {
...
+                res.push(admin.clone());
+            } else if (*admin)
+                .zone_type
+                .as_ref()
+                .map_or(false, |zt| added_zone_types.contains(zt))
+            {
+                // we don't want it, we already have this kind of ZoneType
+            } else if boundary
+                .as_ref()
+                .map_or(false, |b| (*b).contains(&geo::Point(*coord)))
```
TeXitoi: auto deref doesn't work? And for the others `(*something)`?

src/admin_geofinder.rs
antoine-de: it's clearly a bias there, we will associate the object only to the first found zone.

We should clearly discuss it, because it can clearly happen, but the speedup is quite significant since it makes it possible to almost never check the inclusion for big zones likes countries.

what do you think about the tradeoff @kinnou02 @amatissart @TeXitoi ?

buildbotctp:

src/admin_geofinder.rs
antoine-de: oups don't know how I missed this, it's corrected

antoine-de:
what's the status on this PR? (apart with the rebase needed :wink: )

It would really helps a lot with our import process.

TeXitoi:
For our usecase, this PR doesn't change anything. That's a first step, but must be changed in the future because it will not work in all case.

I'm in favor of merging. @pbougue @kinnou02 what do you think?

kinnou02:
I'am also for the merge

PR #241, opened 2018-07-05
do the rubber stuff in parallele

antoine-de:
consume the index given to rubber::index in parallel

this will makes some work in //, but will keep the bulk synchronous.
Note: for the moment the number of threads in configured by an env var to ease the POC, if you prefer (and if the threading is ok for you), I can put this as a cli var of all the binaries

this will mean different things for each importer (but basically it is only useful for addresses for now):

## bano2mimir / oa2mimir
the addresses will be:
* deserialized with serde
* associated to the admins

in // by bulk of 1000 and the bulk will be send before
deserializing/associating to admins in // again

this greatly reduce the import time while keeping the same load on ES.

## osm2mimir
not much here, we'll need more refactoring to do some things in //

## stops2mimir/ntfs2mimir
not much either (in part because the weight system that required all the
stops to be read before sending them to the bulk)

Anyway I don't think this is important since those importer should be
quite quick

## cosmogony2mimir
same, not much done in // with this (same requirement as stops2mirmir)
but the limiting factor here will likely only be the bulk insert in ES
too

Note: Do not merge this PR yet,, it's only for early review, I need to run more bench to give for figures


libs/mimir/src/rubber.rs
```diff
@@ -545,16 +545,20 @@ impl Rubber {
...
+        T: MimirObject + std::marker::Send + 'static,
         I: Iterator<Item = T>,
     {
+        use par_map::ParMap;
         use rs_es::operations::bulk::Action;
         let mut nb = 0;
         let chunk_size = 1000;
-        // fold is used for creating the action and optionally set the id of the object
-        let mut actions = iter.map(|v| {
+        let nb_threads = std::env::var("MIMIR_NB_THREADS")
```
TeXitoi: No structopt option?

libs/mimir/src/rubber.rs
```diff
@@ -545,16 +545,20 @@ impl Rubber {
...
+            .map(|v| v.parse().expect("unable to read the number of threads"))
+            .unwrap_or(1);
+        let mut actions = iter.with_nb_threads(nb_threads).par_map(|v| {
```
TeXitoi: Won't it be better to par_map on chunks?

libs/mimir/src/rubber.rs
```diff
@@ -545,16 +545,20 @@ impl Rubber {
...
             v.es_id()
                 .into_iter()
+                // fold is used for creating the action and optionally set the id of the object
                 .fold(Action::index(v), |action, id| action.with_id(id))
```
TeXitoi: what it so costly here that it's interesting to parallelize?

antoine-de: yes I tried detailing this in the PR description because this changes with the import binary used.

for addresses it's the admin association indeed and it's quite interesting to `//`.
what kind of filtering are you thinking about ?

libs/mimir/src/rubber.rs
antoine-de: by laziness no, since I was not sure about this change I made it as small as possible, but if you insist I can do it.

libs/mimir/src/rubber.rs
antoine-de: just for the record, with #239 #240 and #241 the OA import drops from 17h to 1h20 on my laptop.

I wanted to run all the benchmark with the new gitlab runner because it seems it was about to be unblocked and I preferred a cleaner bench, but it seems to be more complicated that planned :confused:

TeXitoi: I was thinking about #240

buildbotctp:

libs/mimir/src/rubber.rs
antoine-de: on my not to precise benchmarks, with this the import time (cosmogony+admin2mimir+oa+osm[street]) drops from around 17h to 7h.

and the whole process drops to 2h20 with this (+ #239)

libs/mimir/src/rubber.rs
antoine-de: we could but I'm afraid to flood the Elasticsearch And though it better to only parallelise internal work, not the bulk insert.

what do you think about this ?

TeXitoi:
```rust
let mut chunks = iter.with_nb_threads(nb_threads).packed_par_map(chunk_size, |v| {
    v.es_id()
        .into_iter()
        .fold(Action::index(v), |action, id| action.with_id(id))
});
for chunk in chunks.filter(|c| !c.is_empty()) {
    self.es_client
        .bulk(&chunk)
        .with_index(&index.name)
        .with_doc_type(T::doc_type())
        .send()?;
}
```

No parallel bulk, but generating the chunk in parallel.

antoine-de: ah this way ok.

I've done the changes and retested geocoder tester, it's all good.

PR #245, opened 2018-08-10
Update doc

mbounabi:


antoine-de:
I think travis is not happy because rust-fmt has been updated with the new rust release and `cargo fmt --all -- --write-mode diff` has been changed to `cargo fmt --all -- --chek`
.
NOTE: it's really minor, but I think it's better to fork the project on you own user space, it can help keeping the branches of the main repository clean and it can avoid mistakes

antoine-de:
can you rebase on master for travis to be happy ?

or maybe we can merge as is, as the problem is not related to this PR

PR #247, opened 2018-08-10
remove useless log

antoine-de:
in the bano dataset there are 284917 addresses without a street name so
it flood the logs, I propose to move the log as debug (are remove the
debug dump of the address


buildbotctp:

PR #252, opened 2018-08-23
Add an option to choose the number of threads in bragi

kinnou02:
By default 8 threads are created by cpu core, like before.

### DIFF
--- a/libs/bragi/src/lib.rs
+++ b/libs/bragi/src/lib.rs
@@ -54,9 +54,10 @@ extern crate slog_scope;
 
 #[macro_use]
 extern crate failure;
+extern crate num_cpus;
 
 use iron::prelude::Chain;
-use iron::Iron;
+use iron::{Iron, Protocol};
 use rustless::Application;
 use structopt::StructOpt;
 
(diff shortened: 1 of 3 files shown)


libs/bragi/src/lib.rs
```diff
@@ -86,6 +89,9 @@ pub struct Args {
     /// Elasticsearch parameters, override BRAGI_ES environment variable.
     #[structopt(short = "c", long = "connection-string", raw(default_value = "&BRAGI_ES"))]
     connection_string: String,
+    /// Number of threads used to serve http requests, override BRAGI_NB_THREADS environment variable.
+    #[structopt(short = "t", long = "nb-threads", raw(default_value = "&BRAGI_NB_THREADS"))]
+    nb_threads: usize,
```
TeXitoi: you can simplify with something like
```
#[structopt(short = "t", long = "nb-threads", env = "BRAGI_NB_THREADS", raw(default = "&DEFAULT_NB_THREADS"))]
```

same for `--connection-string`

antoine-de:
by curiosity what values do you use ?

kinnou02:
I didn't change it yet, but I am thinking at something like 40 thread per hosts. It's a palliative to the fact that bragi isn't async...

PR #253, opened 2018-08-27
Add a timeout parameters in the API and use it when calling ES

kinnou02:
The `timeout` is in milliseconds.

I'm considering if we need to have a default timeout or not, probably...

Only the first commit (2f7461d) is worth reading, others are fmt and cargo stuff.

### DIFF
--- a/libs/bragi/src/api.rs
+++ b/libs/bragi/src/api.rs
@@ -35,7 +35,8 @@ use mimir::rubber::Rubber;
 use model;
 use model::v1::*;
 use params::{
-    coord_param, dataset_param, get_param_array, paginate_param, shape_param, types_param, timeout_param,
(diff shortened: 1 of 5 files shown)


libs/bragi/src/api.rs
```diff
@@ -257,7 +265,10 @@ impl ApiEndPoint {
                     let all_data = params
                         .find("_all_data")
                         .map_or(false, |val| val.as_bool().unwrap());
-                    let features = query::features(&pt_datasets, all_data, &cnx, &id);
+                    let timeout = params
+                        .find("timeout")
+                        .and_then(|v| Some(time::Duration::from_millis(v.as_u64().unwrap())));
```
TeXitoi: A function with that would be good.

Also:
```rust
let timeout = params
    .find("timeout")
    .and_then(|v| v.as_u64())
    .map(time::Duration::from_millis);
```

PR #255, opened 2018-09-12
Fix addresses multi threading

antoine-de:
In #241 I  tried to run some import stuff in parallel by making rubber do its stuff in parallel.

It turns out I completly messed thing up since the only thing that was done in parallel was creating the [ES Action](https://github.com/CanalTP/mimirsbrunn/blob/3fa41e672bd81c16ce7d6af5f3cfe9f95a6ef202/libs/mimir/src/rubber.rs#L573-L579), so something very trivial. So the import was really multi thread, but all the work was done by he main thread and the additional threads had very few things to do :flushed: 

At that time I tried several improvement at the same time so I think it's how I completely messed up the benchmarks :sob: 

This PR corrects this by making only the addresses importer (OpenAddress and Bano) multi threads.
This removes the env var `MIMIR_NB_THREADS` and add cil args `--nb-threads` to control the number of threads.

The default value is the number of cpus.

Thanks @TeXitoi for the huge help with the fight against the borrow checker!


src/bin/openaddresses2mimir.rs
```diff
@@ -116,32 +129,40 @@ where
...
+                })
+        })
+        .with_nb_threads(nb_threads)
+        .par_map(move |v: OpenAddresse| v.into_addr(&admins_geofinder))
+        .filter(|a| {
+            !a.street.name.is_empty() || {
+                debug!("Address {} has no street name and has been ignored.", a.id);
+                false
+            }
+        });
```
TeXitoi: That's a big copy/paste, can you mutualise it?
```rust
fn deserialize_files<T: Deserialize>(files: impl IntoIter<PathBuf>) -> impl Iterator<T>
```

TeXitoi: Or something like
```rust
impl Rubber {
    fn import_addresses<T: Deserialize<Addr>>(
        &mut self,
        nb_threads: u32,
        files: impl IntoIter<PathBuf>,
        into_addr: impl Fn(T) -> Addr,
    )
```

antoine-de: I don't think this should be the responsibility of rubber (and rubber is already quite huge), but you're absolutely right it needs factorization.

I added a function in a `addr_reader.rs` file for it

antoine-de:
no, what do you mean? :roll_eyes: :man_facepalming:

src/addr_reader.rs
```diff
@@ -0,0 +1,62 @@
...
+use std::marker::{Send, Sync};
+use std::path::PathBuf;
+use Error;
+
+pub fn import_addresses<T, F>(
+    rubber: &mut Rubber,
+    has_headers: bool,
+    nb_threads: usize,
+    dataset: &str,
+    files: impl Iterator<Item = PathBuf>,
```
TeXitoi: Using IntoIter is best practice, then you don't need an `iter()` or `into_iter` in calling code.

kinnou02:
@TeXitoi If it's good for you too I will merge it

antoine-de:
I just changed to `IntoIterator`

PR #259, opened 2018-09-18
fix useless method

antoine-de:
A test method was never called (my bad :confused: ).

The thing is that the test was not passing either, because if we look for a non-existing index, we won't have a specific error as the test state, but a generic one (no object found). (I think this was implemented like this because an index might be missing if no data of that type has been imported yet).

This is done this way because the rubber `get_indexes` method always returns a non-empty list of elements (there is always at least "munin_geo_data" returned)

I think the get_indexes must also be refactored, but it will be for another PR.
I'd vote for:
* https://github.com/CanalTP/mimirsbrunn/blob/master/libs/mimir/src/rubber.rs#L233 => I don't see why it should be the case, we should add all pt_dataset
* if a pt_dataset does not exist, return an `IndexNotFound` error.
* I don't know when you want to search on `munin_global_stops`

As this refactor concerns more Kisio's use case, I'll let you do it.

### DIFF
--- a/libs/bragi/src/query.rs
+++ b/libs/bragi/src/query.rs
@@ -308,7 +308,7 @@ pub fn features(
     if indexes.is_empty() {
         // if there is no indexes, rs_es search with index "_all"
         // but we want to return an error in this case.
-        return Err(BragiError::IndexNotFound);
+        return Err(BragiError::ObjectNotFound);
     }
 
     let result: SearchResult<serde_json::Value> = try!(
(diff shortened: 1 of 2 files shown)


TeXitoi:
> https://github.com/CanalTP/mimirsbrunn/blob/master/libs/mimir/src/rubber.rs#L233 => I don't see why it should be the case, we should add all pt_dataset

That's a feature: when we have more than one dataset, we use the merged dataset, and filter in the query by coverage.

antoine-de:
ah ok.
I think it should be documented :stuck_out_tongue_winking_eye:

pbougue:
OK, so nothing to be done for 1st and 3rd point, behavior is the one expected.
We could do something on the second point when getting an unknown index, but it's not critical and we'll wait to see if the need arises.

PR #260, opened 2018-09-24
Don't check if index exist before doing a query

kinnou02:
We use `ignore_unavailable` to ignore index that don't exist to not have
to check if the index exist before.

https://www.elastic.co/guide/en/elasticsearch/reference/2.4/multi-index.html

waiting for https://github.com/benashford/rs-es/pull/121 fork in the second commit of this PR, don't read that.

### DIFF
--- a/libs/mimir/src/rubber.rs
+++ b/libs/mimir/src/rubber.rs
@@ -178,12 +178,7 @@ pub fn build_proximity_with_boost(coord: &Coord, boost: f64) -> Query {
         ).build()
 }
 
-
-pub fn get_indexes(
-    all_data: bool,
-    pt_datasets: &[&str],
-    types: &[&str],
-) -> Vec<String> {
+pub fn get_indexes(all_data: bool, pt_datasets: &[&str], types: &[&str]) -> Vec<String> {
     if all_data {
         return vec!["munin".to_string()];
     }
@@ -195,7 +190,7 @@ pub fn get_indexes(
         0 => (),
         1 => {
             for pt_dataset in pt_datasets.iter() {
-                    pt_dataset_indexes.push(format!("munin_stop_{}", pt_dataset));
+                pt_dataset_indexes.push(format!("munin_stop_{}", pt_dataset));
             }
         }
         _ => pt_dataset_indexes.push("munin_global_stops".to_string()),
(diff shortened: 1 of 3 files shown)


kinnou02:
It look like we never want to check that an index exist, if that the case `get_indexes` can be simplified,

libs/mimir/src/rubber.rs
```diff
@@ -197,7 +197,7 @@ pub fn get_indexes(
     client: &mut rs_es::Client,
 ) -> Result<Vec<String>, EsError> {
     make_indexes_impl(all_data, pt_datasets, types, |index| {
-        is_existing_index(client, index)
+        Ok(true)
     })
 }
```
TeXitoi: You can cleanup all that: `get_indexes` doesn't need to return a result now, as `make_indexes_impl`.

`client` argument is no longer used (as said in the warning).

kinnou02:
@antoine-de can you check that it doesn't impact others components that your build with mimir?

kinnou02:
btw, I saw that `/features` look for every types, so we search in every index, we might want to infer the type of the object

libs/mimir/src/rubber.rs
```diff
@@ -178,77 +178,40 @@ pub fn build_proximity_with_boost(coord: &Coord, boost: f64) -> Query {
...
-                try!(push(
-                    &mut pt_dataset_indexes,
-                    format!("munin_stop_{}", pt_dataset).as_str(),
-                ))
+                pt_dataset_indexes.push(format!("munin_stop_{}", pt_dataset));
             }
         }
-        _ => try!(push(&mut pt_dataset_indexes, "munin_global_stops")),
+        _ => pt_dataset_indexes.push("munin_global_stops".to_string()),
     };
```
TeXitoi: You can use slice pattern: https://play.rust-lang.org/?gist=4ef77ed0b8f69bf652642a482c165f1b&version=stable&mode=debug&edition=2015

libs/mimir/src/rubber.rs
```diff
@@ -178,77 +178,40 @@ pub fn build_proximity_with_boost(coord: &Coord, boost: f64) -> Query {
...
 
     match types.len() {
```
TeXitoi:
```rust
if type.is_empty() {
```

Issue #248, opened 2018-08-10
Better duplicates handling 👯


antoine-de:
# :dancing_women: Duplicates :dancing_men: 
we added [a way](https://github.com/QwantResearch/geocoder-tester/pull/26) in geocoder tester to check the duplicates in the bragi's response.

## problem
Duplicates in the geocoder response are entries that the user can't differentiate.
This is thus dependent on what you display to your users, so we used:
* for a poi, the object's label + it's address
* for the other objects only the label

The results for the French dataset (bano/osm/poi from our own fafnir importer) are not that bad:

|test | number of duplicates |
|--- | ---|
|poi |0 |
|admins |4 |
|fautes | 22|
|autre | 106|

(for the bano dataset, the results would likely be quite different for the OA dataset).

This duplicates are:
* real data duplicates (twice the same POI in osm for example, or duplicates in bano like [8 Allée du Rhône (Grigny)](https://gist.github.com/antoine-de/2e34d154dbb1edf6c6532d50ba3e1f59) ,and lots of duplicates in OA :stuck_out_tongue_winking_eye: )
* different objects in osm to represent the same thing for us (like this [hotel](https://www.openstreetmap.org/way/286852141) and it's [several](https://www.openstreetmap.org/way/66276085) [buildings](https://www.openstreetmap.org/way/66284816)
* mistakes in OSM, especially on the mapping of associated street (as we handle them anyway). @nlehuby created an [issue on osmose](https://github.com/osm-fr/osmose-backend/issues/318) to explain the problem and track them
* tricky corner case as the street "rue carnot" that is in the city [Montreuil](https://fr.wikipedia.org/wiki/Montreuil_(Pas-de-Calais)) and the city [Montreuil](https://fr.wikipedia.org/wiki/Montreuil_(Seine-Saint-Denis)) :stuck_out_tongue_winking_eye: 
* tricky corner case with different streets with the same name in the same city (as the famous "rue Jean Jaurès" in Lille (And I think we can have the same problem all around the world, especially in huge city like berlin). For this I think we should have a better algorithm to group streets (not just 1 street by city / or associated streets) 

### post-process
I think it's important to limit the number of duplicates in the ES database (so for example that if we want to integrate more OA datasets + some addresses from OSM, we'll need to handle the duplicates in the import pipeline).

However since the number of duplicates is quite limited for the moment I think we could do some post-processing to reduce their number even more.

[...]

PR #265, opened 2018-10-05
Remove Bragi duplicates

jbgriesner:
This PR is a first proposition to handle the duplicate results in bragi response.
This issue was discussed here : https://github.com/CanalTP/mimirsbrunn/issues/248

Basically the idea is to remove results with the same (name, label, postcode, smallest_admin).
Then we enrich remaining results with identical name and label to add either their postcode or the smallest_admin zone_type as a label to distinguish it.

Simple functional test and unit test come along the changes.


libs/bragi/src/api.rs
```diff
@@ -96,6 +101,86 @@ fn add_distance(autocomp_resp: &mut model::Autocomplete, origin_coord: &Coord) {
...
+    }
+
+    fn smallest_admin(admins: &Vec<Arc<mimir::Admin>>) -> Option<String> {
+        let mut zones: Vec<Option<ZoneType>> = Vec::new();
+
+        zones.extend(admins.iter().map(|a| a.clone().zone_type));
+        zones.sort();
+
+        if !zones.is_empty() {
+            match zones[0] {
```
TeXitoi:
```rust
if let Some(first_zone) = admins.iter().map(..).min() {
```
If you have 1 None, it will be in first position.

libs/bragi/src/api.rs
```diff
@@ -414,9 +502,138 @@ impl ApiEndPoint {
...
+
                     render(client, response)
                 })
             });
         })
     }
 }
+
+#[allow(dead_code)]
+fn get_feature(
```
TeXitoi: https://doc.rust-lang.org/book/second-edition/ch11-03-test-organization.html

libs/bragi/src/api.rs
```diff
@@ -414,9 +502,138 @@ impl ApiEndPoint {
...
+    nme: Option<String>,
+    lbl: Option<String>,
+    cde: Option<String>,
+    adm_name: String,
+    zt_adm: Option<ZoneType>,
+) -> model::Feature {
+    use geojson::{Geometry, Value};
+    model::Feature {
```
TeXitoi: You can use something like
```rust
model::Feature {
    feature_type: "Feature_type".to_string(),
    // ...
    .. Default::default()
}

File: libs/bragi/src/api.rs [hunk 4]
```diff
@@ -414,9 +502,138 @@ impl ApiEndPoint {
...
+#[allow(dead_code)]
+fn get_features() -> Vec<model::Feature> {
+    let mut features: Vec<model::Feature> = Vec::new();
+    features.push(get_feature(
+        Some("2 Rue du Doublon".to_string()),
+        Some("2 Rue du Doublon (Paris)".to_string()),
+        Some("75012".to_string()),
+        "".to_string(),
+        Some(ZoneType::City),
+    ));
```
TeXitoi: it will be better if
```rust
fn get_feature(&str, &str, &str, &str, ZoneType) -> Feature
```

libs/bragi/src/model.rs
```diff
@@ -57,7 +57,17 @@ pub struct Geocoding {
     query: Option<String>,
 }
 
-#[derive(Serialize, Debug)]
+impl Geocoding {
+    #[allow(dead_code)]
```
TeXitoi: No

libs/bragi/src/model.rs
```diff
@@ -67,12 +77,19 @@ pub struct Feature {
     pub distance: Option<u32>,
 }
 
-#[derive(Serialize, Debug)]
+impl PartialEq for Feature {
+    fn eq(&self, other: &Feature) -> bool {
+        self.properties.geocoding.label == other.properties.geocoding.label
+            && self.properties.geocoding.name == other.properties.geocoding.name
```
TeXitoi: No

jbgriesner:
Thanks for your review !

antoine-de:
we close this after the discussion on the duplicates, we don't think this as a big problem for the moment, see #248

PR #266, opened 2018-10-12
add a few metrics to prometheus

kinnou02:
Call to ES for /features and /reverse are now measured.
Add a gauge to keep track of the number of request currently handled


libs/mimir/src/rubber.rs
```diff
@@ -69,6 +70,14 @@ const SYNONYMS: [&'static str; 17] = [
     "anpe,pole emploi",
 ];
 
+lazy_static! {
+    static ref ES_REQ_HISTOGRAM: prometheus::Histogram = register_histogram!(
+        "bragi_elasticsearch_reverse_duration_seconds",
+        "The elasticsearch request latencies in seconds.",
```
TeXitoi: it lacks reverse in the description

PR #269, opened 2018-10-16
Do not expose internal serveur message to the client

antoine-de:
return in the json response a generic "internal server error" on
error 500 (or "service unavailable" when ES is not reachable, status code 503) not to have
error messages like:

```json
{
   "short":"query error",
   "long":"invalid query 503 Service Unavailable -
{\"error\":{\"root_cause\":[],\"type\":\"search_phase_execution_exception\",\"reason\":\"all
shards
failed\",\"phase\":\"query\",\"grouped\":true,\"failed_shards\":[]},\"status\":503}"
}
```

the error is logged to trace it.

### DIFF
--- a/libs/bragi/src/model.rs
+++ b/libs/bragi/src/model.rs
@@ -380,7 +380,7 @@ pub mod v1 {
     use super::BragiError;
     use iron;
     use mimir;
-    use rs_es;
+    use rs_es::error::EsError;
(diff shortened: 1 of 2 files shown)


antoine-de:
:ping_pong:, don't you like the nicely formated response :roll_eyes:  ?

PR #270, opened 2018-10-18
Timeout for the ES query

jbgriesner:
A timeout parameter has been recently added (https://github.com/CanalTP/mimirsbrunn/commit/4b2d149bcc2958736791ab1073fed7f0e25b5f7b) but it handles only the network and not the time limit allowed to the ES query.

This PR proposes to use the same timeout both for the network and for the query.

This parameter can be defined at two places:
- as a command line argument launching bragi: `default_es_max_timeout=...`
- or directly into the user query: `timeout=...`

In this PR we changed the previous behavior: instead to overwrite the default timeout with the query one, we keep the minimum. In other words the default timeout becomes the max.

### DIFF
--- a/libs/bragi/src/api.rs
+++ b/libs/bragi/src/api.rs
@@ -86,10 +86,11 @@ fn parse_timeout(
         .find("timeout")
         .and_then(|v| v.as_u64())
         .map(time::Duration::from_millis)
-        .and_then(|t| match default_timeout {
-            Some(dt) => Some(t.min(dt)),
-            None => Some(t),
-        }).or(default_timeout)
+        .map(|t| match default_timeout {
+            Some(dt) => t.min(dt),
+            None => t,
+        })
+        .or(default_timeout)
 }
 
 fn add_distance(autocomp_resp: &mut model::Autocomplete, origin_coord: &Coord) {
--- a/libs/bragi/src/api.rs
+++ b/libs/bragi/src/api.rs
@@ -89,8 +89,7 @@ fn parse_timeout(
         .map(|t| match default_timeout {
             Some(dt) => t.min(dt),
             None => t,
-        })
-        .or(default_timeout)
+        }).or(default_timeout)
 }
 
 fn add_distance(autocomp_resp: &mut model::Autocomplete, origin_coord: &Coord) {
--- a/libs/bragi/src/api.rs
+++ b/libs/bragi/src/api.rs
@@ -86,7 +86,10 @@ fn parse_timeout(
         .find("timeout")
         .and_then(|v| v.as_u64())
         .map(time::Duration::from_millis)
-        .or(default_timeout)
+        .and_then(|t| match default_timeout {
+            Some(dt) => Some(t.min(dt)),
+            None => Some(t),
+        }).or(default_timeout)
 }
 
 fn add_distance(autocomp_resp: &mut model::Autocomplete, origin_coord: &Coord) {
@@ -115,7 +118,7 @@ where
 
 pub struct ApiEndPoint {
     pub es_cnx_string: String,
-    pub default_es_timeout: Option<time::Duration>,
+    pub default_es_max_timeout: Option<time::Duration>,
 }
 
 impl ApiEndPoint {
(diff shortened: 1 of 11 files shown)


libs/bragi/src/api.rs
```diff
@@ -115,7 +118,7 @@ where
 
 pub struct ApiEndPoint {
     pub es_cnx_string: String,
-    pub default_es_timeout: Option<time::Duration>,
+    pub default_es_max_timeout: Option<time::Duration>,
```
antoine-de: `max_es_timeout` no ?

libs/bragi/src/query.rs
```diff
@@ -264,18 +265,37 @@ fn query(
...
-        .with_from(offset)
-        .with_size(limit)
-        .send()?;
+    let result: SearchResult<serde_json::Value> = match timeout {
+        Some(t) => {
+            let duration = t.as_secs() as f64 + t.subsec_nanos() as f64 * 1e-9;
+
+            client
+                .search_query()
+                .with_timeout(&format!("{}s", duration.to_string()))
```
antoine-de: you don't have to copy paste all the search query, you can just  do something like (not checked):

```rust
let query = 
            client
                .search_query()
                .with_timeout(&format!("{}s", duration.to_string()))
                .with_ignore_unavailable(true)
                .with_indexes(
                    &indexes
                        .iter()
                        .map(|index| index.as_str())
                        .collect::<Vec<&str>>(),
                ).with_query(&query)
                .with_from(offset)
                .with_size(limit);

if let Some(timeout) = timeout {
query.with_timeout(...);
}
let result = query.send()?;

```

libs/bragi/src/query.rs
```diff
@@ -264,18 +265,37 @@ fn query(
...
-                .iter()
-                .map(|index| index.as_str())
-                .collect::<Vec<&str>>(),
-        ).with_query(&query)
-        .with_from(offset)
-        .with_size(limit)
-        .send()?;
+    let result: SearchResult<serde_json::Value> = match timeout {
+        Some(t) => {
+            let duration = t.as_secs() as f64 + t.subsec_nanos() as f64 * 1e-9;
```
antoine-de: I'm not a fan of this implementation
It's not great to debug on debug format(and it would require a test to have no regressions), but it [*seems to respect ES duration format](https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=a89628cc094ced03086819fb18218235).

amatissart:
To illustrate how setting ES timeout helps, here are response times during a test with 20 req/s (long fuzzy queries) on a single ES node with 16 cores.

Without timeout - No HTTP error (200 only), but the ES node keeps busy for 2 minutes after the last request.
```
Read : 1178 values for sample,request
Percentile 50 : 34941.563
Percentile 90 : 100830.088
Percentile 95 : 102769.465
Percentile 99 : 117499.393
```

With ES timeout (1200ms) - Most requests return HTTP 503.
```
Read : 1186 values for sample,request
Percentile 50 : 1230.859
Percentile 90 : 2101.977
Percentile 95 : 2282.804
Percentile 99 : 2406.509
```

jbgriesner:
To highlight more the behavior of bragi with the ES timeout we made a new test (still at 20 req/sec during 5 minutes) with 2/3 of standard queries and 1/3 of fuzzy queries. 

We tried two scenarios: (i) without timeout and (ii) with timeout.

To sum up, we observed that in case (i) bragi returns only 200 but with large response time, whereas in the case (ii) we had much lower response time but 2/3 of 200 and 1/3 of 503.

We put below the raw response time percentiles for these two scenarios:  
- without ES timeout: 
```
Read : 5997 values for sample,request
Percentile 50 : 13583.895
Percentile 90 : 36698.779
Percentile 95 : 57063.129
Percentile 99 : 93593.558
```
- with ES timeout
```
Read : 5943 values for sample,request
Percentile 50 : 1212.447
Percentile 90 : 2035.398
Percentile 95 : 2228.66
Percentile 99 : 2421.928
```

These tests seem to lead to the conclusion that the ES timeout tend to prevent the bragi overload.

antoine-de:
:ping_pong: did you had time to review this ?

PR #229, opened 2018-06-18
Better search

antoine-de:
Since #211 as it is worsen the geocoder-test, this PR includes #211 and #213 and add some more stuff:

* add a `name` (correctly indexed) the the addresses (so they can also enjoy theboost on the `name`
* do not take the label's size into account (as the label is a generated field). check [the commit](https://github.com/CanalTP/mimirsbrunn/commit/1075633de9f2504bcaf45f10db67f98fdaec62bb) for more information

The geocoder tester (run with [this script](https://github.com/QwantResearch/geocoder-tester/blob/master/run_qwant.sh)) seems happy with this version:

```diff
--- results on master 
+++ results on this PR
-summary for france_autre ==== 1630 failed, 9386 passed, 1 skipped, 1066 deselected in 823.38 seconds ====
-++++++++++ result france_autre: 85% (/) (/11016)
+summary for france_autre === 1568 failed, 9577 passed, 1 skipped, 1066 deselected in 1361.40 seconds ====
+++++++++++ result france_autre: 85% (/) (/11145)
 running test france_fautes
-summary for france_fautes ========== 196 failed, 756 passed, 11131 deselected in 155.50 seconds ==========
-++++++++++ result france_fautes: 79% (/) (/952)
+summary for france_fautes ========== 187 failed, 765 passed, 11260 deselected in 205.41 seconds ==========
+++++++++++ result france_fautes: 80% (/) (/952)
 running test france_admins
-summary for france_admins ============ 12 failed, 38 passed, 12033 deselected in 5.55 seconds ============
-++++++++++ result france_admins: 76% (/) (/50)
+summary for france_admins ============ 6 failed, 44 passed, 12162 deselected in 4.70 seconds =============
+++++++++++ result france_admins: 88% (/) (/50)
 running test france_poi
-summary for france_poi =========== 49 failed, 15 passed, 12019 deselected in 10.65 seconds ============
-++++++++++ result france_poi: 23% (/) (/64)
+summary for france_poi ============ 26 failed, 38 passed, 12148 deselected in 9.39 seconds ============
+++++++++++ result france_poi: 59% (/) (/64)

```

PR #271, opened 2018-10-18
Fix labels normalization

antoine-de:
We don't want to use the size of a label since it is generated.
It has been implemented in https://github.com/CanalTP/mimirsbrunn/pull/229 but I forgot to also remove the norms for the full field (we want to boost a response when we match entire tokens.

The test case for us was that the query `Lyon` was having the results:

```
* Lyon, Coahoma County, Mississippi, United States <---- small village in
mississipi
* Lyon, Métropole de Lyon, Auvergne-Rhône-Alpes, France <---- big french
city with a lot of population
```

with this PR, we now have the french city first

### DIFF
--- a/json/addr_settings.json
+++ b/json/addr_settings.json
@@ -109,6 +109,9 @@
                                 "enabled": false
                             }
                         }
+                    },
+                    "norms": {
+                        "enabled": false
                     }
                 },
                 "full_label": {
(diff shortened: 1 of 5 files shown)


antoine-de:
with that many approvals :stuck_out_tongue_winking_eye:  , is this PR ok for you ?

PR #274, opened 2018-12-07
More stable street id

jbgriesner:
We want streets id as stable as possible.
The solution this PR proposes is the following:
- for streets divided by several ways we take the id of the corresponding relation (tagged "type=associatedStreet") with the prefix "street:osm:rel:<id>"
- for all other streets we select the min id of the corresponding ways with the prefix "street:osm:way:<id>"

### DIFF
--- a/src/osm_reader/street.rs
+++ b/src/osm_reader/street.rs
@@ -85,6 +85,7 @@ pub fn streets(
 
     for rel in objs_map.iter().filter_map(|(_, obj)| obj.relation()) {
         let way_name = rel.tags.get("name");
+
         for ref_obj in &rel.refs {
             use mdo::option::*;
             let objs_map = &objs_map;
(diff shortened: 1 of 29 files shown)


src/osm_reader/street.rs
```diff
@@ -99,7 +99,7 @@ pub fn streets(
                 way_name =<< way_name.or_else(|| way.tags.get("name"));
                 let admin = get_street_admin(admins_geofinder, objs_map, way);
                 ret ret(street_list.push(mimir::Street {
-                    id: way.id.0.to_string(),
+                    id: format!("street:osm:rel:{}", rel.id.0.to_string()),
```
amatissart: Admins from Cosmogony currently have ids such as `admin:osm:relation:9533`
Instead of `rel`, I would use `relation` here too for consistency.

TeXitoi: We already use `admin:osm:relation:215390`, so `relation` is definitively the good way.

TeXitoi:
Thanks

PR #281, opened 2018-12-29
move to rust 2018

antoine-de:
There are pending PR, and they likely will be in conflict with those changes, but since they have not been active for a while I think it's a good time to move to rust 2018.

### DIFF
--- a/src/admin_geofinder.rs
+++ b/src/admin_geofinder.rs
@@ -140,7 +140,7 @@ impl AdminGeoFinder {
     }
 
     /// Iterates on all the admins with a not None boundary.
-    pub fn admins<'a>(&'a self) -> Box<Iterator<Item = Admin> + 'a> {
+    pub fn admins<'a>(&'a self) -> Box<dyn Iterator<Item = Admin> + 'a> {
         let iter = self
             .admins
             .get(&Rect::from_float(
@@ -159,7 +159,7 @@ impl AdminGeoFinder {
     }
 
     /// Iterates on all the `Rc<Admin>` in the structure as returned by `get`.
-    pub fn admins_without_boundary<'a>(&'a self) -> Box<Iterator<Item = Arc<Admin>> + 'a> {
+    pub fn admins_without_boundary<'a>(&'a self) -> Box<dyn Iterator<Item = Arc<Admin>> + 'a> {
         let iter = self
             .admins
             .get(&Rect::from_float(
--- a/src/osm_reader/poi.rs
+++ b/src/osm_reader/poi.rs
@@ -28,9 +28,9 @@
 // https://groups.google.com/d/forum/navitia
 // www.navitia.io
 
-extern crate mimir;
-extern crate osm_boundaries_utils;
-extern crate osmpbfreader;
+use mimir;
+use osm_boundaries_utils;
+use osmpbfreader;
 
 use self::osm_boundaries_utils::build_boundary;
 use super::osm_utils::get_way_coord;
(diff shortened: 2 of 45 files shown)


src/osm_reader/poi.rs
```diff
@@ -68,9 +68,9 @@ impl Default for PoiConfig {
     }
 }
 impl PoiConfig {
-    pub fn from_reader<R: io::Read>(r: R) -> Result<PoiConfig, Box<Error>> {
-        let mut res: PoiConfig = try!(serde_json::from_reader(r));
-        try!(res.check());
+    pub fn from_reader<R: io::Read>(r: R) -> Result<PoiConfig, Box<dyn Error>> {
+        let mut res: PoiConfig = r#try!(serde_json::from_reader(r));
```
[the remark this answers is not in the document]
antoine-de: I tried [rerast](https://github.com/google/rerast/blob/master/COOKBOOK.md#replace-trysomething-with-something) but I did not manage to use it so I had to update them manually :roll_eyes:

src/admin_geofinder.rs
```diff
@@ -140,7 +140,7 @@ impl AdminGeoFinder {
     }
 
     /// Iterates on all the admins with a not None boundary.
-    pub fn admins<'a>(&'a self) -> Box<Iterator<Item = Admin> + 'a> {
+    pub fn admins<'a>(&'a self) -> Box<dyn Iterator<Item = Admin> + 'a> {
```
TeXitoi: Should use `impl Iterator<Item = Admin>`

src/admin_geofinder.rs
```diff
@@ -159,7 +159,7 @@ impl AdminGeoFinder {
     }
 
     /// Iterates on all the `Rc<Admin>` in the structure as returned by `get`.
-    pub fn admins_without_boundary<'a>(&'a self) -> Box<Iterator<Item = Arc<Admin>> + 'a> {
+    pub fn admins_without_boundary<'a>(&'a self) -> Box<dyn Iterator<Item = Arc<Admin>> + 'a> {
```
TeXitoi: Should use `impl Iterator`

tests/tests.rs
```diff
@@ -148,7 +148,7 @@ impl<'a> ElasticSearchWrapper<'a> {
         &self,
         word: &str,
         predicate: F,
-    ) -> Box<Iterator<Item = mimir::Place> + 'b>
+    ) -> Box<dyn Iterator<Item = mimir::Place> + 'b>
```
TeXitoi: `impl Iterator`

tests/tests.rs
```diff
@@ -159,7 +159,7 @@ impl<'a> ElasticSearchWrapper<'a> {
...
```
TeXitoi: `impl Iterator`

tests/tests.rs
```diff
@@ -171,7 +171,7 @@ impl<'a> ElasticSearchWrapper<'a> {
         word: &str,
         predicate: F,
         search_on_global_stops: bool,
-    ) -> Box<Iterator<Item = mimir::Place> + 'b>
+    ) -> Box<dyn Iterator<Item = mimir::Place> + 'b>
```
TeXitoi: `impl Iterator`

tests/tests.rs
```diff
@@ -221,19 +221,19 @@ impl<'a> ElasticSearchWrapper<'a> {
...
                                 .filter(predicate),
                         )
-                            as Box<Iterator<Item = mimir::Place>>)
+                            as Box<dyn Iterator<Item = mimir::Place>>)
                     }
                     _ => None,
                 }
             })
-            .unwrap_or(Box::new(None.into_iter()) as Box<Iterator<Item = mimir::Place>>)
+            .unwrap_or(Box::new(None.into_iter()) as Box<dyn Iterator<Item = mimir::Place>>)
```
TeXitoi: Some tricks may be needed here to switch to `impl Iterator`

PR #283, opened 2019-01-15
remove Box<dyn Iterator<Item = _>>

TeXitoi:


TeXitoi:
Easier to read with "Hide whitespace changes"

PR #286, opened 2019-01-22
Drop iron/rustless and migrate to actix

### LINKED ISSUE #180: Better API framework
I open an issue about the fact that [rustless](https://github.com/rustless/rustless), the framework used in bragi is dead and we should migrate from it.

The alternative I see (don't hesitate if there are others):
### [Rocket](https://crates.io/crates/rocket)
the best one as I see it, but still stuck in nightly (for [quite a long time](https://github.com/SergioBenitez/Rocket/issues/19) :sob: )

### [Gotham](https://crates.io/crates/gotham)

fairly new but async and seems nice, even if argument parsing does not seems as shiny as rocket.
Since it's new we don't really know if it will die as rustless or not...

### [Rouille](https://crates.io/crates/rouille)
quite low level, I don't know it well

### [Iron](https://crates.io/crates/iron)
the more used, but quite old and not as shiny as the others

## which one should be chosen ?

* based on crates download (there is quite some differences :astonished: ) , I'd say Iron > Rouille > Rocket > Gotham
* based on the project activity, I'd say Rocket > Gotham > Rouille > Iron
* based on the number of core contributor (> +1000Loc) I'd say Iron | Gotham (2 main contributors) > Rouille | Rocket (only one)
* Based on my preferences I'd say Rocket :rocket: 

what's your though on this ?

antoine-de:
:tada: :man_dancing:  Migrate the bragi api from Iron/rustless to actix :man_dancing: :tada: 

fixes #180 

I think this is now way easier to maintain, actix is quite a nice framework to work with, and this will make it possible to have async ES calls.

Note:
Since the argument parser is now stricter, this PR is breaking the api on several points:
* the POSTed Geosjon shape should now be valid geojson (this was not the case before)
* the array parameters (`pt_dataset` and `type`) can no longer be parsed "à la oneagain". Before we could do:
  * `type[]=toto&type[]=tata`
  * `type=toto&type=tata`
  * `type=[toto,tata]`
now only `type[]=toto&type[]=tata` is accepted.

Since those are improvement and on not overly used part of the API, I think this can be accepted (I need to do a PR on navitia to correctly handle the `pt_dataset` though.

Also to be noted: lots of error messages have changed. This is not a breaking change, but It still need to be noted. I tried to test the new errors (mainly in `tests/canonical_import_process_test.rs`)

TODO:
 - [x] run geocoder tester (same number of erreors)
 - [x] check prometheus for regressions (ok but handler that are now a real path: `status` => `/status` and content are gziped (but no pb for prom))
 - [x] benchmark (actix is a bit slower than iron, but by a very small margin)
 - [x] patch navitia for the `pt_dataset`s handling ( https://github.com/CanalTP/navitia/pull/2662 )

TODO for other PRs:
 - [ ] better argument validation errors. There are lots of room to improve the errors on argument parsing (there is not the wrongly formatted field in the response for example).


samscott89:
Hey! Just dropping by after suggesting serde_qs in the serde_urlencoded issue. I'm a big fan of actix, so I thought it might be nice to create an `actix` feature in `serde_qs` for ease of use.

Docs: [QsQuery](https://docs.rs/serde_qs/0.4.5/serde_qs/actix/struct.QsQuery.html).
(I haven't yet tested, so would be good to know if works as expected).

antoine-de:
great, I'll try it asap, thanks a million!

libs/bragi/src/prometheus_middleware.rs
```diff
@@ -0,0 +1,68 @@
...
+    fn default() -> Self {
+        PrometheusMiddleware {}
+    }
+}
+
+impl<S> Middleware<S> for PrometheusMiddleware {
+    fn start(&self, req: &HttpRequest<S>) -> Result<Started> {
+        HTTP_REQ_HISTOGRAM
+            .get_metric_with(&labels! {
+                "handler" => req.path(),
```
amatissart: `handler` will now take any value from the path, including non-existing routes.
The previous behavior used to consider validated routes only.  

Is that a problem, considering that prometheus labels are not supposed to contain unbounded sets of values ?

antoine-de: good catch!

I don't really know how to change this.

Do you think it's a problem @kinnou02 ?

antoine-de: hum I did not manage to bypass this with [scopes](https://docs.rs/actix-web/0.7.18/actix_web/struct.Scope.html), but we can activate the prometheus handle resource by resource

like:

```rust
    App::with_state(ctx)
        .resource("/", |r| r.f(entry_point))
        .resource("/autocomplete", |r| {
            r.middleware(prometheus_middleware::PrometheusMiddleware::default());
            r.method(http::Method::GET).with(autocomplete);
            r.method(http::Method::POST)
                .with_config(post_autocomplete, |(_, _, json_cfg)| {
                    json_cfg.error_handler(|err, _req| {
                        ActixError::InvalidJson(format!("{}", err)).into()
                    });
                });
        })
        .resource("/status", |r| {
            r.middleware(prometheus_middleware::PrometheusMiddleware::default());
            r.with(status)
        })
```

the avantage is that we can not trace some routes like the `status` or `metrics`, the downside is that we can forget to add it on new routes :disappointed: 

what do you think @amatissart @kinnou02 ?

antoine-de: another solution would be to set the resource name at creation and use this name instead of the `path` for prometheus.

This way we can have the same name as before (`autocomplete` and not `/autocomplete`), but as the previous solution, we need to explicitly set this name (it's not that bad, there are not that many routes)

libs/bragi/src/server.rs
```diff
@@ -0,0 +1,52 @@
...
+            r.f(default_404)
+        })
+}
+
+pub fn runserver() {
+    let args = Args::from_args();
+    let ctx: Context = (&args).into();
+    server::new(move || create_server(ctx.clone()))
+        .bind(&args.bind)
+        .unwrap()
```
TeXitoi: Maybe just return an Err

libs/bragi/src/prometheus_middleware.rs
kinnou02: I like the idea of settings the name explicitly

antoine-de: ok for the explicit name.

Is it ok to have an empty name or do you want an "unknown" by default ?

libs/bragi/src/routes/autocomplete.rs
```diff
@@ -0,0 +1,149 @@
...
+}
+
+impl JsonParams {
+    fn get_es_shape(&self) -> Result<Vec<(f64, f64)>, model::BragiError> {
+        match &self.shape {
+            GeoJson::Feature(f) => {
+                match &f.geometry {
+                    Some(geom) => {
+                        match &geom.value {
+                            geojson::Value::Polygon(p) => {
```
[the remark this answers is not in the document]
antoine-de: nop it was messy because of the thrice enum/value combinaison.

If you want I can still try, but I though it was not worth it, especially since this version is done and we would loose the detailed errors.

PR #287, opened 2019-01-25
Autocomplete queries with "lang" parameter

amatissart:
This PR adds a new "lang" parameter on autocomplete endpoints. It's used to query language fields (implemented in #285 for admins), and format the response with localized names and labels.

In this PR, a single "lang" value is accepted in the query.

**Changes in ES query:**
 * use multi-match to query both default and localized fields
 * boosts ratio on names/labels fields are modified, to focus on matching names (keeping the same sum, so the score should not be too different). 
According to my tests, this is needed to reduce the importance of localized labels (with their own specific terms) in tf-idf calculation.
 * use a "replace" mode on weight boost (so that the weight importance does not depend on the query itself). The impact of "weight" is more predictable this way. 
We should be careful the calibration is ok though. I mostly tested it with all admins from cosmogony on Europe. It may not be suitable if only cities (with smaller population values) are imported.

**Changes in response format:**
 * replace "name" and "label" with the user language
 * remove (potentially large) "labels" and "names" in "administrative_regions" list


libs/bragi/src/api.rs
```diff
@@ -316,10 +317,11 @@ impl ApiEndPoint {
                     shape_param(params);
                     types_param(params);
                     timeout_param(params);
+                    params.opt_typed("lang", json_dsl::string());
```
antoine-de: hum if the param `lang` will likely be an array in the future (and is treaty as an array afterward), I'd prefer to already name it `lang[]`.
If it's bothersome to parse it as an array with `valico` (what a pain 🙄 ) no problem just name it `lang[]` and I'll handle it as an array when rebasing my PR #286

tests/canonical_import_process_test.rs
```diff
@@ -164,6 +167,28 @@ fn melun_test(bragi: &BragiHandler) {
     assert_eq!(poi_addr["city"], "Melun");
 }
 
+fn lang_test(bragi: &BragiHandler) {
+    let all_francia = bragi.get("/autocomplete?q=Francia&lang=es");
+    let result = all_francia.first().unwrap();
+    assert_eq!(result["name"], "Francia");
```
antoine-de: you should also test somewhere that the admin have been cleaned up

libs/bragi/src/query.rs
```diff
@@ -156,15 +159,38 @@ fn build_query(
         .with_boost(30.)
         .build();
 
+    let format_names_field = |lang| format!("names.{}", lang);
+    let format_labels_field = |lang| format!("labels.{}", lang);
+    let format_labels_prefix_field = |lang| format!("labels.{}.prefix", lang);
+
+    let build_multi_match =
+        |default_field: &str, lang_field_formatter: &Fn(&'a &'a str) -> String| {
```
antoine-de: isn't there a double `&'a` ?

libs/bragi/src/model.rs
```diff
@@ -212,6 +226,27 @@ fn get_citycode(admins: &Vec<Arc<mimir::Admin>>) -> Option<String> {
         .map(|admin| admin.insee.clone())
 }
 
+fn serialize_administrative_regions<'a, S>(
```
antoine-de: can you add a comment explaining this serializer ?

tests/canonical_import_process_test.rs
amatissart: done

libs/bragi/src/model.rs
amatissart: done

libs/bragi/src/query.rs
amatissart: hum, it seems to be required, when iterating on a slice of `&str`

libs/bragi/src/api.rs
amatissart: I have mixed feelings about this.  
`lang` is the parameter used by geocoder-tester and by other other geocoders. So handling the case with a single `lang` could still be useful :grin:
On the other side, I don't want to break the consistency of your brand new API :smirk:

antoine-de: in the future, will we have multiple values for `lang` ?

amatissart: That's very likely. Typically, a search in Breton would need to retrieve french names too, as a fallback.

antoine-de: ok and what about `lang` for the moment, and if we need more complex stuff, we'll read the browser's lang header or a `langs[]` / `fallback_lang[]` / ... parameter ?

amatissart: Sounds good. Especially as another approach could be to implement these fallback rules at index time, or as a server-side configuration.

libs/bragi/src/model.rs
```diff
@@ -94,6 +94,7 @@ pub struct GeocodingResponse {
     // pub state: Option<String>,
     // pub country: Option<String>,
     // pub geohash: Option<String>,
+    #[serde(serialize_with = "serialize_administrative_regions")]
```
antoine-de: ahh, I understand why you need this, but would'nt it be better to use the opportunity to remove the leak of `mimir` object here ?

An `model::AssociatedAdministativeRegion` would be way better.

we'd need to check for retro-compatibility breakage though

amatissart: I agree there must be a better way.  
Ideally, I would also need to serialize the associated admins with a language context, to be able to return the appropriate names and labels in the response.

So you would create a new struct with explicit fields ? And possibly duplicate some of the serde rules ?

antoine-de: yes a new struct with explicit fields would be way easier to control.

we need to check that no one uses those fields though

what do you think @kinnou02 @TeXitoi ?

kinnou02: You want to create a `model::Admin` that implements `From<mimir::Admin>`? 
Sound good

antoine-de: more a `model::AssociatedAdmin` (because we already have a `model::Admin` but yes, that's the idea.

amatissart: I just pushed a version with a new struct `AssociatedAdmin` .
It makes easier to implement `from_with_lang` for all objects : now associated admins are translated with the user language.

What do you think ? Any problem with retro-compatibility ?

libs/bragi/src/model.rs
```diff
@@ -72,6 +73,57 @@ pub struct Properties {
...
 
+#[derive(Serialize, Debug)]
+pub struct AssociatedAdmin {
+    pub id: String,
+    pub insee: String,
+    pub level: u32,
+    pub label: String,
+    pub name: String,
+    pub zip_codes: Vec<String>,
+    pub weight: f64,
```
antoine-de: should we output this ?

libs/bragi/src/model.rs
```diff
@@ -72,6 +73,57 @@ pub struct Properties {
...
+    pub bbox: Option<geo::Bbox<f64>>,
+    #[serde(default)]
+    pub zone_type: Option<cosmogony::ZoneType>,
+    #[serde(default)]
+    pub parent_id: Option<String>, // id of the Admin's parent (from the cosmogony's hierarchy)
+    #[serde(default)]
+    pub codes: Vec<mimir::objects::Code>,
+}
+
+impl FromWithLang<&mimir::Admin> for AssociatedAdmin {
```
antoine-de: open question: can't you take it by value here ? since the other from move the values, wouldn't it make it possible to remove lot's of `clone()` ?

libs/bragi/src/model.rs
antoine-de: seems ok for navitia, it seems it does not use lots of fields:
https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/autocomplete/geocodejson.py#L109-L118

it's the only part that reads associated admins in navitia @kinnou02 ?

kinnou02: I think so

libs/bragi/src/model.rs
amatissart: In mimir objects, admins are `Vec<Arc<Admin>>`. 
So I suppose we would need to use [Arc::try_unwrap](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.try_unwrap) to move the admin, assuming there is no other conflicting reference.
It seems to be ok here, as `Arc` comes from the deserialization of the ES response. 
But is that not a bit clumsy ? :thinking:

antoine-de: oups I forgot those Arc, no leave it like this, no problems!

libs/bragi/src/model.rs
amatissart: I removed `weight` from AssociatedAdmin

amatissart:
Some details about performance impact :

I ran imports (Europe) and complete runs of https://github.com/QwantResearch/geocoder-tester/, before and after this change.

Geocoder-tester scores are not very relevant in this case, as most tests that are using `lang` also expect a `country` field in the response (not provided by bragi)

**master (4a9507a)**

image


```
report on 'geocoder-tester'
queries made on http://*.*.*.*:4000/autocomplete | version = v1.5.2-6-g4a9507a 

 region | category | failed | total | ratio |    duration    
--------|----------|--------|-------|-------|----------------
   _    |  fuzzy   |  277   |  993  |  72%  | 0:01:28.080000 
   _    |  admins  |  3393  | 4168  |  19%  | 0:03:23.560000 
   _    |   poi    |  8932  | 9521  |  6%   | 0:13:37.630000 
   _    |  autre   |  1810  | 11508 |  84%  | 0:06:50.560000 
```

---

**bragi-i18n**
image

```
report on 'geocoder-tester'
queries made on http://*.*.*.*:4000/autocomplete | version = v1.5.2-11-g4505821 

 region | category | failed | total | ratio |    duration    
--------|----------|--------|-------|-------|----------------
   _    |  fuzzy   |  264   |  993  |  73%  | 0:01:23.710000 
   _    |  admins  |  3326  | 4168  |  20%  | 0:03:32.170000 
   _    |   poi    |  8928  | 9521  |  6%   | 0:13:46.860000 
   _    |  autre   |  1788  | 11508 |  84%  | 0:07:02.970000 

```

PR #288, opened 2019-02-07
reverse: Fix proximity score to always return the closest address

amatissart:
As a Qwant Maps user pointed out (https://github.com/QwantResearch/qwantmaps/issues/28), the `/reverse` endpoint may return a wrong address is some cases. This is due to the gaussian decay function that returns **the exact same score** for multiple addresses that are only a few meters apart.

This PR suggests to replace the current function with an [exponential decay function](https://www.elastic.co/guide/en/elasticsearch/guide/2.x/decay-functions.html#img-decay-functions).

Note that this change is only about the `build_proximity_with_boost` function that is defined in `rubber.rs` and used by the reverse geocoder.  
A [similar function exists](https://github.com/CanalTP/mimirsbrunn/blob/master/libs/bragi/src/query.rs#L103) in the autocomplete query and keeps unchanged.


antoine-de:
do you want it for the next release ?

amatissart:
As you wish. There's no rush on this.

Issue #282, opened 2019-01-10
Migration to the geo-type crates


antoine-de:
For the moment mimir depends on an old `geo` crate, we should move the dependency on the new [https://crates.io/crates/geo-types](`geo-types`) crate.

PR #292, opened 2019-02-22
 update cosmogony to get new streaming capabilities 

antoine-de:
this brings some more changes because of quite a big `geo` update.


antoine-de:
fixes #282

PR #293, opened 2019-02-22
correct timeout handling

antoine-de:
the timeout handling has been broken by
https://github.com/CanalTP/mimirsbrunn/pull/286

this fixes it by parsing the timeout as an integer as before (and not a
duration)

Note: without this, the navitia integration is broken and maybe others too. poke @amatissart

### DIFF
--- a/libs/bragi/src/routes/autocomplete.rs
+++ b/libs/bragi/src/routes/autocomplete.rs
@@ -49,7 +49,8 @@ pub struct Params {
     limit: u64,
     #[serde(default)]
     offset: u64,
-    timeout: Option<Duration>,
+    /// timeout in milliseconds
+    timeout: Option<u64>,
     lat: Option<f64>,
     lon: Option<f64>,
     #[serde(default, rename = "type")]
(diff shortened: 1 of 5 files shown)


libs/bragi/src/routes/autocomplete.rs
```diff
@@ -73,6 +74,9 @@ impl Params {
     fn langs(&self) -> Vec<&str> {
         self.lang.iter().map(|l| l.as_str()).collect()
     }
+    fn timeout(&self) -> Option<Duration> {
+        self.timeout.map(Duration::from_secs)
```
amatissart: Don't we want to parse it as milliseconds ?
as before in https://github.com/CanalTP/mimirsbrunn/blob/v1.7.0/libs/bragi/src/api.rs#L91

antoine-de: oups you're right 🤦‍♂️

libs/bragi/src/routes/autocomplete.rs
```diff
@@ -49,7 +49,8 @@ pub struct Params {
     limit: u64,
     #[serde(default)]
     offset: u64,
-    timeout: Option<Duration>,
+    /// timeout in seconds
+    timeout: Option<u64>,
```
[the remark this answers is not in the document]
antoine-de: gnagnagna

PR #294, opened 2019-02-26
add some retrocompatibility on some admin's id

antoine-de:
navitia needs some id to be formated in an uggly way :wink:

### DIFF
--- a/src/bin/cosmogony2mimir.rs
+++ b/src/bin/cosmogony2mimir.rs
@@ -44,7 +44,12 @@ use std::collections::BTreeMap;
 use structopt::StructOpt;
 
 trait IntoAdmin {
-    fn into_admin(self, _: &BTreeMap<ZoneIndex, String>, langs: &[String]) -> Admin;
+    fn into_admin(
+        self,
+        _: &BTreeMap<ZoneIndex, String>,
+        langs: &[String],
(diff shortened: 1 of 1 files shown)


src/bin/cosmogony2mimir.rs
```diff
@@ -59,15 +64,28 @@ fn get_weight(tags: &osmpbfreader::Tags, center_tags: &osmpbfreader::Tags) -> f6
...
         let zip_codes = admin::read_zip_codes(&self.tags);
         let label = self.label;
         let weight = get_weight(&self.tags, &self.center_tags);
         let center = self.center.map_or(mimir::Coord::default(), |c| {
             mimir::Coord::new(c.lng(), c.lat())
         });
-        let format_id = |id| format!("admin:osm:{}", id);
+        let format_id = |id| {
+            if french_id_retrocompatibility && insee != "" {
+                // for retrocompatibity reasons, Kisio Digital needs the
```
pbougue: minor: Better mentioning Navitia than unknown Kisio Digital (repo is CanalTP, etc.)

antoine-de:
I'm not a fan of generating one kind of id for each country.

The communication between system should be based on other ids, because we can't do a universal way to link data.

PR #295, opened 2019-03-05
Use streaming capabilities of cosmogony to reduce memory usage

kinnou02:
We read the file two times now, it doesn't seem to have any impacts on performance:

Before:
   duration: 4.25 min
   memory: 10 go

After:
   duration: 4.17 min
   memory: 250 mo

I'am not a fan of the `ZoneWeight` struct, I would like to create the map and the max_weight in one shot, but without having that struct, and I don't know how to do that without reading one more time :D

### DIFF
--- a/src/bin/cosmogony2mimir.rs
+++ b/src/bin/cosmogony2mimir.rs
@@ -141,12 +141,13 @@ fn index_cosmogony(args: Args) -> Result<(), Error> {
     info!("building maps");
 
     let mut max_weight = 0.0;
-    let cosmogony_id_to_osm_id: BTreeMap<_, _> = read_zones(&args.input)?
-        .map(|z| {
-            max_weight = f64::max(max_weight, get_weight(&z.tags, &z.center_tags));
-            (z.id.clone(), z.osm_id.clone())
-        })
-        .collect();
+    let mut cosmogony_id_to_osm_id = BTreeMap::new();
+    for z in read_zones(&args.input)? {
+        max_weight = f64::max(max_weight, get_weight(&z.tags, &z.center_tags));
+        cosmogony_id_to_osm_id.insert(z.id.clone(), z.osm_id.clone());
+    }
+    let max_weight = max_weight;
+    let cosmogony_id_to_osm_id = cosmogony_id_to_osm_id;
 
     info!("importing cosmogony into Mimir");
     let admins = read_zones(&args.input)?.map(|z| {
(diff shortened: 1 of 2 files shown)


src/bin/cosmogony2mimir.rs
```diff
@@ -116,42 +125,53 @@ impl IntoAdmin for Zone {
...
-        .map(|z| (z.id.clone(), z.osm_id.clone()))
-        .collect();
+fn index_cosmogony(args: Args) -> Result<(), Error> {
+    info!("building maps");
 
-    let mut admins: Vec<_> = cosmogony
-        .zones
-        .into_iter()
+    let cosmogony_id_to_osm_id: BTreeMap<_, _> = read_zones(&args.input)?
         .map(|z| {
```
antoine-de: it's indeed a pain to have this structure, instead of having the max aside the cosmogony_id_to_osm_id map.

@TeXitoi do you have an idea on how to make this better ?

src/bin/cosmogony2mimir.rs
```diff
@@ -39,16 +39,23 @@ use mimir::objects::Admin;
 use mimir::rubber::{IndexSettings, Rubber};
 use mimirsbrunn::osm_reader::admin;
 use mimirsbrunn::osm_reader::osm_utils;
-use mimirsbrunn::utils::normalize_admin_weight;
+use mimirsbrunn::utils::normalize_weight;
 use std::collections::BTreeMap;
 use structopt::StructOpt;
 
+#[derive(StructOpt, Debug)]
```
antoine-de: you don't need structopt here

src/bin/cosmogony2mimir.rs
kinnou02: I could do something like this, but I don't think it's very rustic:
```
let mut max_weight = 0.0;
let cosmogony_id_to_osm_id: BTreeMap<_, _> = read_zones(&args.input)?
        .map(|z| {
            max_weight = f64::max(max_weight, get_weight(&z.tags, &z.center_tags));
            (z.id.clone(), z.osm_id.clone(),)
        })
        .collect();
```

antoine-de: not very rustic, but not than bad, I think I prefer this version.

TeXitoi: This version seems OK. It can also be simply done with a for loop (sometime, a for loop is simpler :smile: )

PR #296, opened 2019-03-19
add labels/names fields

jbgriesner:
This PR is the extension of https://github.com/CanalTP/mimirsbrunn/pull/285 for POI.
Basically it simply adds the 'names' and 'labels' fields to the POI object.
The index POI settings are changed accordingly to these changes.
A basic test comes along the PR.

### DIFF
--- a/json/poi_settings.json
+++ b/json/poi_settings.json
@@ -1,6 +1,7 @@
 {
     "template": "munin_poi_*",
     "settings": {
+		"refresh_interval" : "60s",
         "analysis": {
             "filter": {
                 "prefix_filter": {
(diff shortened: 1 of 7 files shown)


src/utils.rs
```diff
@@ -41,6 +41,39 @@ pub fn format_label(admins: &[Arc<mimir::Admin>], name: &str) -> String {
     }
 }
 
+pub fn format_international_labels(
```
jbgriesner: We will use this function in our specific POI importer tool [here](https://github.com/jbgriesner/fafnir/blob/1c9afdd9b7ef40160c242b6c8d1f77419042b665/src/lib.rs#L215).
If you think it's outside of your scope or if you'll not use it, then we can just put it directly into our tool (fafnir).

json/poi_settings.json
```diff
@@ -1,6 +1,7 @@
 {
     "template": "munin_poi_*",
     "settings": {
+		"refresh_interval" : "60s",
```
amatissart: indent use tabs here, instead of spaces

src/utils.rs
```diff
@@ -41,6 +41,39 @@ pub fn format_label(admins: &[Arc<mimir::Admin>], name: &str) -> String {
...
+    poi_names: &mimir::I18nProperties,
+    default_poi_name: &str,
+    default_poi_label: &str,
+    langs: &[String],
+) -> mimir::I18nProperties {
+    let labels = langs
+        .iter()
+        .filter_map(|ref lang| {
+            let local_poi_name = poi_names.get(lang).unwrap_or(default_poi_name);
+            let i18n_poi_label = match admins.iter().position(|adm| adm.is_city()) {
```
amatissart: wouldn't `.find` be more suitable here ?

tests/bragi_poi_test.rs
```diff
@@ -254,3 +254,76 @@ fn poi_from_osm_with_address_addr_test(bragi: &mut BragiHandler) {
...
+    let result = es
+        .rubber
+        .index("munin_poi", &index_settings, std::iter::once(i18n_poi));
+
+    es.refresh();
+
+    let mut bragi = BragiHandler::new(format!("{}/munin", es.host()));
+
+    // We look for the Colisée in spanish
+    let poi = bragi.get("/autocomplete?q=Colisée&lang=es");
```
amatissart: these tests do not ensure that i18n names and labels are properly indexed.
What about `?q=coliseo&lang=es` for example ?

src/utils.rs
```diff
@@ -41,6 +41,40 @@ pub fn format_label(admins: &[Arc<mimir::Admin>], name: &str) -> String {
     }
 }
 
+pub fn format_international_labels(
```
antoine-de: `format_international_poi_label` no ?

antoine-de: the international formating is something that we'll need to work on very soon, I'm fine for letting it unused here, this way it will be easier to change

PR #298, opened 2019-03-25
Object filtering

### LINKED ISSUE #243: Filter the autocomplete results
For now, if you want to filter the search results from Mimir, you can get some inconsistencies. See for instance 
* this issue: https://github.com/CanalTP/mimirsbrunn/issues/230
* this PR, with some solutions and some new issues: https://github.com/CanalTP/mimirsbrunn/pull/235

So, what can we do ?

# Our constraints

## Retrocompat

`autocomplete?q=foo&type=city` should get you only zones with `city` as `zone_type`

## Coherency between question and answer

We have a `type` field in our answer. 
POIs have also a `poi_types` field.

For the sake of consistency, if we have filter on `type` or on `poi_type`, we shoud give the parameter the same name, and it should accept as values only what we can have as value in theses fields in our answer.

## Geocodejson

We need to have a `type` field in the answer for each element.
The specs also recommends some values (`house` for addresses, `street` for streets, etc)

# What do we want

## Filter by type

We want to filter in any kind of Mimir type: zones, address, streets, pois and stop_areas.

We also want to combine these filters to keep many types.

But we should be aware that there is an inconsistency in the current bragi response as `Zones` actually have a `type` which is not `zone` (a state_district has `type` = `state_district`) :shit:

like:

```json=
"geocoding": {
    "citycode": "75",
    "id": "admin:osm:relation:71525",
    "label": "Paris, Île-de-France, France",
    "level": 6,
    "name": "Paris",
    "type": "state_district" // <-- it's not a classic type, it's a ZoneType
}
```

## Filter by subtype

Some Mimir types have subtypes: POIs have `poi_types` and zones have `zone_type` and we should be able to filter on those types too.

As said before, the zones are not consistent in this matter as their `zone_type` is hold by their `type` :stuck_out_tongue_winking_eye: 

# Proposals

## Solution 1

2 parameters:
* `type`
* `poi_type` or `zone_type`

**Solution 1a**:

We fix the coherency issue between the question and the answer: we change the type of the admin to `zone` (or `administrative_region` but I find it less generic as we might want to add non administrative regions like `parc du mercantour` or `disneyland paris`) and add a `zone_type` field in the `Zones`.

This is the cleaner option, but it breaks the retrocompat.

**Solution 1b**:

We do not change the answer, and keep the inconsistency : you ask for `type = zone` and `zone_type = state_district` and you get `type = state_district` 

### how do we handle strange questions?

We need to decide what to do if we only get the `subtype` parameter (but no `type` paramater): `/autocomplete?q=foo&zone_type[]=country`

We could either assume that the user wanted to also filter by type. 

Or we could error instead of trying to guess what the user want.
I prefer this option, because with good error message, we can teach the user how the API works (and as we don't have any doc about the API parameter for now, this is better than to have to read the source code to find the name of the parameter and its accepted values)

## Solution 2

Only one `type` parameter.

It could have as value anything in the usual Mimir response:
* `city`, `state_district`, `state`, etc
* `poi`
* `house`, `street`

We cannot filter by zone (or we have to list all the zone types in the type parameter)

We cannot filter by poi types.

What do you think ?

GuillaumeGomez:
Supersedes #235 and fixes #243.

The point of this PR is to allow to filter over zone types and poi types in a more precise way. Most changes are about that (except a few that are just cleanups 😝 ).

One of the big changes here is the changes of poi_type's id field: it's now indexed as is to allow searches over it.


libs/bragi/src/routes/autocomplete.rs
```diff
@@ -19,16 +19,33 @@ enum Type {
     StopArea,
     #[serde(rename = "street")]
     Street,
+    #[serde(rename = "zone")]
```
antoine-de: can you add a comment saying that `City` is there only for retrocompatibility ? (that's the case no ?)

src/osm_reader/poi.rs
```diff
@@ -116,14 +114,6 @@ impl PoiConfig {
         }
         Ok(())
     }
-    fn convert_id(&mut self) {
```
antoine-de: I don't recall why we needed this, and why we don't need it now, can you explain it a bit ?

tests/bragi_osm_test.rs
```diff
@@ -124,7 +125,7 @@ fn zip_code_admin_test(bragi: &mut BragiHandler) {
 
 fn bbox_admin_test(bragi: &mut BragiHandler) {
     let all_20 = bragi.get("/autocomplete?q=77000 Vaux-le-Pénil");
-    let first_city = all_20.iter().find(|e| get_value(e, "type") == "city");
+    let first_city = all_20.iter().find(|e| get_value(e, "type") == "zone");
```
antoine-de: hum, so now the city change to  `zone` in the output too ?

that's fine for me (way more consistent), but we need to check that this will not break the [navitia](https://github.com/canaltp/navitia) integration

any thoughts on this @TeXitoi @kinnou02 ?

tests/tests.rs
```diff
@@ -210,11 +210,11 @@ impl<'a> ElasticSearchWrapper<'a> {
 }
 
 fn launch_and_assert(
-    cmd: &'static str,
-    args: Vec<std::string::String>,
+    cmd: &str,
+    args: &[std::string::String],
```
antoine-de: 👍

tests/tests.rs
```diff
@@ -325,23 +325,31 @@ pub fn get_values<'a>(r: &'a [Map<String, Value>], val: &'a str) -> Vec<&'a str>
...
-    e.get(val).and_then(|l| l.as_str()).unwrap_or("")
+pub fn get_value<'a>(e: &'a Map<String, Value>, val: &str) -> &'a str {
+    e.get(val).and_then(|l| l.as_str()).unwrap_or_else(|| "")
 }
 
 pub fn get_types(r: &[Map<String, Value>]) -> Vec<&str> {
+    get_given_types(r, "type")
+}
+
+pub fn get_given_types<'a>(r: &'a [Map<String, Value>], key: &str) -> Vec<&'a str> {
```
antoine-de: hum I'm not convinced by this function name, we do not return a type here no ?

libs/bragi/src/model.rs
```diff
@@ -289,25 +292,19 @@ impl FromWithLang<mimir::Admin> for GeocodingResponse {
...
             id: other.id,
             citycode: insee,
-            level: level,
-            place_type: type_,
-            name: name,
-            postcode: postcode,
-            label: label,
+            place_type: "zone".to_owned(),
+            level,
+            zone_type: if zone_type.is_empty() { None } else { Some(zone_type) },
```
antoine-de: can `zone_type` be empty ?

Cargo.lock
```diff
@@ -3,7 +3,7 @@ name = "MacTypes-sys"
 version = "2.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "libc 0.2.49 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.50 (registry+https://github.com/rust-lang/crates.io-index)",
```
antoine-de: why has the cargo.lock changed ?

GuillaumeGomez: I needed to update `cosmology` and updated everything alongside.

libs/bragi/src/model.rs
GuillaumeGomez: Yes and it gave us some issues. :)

tests/tests.rs
GuillaumeGomez: I wanted to keep the name alongside "get_types". What do you suggest instead?

src/osm_reader/poi.rs
GuillaumeGomez: It's duplicated: we prepend "poi_type:" in a type that is already very **clear** about what it is: `{poi_types: [{ id: "poi_type:..." }, ...]}`. We couldn't find out a reason to keep it so I just removed it.

libs/bragi/src/query.rs
```diff
@@ -420,6 +443,14 @@ pub fn autocomplete(
             .map(|v| v.iter().map(|&l| l.into()).collect())
     }
 
+    // Perform parameters validation.
+    if !zone_types.is_empty() && !types.iter().any(|s| *s == "zone") {
+        return Err(BragiError::InvalidParam("zone_type parameter requires to have 'type=zone'"));
```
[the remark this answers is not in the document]
GuillaumeGomez: You're absolutely right, good catch!

libs/bragi/src/routes/autocomplete.rs
GuillaumeGomez: Not as far as I know? We still have some `City`s going around. ;)

GuillaumeGomez:
> Can you add a bit a description in the PR title to explain a bit the changes ? It helps the review.

Done!

> This supersedes #235 and closes #243 right ?

It does indeed. Added it into the description.

> Note: you should run cargo fmt --all to please travis 😜

Urg... Completely forgot about that. :-.

I'll update the PR tomorrow.

src/osm_reader/poi.rs
antoine-de: it is indeed redundant but It seems it was done for retro-compatibility reasons with the old navitia's autocomplete system :disappointed: 

It might be possible to move this retrocompatibility layer in navitia, but I don't know if the navitia team is ok with this.

any thoughts @kinnou02 @TeXitoi ?

libs/bragi/src/routes/autocomplete.rs
```diff
@@ -19,16 +19,33 @@ enum Type {
...
+            Type::Street => "street",
+            Type::Zone => "zone",
+        }
+    }
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(untagged)]
+enum PoiType {
+    Whatever(String),
```
antoine-de: why do you need this enum ? can't you just use a `String` ?

libs/bragi/src/model.rs
antoine-de: can you give me an example where it is empty ? (by curiosity I'm fine with this test)

libs/bragi/src/routes/autocomplete.rs
antoine-de: Hum I might have missed something, but it seems the `type[]=city` just work as `type[]=zone` and not `type[]=zone&zone_type[]=city` as some can expected.

I'm fine with this as my understanding of this subject is that `city` is there only for retro-compatibility reason no ? (thus I think it's worth mentioning it, and explaining in the comments that city==zone)

libs/bragi/src/routes/autocomplete.rs
GuillaumeGomez: In case we want to have a stable Poi API some day. It allows to have things ready for it.

tests/tests.rs
antoine-de: not convinced but maybe `get_values` ?

GuillaumeGomez: There is a function with this name already (in the same file).

src/osm_reader/poi.rs
amatissart: We are not using this importer on our side, so keeping this conversion would be fine.
It makes the API testing a bit more confusing though.

libs/bragi/src/routes/autocomplete.rs
amatissart: yes indeed, `type=zone` is equivalent to the current `type=city`, that should be considered as deprecated.

tests/tests.rs
antoine-de: any idea @amatissart ?

amatissart: well, it seems `get_given_types` does exactly the same thing as `get_values` :rofl: 
So we don't need both.

src/osm_reader/poi.rs
TeXitoi: I personally think that it should be done in navitia side, but I don't have any power to pressure this update on navitia. Removing the prefix might slow down the merge of this PR, maybe doing that in an independent PR is a better idea.

tests/tests.rs
GuillaumeGomez: :rofl: Makes things simpler!

GuillaumeGomez:
Updated (put back the "poi_type:").

tests/bragi_osm_test.rs
kinnou02: Seems good, I will add a test for it in jormun later

src/osm_reader/poi.rs
kinnou02: I agree, we can do this on the jormun side and then change it in mimir.

PR #300, opened 2019-04-10
Add geoshape in elasticsearch to speedup geographic search

kinnou02:
https://www.elastic.co/guide/en/elasticsearch/guide/current/geo-shapes.html
geoshape use geohash to index a shape as `term` it's faster than
than doing complex geographic computing for each point.
Obviously we loose on precision, it's possible to have a few meters of
errors.


antoine-de:
I put this as wip as we need to change rs-es (linked to https://github.com/benashford/rs-es/issues/129) to use this new geohash

antoine-de:
I added the use of the new field in bragi, we need to make benchmarks now

kinnou02:
I have done a few bench, the geoshape implementation is way faster.
my bench using gatling is at 40% of error on master with 1 request per second while geoshape is at 0%, the timing result isn't usable...
On a selected request geoshape is at 78ms when master is at 400ms

libs/mimir/src/objects.rs
```diff
@@ -204,6 +204,8 @@ pub struct Poi {
     pub label: String,
     pub name: String,
     pub coord: Coord,
+    #[serde(skip_deserializing)]
+    pub coord_hash: Option<CoordHash>,
```
antoine-de: my last concern is the naming of this field.

wouldn't `approx_coord` be more explicit ? and I think we should add a comment explaining why there are 2 coord

antoine-de:
this seems ok for me.

anyone wants to also review it or can we merge it ?

amatissart:
I am not familiar with this feature, but it looks good to me

PR #301, opened 2019-04-11
update rs-es and navitia_model

kinnou02:
rs-es has been update to use [reqwest](https://docs.rs/reqwest/0.9.14/reqwest/)
in place of hyper directly. This update allows us to move from the very old hyper that was used :tada:
Reqwest look a lot like hyper 0.10, so there is only a few changes that are mostly cosmetics.

navitia_model has been renamed in `transit_model` and is not published on crates.io.

### DIFF
--- a/libs/mimir/src/rubber.rs
+++ b/libs/mimir/src/rubber.rs
@@ -73,7 +73,7 @@ lazy_static::lazy_static! {
     .unwrap();
 }
 
-fn do_req(resp: reqwest::Response) -> Result<reqwest::Response, EsError> {
+fn check_response(resp: reqwest::Response) -> Result<reqwest::Response, EsError> {
     let mut resp = resp;
     let status = resp.status();
     match status {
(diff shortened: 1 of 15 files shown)


libs/mimir/src/rubber.rs
```diff
@@ -73,11 +73,20 @@ lazy_static::lazy_static! {
     .unwrap();
 }
 
+fn do_req(resp: reqwest::Response) -> Result<reqwest::Response, EsError> {
+    let mut resp = resp;
+    let status = resp.status();
+    match status {
+        StatusCode::OK | StatusCode::CREATED | StatusCode::NOT_FOUND => Ok(resp),
+        _ => Err(EsError::from(&mut resp)),
```
antoine-de: by curiosity, do you know why the EsError mutate the response ?

I'm not a fan of the name of the function. `check_response` ? `wrap_response`  ?

antoine-de: or maybe we can use https://docs.rs/reqwest/0.9.14/reqwest/struct.Response.html#method.error_for_status
but we then need to change a bit more core because of the different error type, so it's up to you

kinnou02: Using `error_for_status()` will also consider 404 as errors, they might be "normal", I took this implementation of `do_req` from rs-es. 
Typically, it will impact this function of rubber: https://github.com/CanalTP/mimirsbrunn/blob/7b81f2204dffa76b4422c0840cf15f12e6784304/libs/mimir/src/rubber.rs#L382-L406
As for the `mut`, this implementation of `from` read the response body, that why `mut` is required

PR #303, opened 2019-04-15
Only use french retrocompatible id for cities

kinnou02:
Some city_district have the same insee than their city: we were loosing
some admins during the import and had an infinite loop when recreating the
hierarchy of addresses.
Now we only construct admin's id with insee for cities, and we always
create the id with `zones_osm_id` to make sure we have coherent id between
admin and their parents

This PR also include a check when constructing the admin hierarchies to prevent infinite loop: if we add an admin that had already been added we stop the loop.

### DIFF
--- a/src/bin/cosmogony2mimir.rs
+++ b/src/bin/cosmogony2mimir.rs
@@ -81,7 +81,7 @@ impl IntoAdmin for Zone {
         });
         let format_id = |id, insee| {
             // for retrocompatibity reasons, Navitia needs the
-            // french admins to have an id with the insee
+            // french admins to have an id with the insee for cities
             match insee {
                 Some(insee) if french_id_retrocompatibility => format!("admin:fr:{}", insee),
                 _ => format!("admin:osm:{}", id),
@@ -92,7 +92,10 @@ impl IntoAdmin for Zone {
             .and_then(|id| zones_osm_id.get(&id))
             .map(|(id, insee)| format_id(id, insee.as_ref()));
         Admin {
-            id: format_id(&self.osm_id, insee.as_ref()),
+            id: zones_osm_id
+                .get(&self.id)
+                .map(|(id, insee)| format_id(id, insee.as_ref()))
+                .expect("unable to find zone id in zones_osm_id"),
             insee: insee.unwrap_or("".to_owned()),
             level: self.admin_level.unwrap_or(0),
             label: label,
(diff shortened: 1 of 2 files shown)


src/bin/cosmogony2mimir.rs
```diff
@@ -143,7 +147,14 @@ fn index_cosmogony(args: Args) -> Result<(), Error> {
     let mut cosmogony_id_to_osm_id = BTreeMap::new();
     for z in read_zones(&args.input)? {
         max_weight = f64::max(max_weight, get_weight(&z.tags, &z.center_tags));
-        let insee = admin::read_insee(&z.tags).map(|s| s.to_owned());
+        let insee = if let Some(zt) = z.zone_type {
```
antoine-de: minor: if think you could have wrote this:

```rust
match z.zone_type {
Some(City) => ...,
_ => None
}
```

src/bin/cosmogony2mimir.rs
```diff
@@ -91,8 +91,12 @@ impl IntoAdmin for Zone {
             .parent
             .and_then(|id| zones_osm_id.get(&id))
             .map(|(id, insee)| format_id(id, insee.as_ref()));
+        let osm_id = self.osm_id.to_string();
         Admin {
-            id: format_id(&self.osm_id, insee.as_ref()),
+            id: zones_osm_id
+                .get(&self.id)
+                .map(|(id, insee)| format_id(id, insee.as_ref()))
```
antoine-de: it's a bit tricky here, the `insee` variable is not the same as the one on the line below, it might be easier to give it a different name.

I don't really understand what you are doing here. can you add a comment ?
how can there not be the admin `ZoneIndex` in zones_osm_id?

kinnou02: You are right the zone should always be found in `zones_osm_id`, I added the `unwrap_or_else` as a safe guard because I don't like the idea of using a `unwrap` that could panic here, mostly because I think it isn't idiomatic, but I can do that if think that it's the good way.

If we keep the `unwrap_or_else` I should'nt pass `None` for the `insee`.

antoine-de: hum for once I'm a bit in favor of `expect` as it should by construction it cannot happen, and I think it will make the code a bit easier to understand (and since it's an import tool, an `expect` is not deadly as it is in bragi)

PR #304, opened 2019-04-30
Localized labels

### LINKED ISSUE #181: Localized label
The mimir label is a nice feature that creates a pretty label to display without any logic outside mimir.

## Address / street Label
In mimir for the moment there is some label only for addresses and streets.

It however works well for one kind of address formating, the french one:
 `{house_number} {street} ({city})`

 If we want localized label (and I think we do :laughing: ), the [opencage](https://github.com/OpenCageData/address-formatting) approach seems nice.

 They have a set of formatting rules for each countries (with even [some custom regexp](https://github.com/OpenCageData/address-formatting/blob/master/conf/countries/worldwide.yaml#L609))

 They might not be perfect, but it's the formatting used by [libpostal](https://github.com/openvenues/libpostal#features) so I think it's a very good starting point.

 One drawback of their label is they are multi-line, but just striping the new lines might be enough for mimir (or we could have a `multiline_label` along the `label`)

 The [opencage](https://github.com/OpenCageData/address-formatting) repository contains only the configuration, they provide a perl implementation (and libpostal have a python one that I've never been able to use :sweat:), we could add a rust implementation :dancers:

 For this to work we need to have a fine grained admin hierarchy (#178 and #176) and we need to know in which country is an object before creating it's label (so either cli based configuration or country lookup)

## Admin label

Admin label is quite basic for the moment too, we [just concatenate the name and the postal code of the admin](https://github.com/CanalTP/mimirsbrunn/blob/master/src/osm_reader/admin.rs#L154)

Even in france it's not very nice since this way there is city duplicates.
eg. the search `torcy` gives:

```
Torcy (77200)
Torcy (71210)
Torcy (62310)
```

so it's quite hard for the user to choose between [torcy in seine et marne](https://www.openstreetmap.org/relation/402764), [torcy in saone et loire](https://www.openstreetmap.org/relation/78000) and [torcy in pas de calais](https://www.openstreetmap.org/relation/1208541)

The opencage approach seems quite address oriented and I have not found a international way to format admins, so I think putting all the admin hierarchy in a coma separated would be not so bad (even if we'd thus loose the postal code :confused:)

## Poi label

I don't really the right way to go for the Pois label (for the moment it's `{name} {city}`), but since a lot of poi can have the same name in a city (`parking` :dizzy_face:) I think there might be room for improvement

# i18n
There is also the question on when this label is created.
Even if mimir does not support custom lang at the query I think one day it will (that's for again another issue :cold_sweat:) we might want to think about this in the label refactoring.

I personnaly don't think that's a good idea to display a french address with a german formatting to a german (`rue spontini 7 paris`) user since it will then be difficult to use the address in france.

So I think we can keep the label creation at import time which is way easier :stuck_out_tongue_closed_eyes:

antoine-de:
use  [opencage's configuration](https://github.com/CanalTP/address-formatter-rs) to have a localized label.

linked to #181, I don't know if we can say that this closes it, and I fell the label needs a little bit more love.

I tried to sum up what this PR does in [this comment](https://github.com/CanalTP/mimirsbrunn/issues/181#issuecomment-499809955)

## Note
* I also removed serde_derive as stated in the serde's documentation since I had some strange behavior with rs-es serde's imports.
* a `country_codes` field has been added to all the objects.
It will makes it easy to add it in the ES index to only search for object in france (and bypass the costly shape filtering), and to add it in the API (not done yet), for the user to use this field (to format the phone number for example).
This field is a list, because of the edge case where some places has more than one (think about [Guadeloupe](https://www.openstreetmap.org/relation/1047206) that has `GP` and `FR`, or [Guam](http://cosmogony.world/#/7.69/144.10/13.13/54742) with `GU` and `US`
* the function `https://github.com/CanalTP/mimirsbrunn/pull/304/commits/d4ee60121dc4016d80a20b3fa0e4b787651a8f13`, used in fafnir has been [move and refactored a bit](https://github.com/CanalTP/mimirsbrunn/pull/304/commits/d4ee60121dc4016d80a20b3fa0e4b787651a8f13) (poke @amatissart )


antoine-de:
I think this PR can now be reviewed

src/labels.rs
```diff
@@ -0,0 +1,375 @@
...
+    address_formatter::FORMATTER
+        .short_addr_format_with_config(
+            place.into_place(admins),
+            address_formatter::Configuration {
+                country_code,
+                ..Default::default()
+            },
+        )
+        .map_err(|e| warn!("impossible to format label: {}", e))
+        .unwrap_or_else(|_| "".to_owned())
```
kinnou02: This seems a bit risky

PR #306, opened 2019-05-07
Add nb to the id to prevent "disappearances" of elements at the same position

GuillaumeGomez:


antoine-de:
I see the point of this change, but:
* Even with the retro-compatibility stuff, I'd like to check if kisio can use it, of if it will break stuff on their side.
* Is adding the house number sufficient here ? Won't we have other problems for different streets too ?

You changed it for bano, but it would be better to do it also for openaddresses too

GuillaumeGomez:
With this, on bano we still have around 12K duplicates. This is not good enough. The only way to efficiently de-duplicate would certainly be to put the street's name inside the id...

kinnou02:
> With this, on bano we still have around 12K duplicates. This is not good enough. The only way to efficiently de-duplicate would certainly be to put the street's name inside the id...

for bano we could use the fantoir, it will be less hideous...

GuillaumeGomez:
What is the "fantoir"?

kinnou02:
https://wiki.openstreetmap.org/wiki/FR:Key:ref:FR:FANTOIR
it's more or less an id of street in France

GuillaumeGomez:
Oh I see! Wouldn't it be better to have something that would work for both openaddresses and bano?

GuillaumeGomez:
So with the id, we have around 2K cases where the position and the number is the same, but not the street name. In those cases, a lot are coming from errors. With this information; is it worth it trying to make something more advanced? With this PR, we handle more than 99.9999% of the addresses.

kinnou02:
I don't think it is worth it.

antoine-de:
if we really don't want duplicates, why don't we use an `uuid` (or even better, let ES handle it by not supplying the id (the only argument for a manual id would be to keep `coord:<uuid>`)) ?

I don't really see where this could cause some problem.

any thoughts @kinnou02 @amatissart @GuillaumeGomez ?

GuillaumeGomez:
Are we supposed to be able to easily find an object or not? Because currently it's possible, using a uuid would make it very complicated.

amatissart:
On our side these ids will be used in URLs (https://www.qwant.com/maps/place/addr:2.307281;48.870237), stored in user favorites, etc.  

So we'd like to keep deterministic stable ids (as far as possible), and ideally not related to a particular data source, to be able to switch or merge other sources in the future.

antoine-de:
I did some tests on navitia, the formatting of the id does not seems to be a problem, we'll be soon able to merge this PR, sorry for the delay!

PR #307, opened 2019-05-23
Output the stop's lines

antoine-de:
When stops have been imported with a ntfs (and not only a gtfs's stops.txt file) we store the stop's lines to output them in bragi.

This will makes it possible to add a nice picto in the search results with the lines passing by the stop. It will makes it easier to differentiate the results at first glance.

Note: the lines are not indexed by ES, so this should not change any results.


libs/mimir/src/objects.rs
```diff
@@ -266,12 +267,43 @@ pub struct PhysicalMode {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub network: Option<Network>,
+    pub physical_mode: Vec<PhysicalMode>,
+}
+
+// we want the lines to be sorted in a way where
+// line-3 is before line-11, so be use a humane_sort
+impl humanesort::HumaneOrder for Line {
+    fn humane_cmp(&self, other: &Self) -> Ordering {
+        self.name.humane_cmp(&other.name)
```
kinnou02: Sorting should be done by using`sort_order`, then `code` and finally `name`.
the reference implementation is here: https://github.com/CanalTP/navitia/blob/dev/source/type/line.cpp#L82-L96

src/bin/ntfs2mimir.rs
```diff
@@ -67,6 +67,54 @@ struct Args {
...
 }
 
+fn get_lines(idx: Idx<navitia::StopArea>, navitia: &transit_model::Model) -> Vec<mimir::Line> {
+    use humanesort::HumaneSortable;
+    let mut lines: Vec<_> = navitia
+        .get_corresponding_from_idx(idx)
+        .into_iter()
+        .map(|l_idx| {
+            let line = &navitia.lines[l_idx];
+            mimir::Line {
```
kinnou02: Don't you think that using `From` would make this easier to read?

antoine-de: I cannot use just `From` because I need the `navitia` object, but I can still hide this behind a trait, you're right

libs/mimir/src/objects.rs
antoine-de: ahhh ok, I'll change this

libs/mimir/src/objects.rs
```diff
@@ -266,12 +267,103 @@ pub struct PhysicalMode {
...
+}
+
+impl FromTransitModel<transit_model::objects::Line> for Line {
+    fn from_transit_model(
+        l_idx: transit_model::collection::Idx<transit_model::objects::Line>,
+        navitia: &transit_model::Model,
+    ) -> Self {
+        let line = &navitia.lines[l_idx];
+        Self {
+            id: line.id.clone(),
```
kinnou02: You need to add the navitia prefix on ids: `line:` for line by example.

antoine-de: you shouldn't this be handled by transit_model ?

PR #320, opened 2019-07-04
Add benches

GuillaumeGomez:
As discussed with @antoine-de, it'd be interesting to be able to track performances of the different crates. To start this effort, I moved a few things into a new "tools" crate to avoid duplicating code. Then I also created a first bench file which can be used as base to add more.


antoine-de:
while this is interesting, I'm a bit troubled by the fact that this works only for nightly rust :confused: 

And I don't really see how we can use the benchmark in our CI.

This being said I think moving the tests helper is nice (even if I'm not a fan of the `tools` name, I don't really have a better idea).

any thoughts on this ?
maybe we can merge this PR as is, but I'm not sure we'll use the benches

GuillaumeGomez:
PR passed on stable. :)

GuillaumeGomez:
And the CI can run nightly without problems. I'm just surprised it's still nightly only though...

antoine-de:
thanks for this!

PR #325, opened 2019-07-10
Migration to actix 1

antoine-de:
This is purely a technical modification, bringing Actix from 0.7 to 1

I think the overall changes in bragi are quite nice, but:
* I failed to have clean tests, so they are a bit messy.
* the prometheus stuff is more complicated in part because the ability to get the ressource from the request has been lost in the migration.


libs/bragi/src/prometheus_middleware.rs
```diff
@@ -1,70 +1,272 @@
-use actix_web::middleware::{Finished, Middleware, Started};
-use actix_web::{HttpRequest, HttpResponse, Result};
-use prometheus::{
-    histogram_opts, labels, opts, register_counter_vec, register_gauge, register_histogram_vec,
+// shamelessly taken from https://github.com/nlopes/actix-web-prom
+// we are unfortunatly not able to directly use this great crate,
+// because  we want to use the anme of the endpoint for retrocompatibility
```
kinnou02:
```suggestion
// because  we want to use the name of the endpoint for retrocompatibility
```

PR #326, opened 2019-07-11
Feature: Handling private indices

riendegris:
This commit adds a flag 'private' to rubber's `publish_index` function
so that a private index is not aliased by global indices, eg munin.

This flag is propagated to poi2mimir's command line, so that we can
import private POIs.

There are also modifications of the elastic search wrapper to allow
queries on non global indices.

### DIFF
--- a/src/bin/osm2mimir.rs
+++ b/src/bin/osm2mimir.rs
@@ -126,7 +126,12 @@ fn run(args: Args) -> Result<(), mimirsbrunn::Error> {
             };
             info!("importing streets into Mimir");
             let nb_streets = rubber
-                .index(&args.dataset, &street_index_settings, streets.into_iter())
+                .index(
+                    &args.dataset,
+                    false,
+                    &street_index_settings,
+                    streets.into_iter(),
+                )
                 .with_context(|_| {
                     format!(
                         "Error occurred when requesting street number in {}",
@@ -144,6 +149,7 @@ fn run(args: Args) -> Result<(), mimirsbrunn::Error> {
         let nb_admins = rubber
             .index(
                 &args.dataset,
+                false,
                 &admin_index_settings,
                 admins_geofinder.admins(),
             )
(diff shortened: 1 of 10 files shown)


src/bin/osm2mimir.rs
```diff
@@ -126,7 +126,12 @@ fn run(args: Args) -> Result<(), mimirsbrunn::Error> {
             };
             info!("importing streets into Mimir");
             let nb_streets = rubber
-                .index(&args.dataset, &street_index_settings, streets.into_iter())
+                .index(
+                    &args.dataset,
+                    false,
```
antoine-de: hum I'm not sure I'm a fan of passing this bool around as it makes the api a bit difficult to read.

Don't you think that if would be easier to add a `publish_private_index` in rubber ? (with some code factorization inside rubber)

tests/poi2mimir_test.rs
```diff
@@ -64,48 +65,66 @@ pub fn poi2mimir_sample_test(es_wrapper: crate::ElasticSearchWrapper<'_>) {
...
+        .and_then(|s| s.keys().cloned().next())
+        .unwrap_or_else(String::new);
 
-    // Making sure that there are as many POI in ES as there are the input file.
-    assert_eq!(es_wrapper.count("_type:poi"), 2);
+    assert_eq!(poi_index_alias, "munin_poi_mti");
 
-    // We test that the POI that was close to an address has been 'attached' to that address,
-    // and inherited its admin and address: (Note that the data have been manipulated to fit)
+    //let aliases = get_first_index_aliases(raw_indexes);
```
antoine-de: you can remove this line

riendegris: :+1:

src/bin/osm2mimir.rs
kinnou02: Or something typed like an enum

riendegris: agreed, using an enum

PR #334, opened 2019-08-06
Scripts

riendegris:
This is a script to help you setup elasticsearch for mimir.

You need to modify the configuration, at least the path to mimirbrunn and cosmogony. These paths are used to call binaries like osm2mimir, bano2mimir, cosmogony2mimir...

You can specify in the configuration file the bano regions, osm and such.

The script will:

- remove docker if it is running (identified by ES_NAME in the configuration)
- start a new docker
- download the data for bano, ntfs, osm (configurable)
- execute all the xxx2mimir

### DIFF
--- a/scripts/import2mimir.sh
+++ b/scripts/import2mimir.sh
@@ -334,23 +334,23 @@ restart_docker_es 9200
 download_osm
 [[ $? != 0 ]] && { log_error "Could not download osm. Aborting"; exit 1; }
 
+download_bano
+[[ $? != 0 ]] && { log_error "Could not download bano. Aborting"; exit 1; }
+
+download_ntfs
+[[ $? != 0 ]] && { log_error "Could not download ntfs. Aborting"; exit 1; }
+
 generate_cosmogony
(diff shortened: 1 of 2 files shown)


scripts/import2mimir.sh
```diff
@@ -0,0 +1,356 @@
...
+download_bano
+[[ $? != 0 ]] && { log_error "Could not download bano. Aborting"; exit 1; }
+
+import_bano
+[[ $? != 0 ]] && { log_error "Could not import bano into mimir. Aborting"; exit 1; }
+
+import_osm
+[[ $? != 0 ]] && { log_error "Could not import osm into mimir. Aborting"; exit 1; }
+
+download_ntfs
```
kinnou02: You might want to download all data before starting to import them, this way it will fail faster

riendegris: :+1:

antoine-de:
isn't it the same thing as https://github.com/QwantResearch/docker_mimir/ ?

PR #335, opened 2019-08-06
extern crate -> rust 2018

riendegris:
I need a tool that targets bragi's web API.
This tool needs to be aware of the parameters used by bragi.
These parameters were not publicly exposed.

There is also a function to generate query parameters, so that they can be easily incorporated in, eg reqwest:

```
let bragi_params = Params { ... };
let query_params = autocomplete_params_to_query_params(bragi_params);
Client::new().post(...).query(query_params).send()...
```

An alternative to the function might be to use Into<Vec<String, String>> ?

An alternative would be to use OpenAPI, to open the bragi API to other languages...

### DIFF
--- a/libs/bragi/src/prometheus_middleware.rs
+++ b/libs/bragi/src/prometheus_middleware.rs
@@ -3,9 +3,8 @@
 // because  we want to use the name of the endpoint for retrocompatibility
 // (and as a side effect we also added the 'in flight' queries (but for this we could have used the Registry))
 
-use std::marker::PhantomData;
-use std::sync::Arc;
-use std::time::SystemTime;
+// #[macro_use]
+// extern crate prometheus;
 
 use actix_service::{Service, Transform};
 use actix_web::{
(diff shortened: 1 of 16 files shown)


libs/bragi/src/routes/autocomplete.rs
```diff
@@ -125,6 +125,72 @@ impl JsonParams {
     }
 }
 
+/// This function can be used to translate the bragi parameters
+/// into a vector of pairs directly usable in the query part of the URL
+pub fn autocomplete_params_to_query_params(params: Params) -> Vec<(String, String)> {
```
kinnou02: I don't really like this function, it's not really the responsibility of bragi to do this, I'am pretty sure we will forget to update it when we add a parameter. 
I don't see anyway to automate this

riendegris: :+1:
yeah... I don't like this function either. I will revert opening private modules, and completely separate my little CLI tool from the bragi crate.

riendegris:
I've renamed the pull request to reflect the fact that it is no longer about publishing modules that were previously private, eg route/autocomplete.
What remains in the change, and is still valid, are changes related to moving code to rust 2018 standards.

libs/bragi/src/prometheus_middleware.rs
```diff
@@ -3,9 +3,8 @@
 // because  we want to use the name of the endpoint for retrocompatibility
 // (and as a side effect we also added the 'in flight' queries (but for this we could have used the Registry))
 
-use std::marker::PhantomData;
-use std::sync::Arc;
-use std::time::SystemTime;
+// #[macro_use]
```
antoine-de: you can remove those comments

riendegris: :+1:

PR #338, opened 2019-08-08
Modify the filter on shape

riendegris:
This change introduces a modification of the filtering based on shapes:

The filter becomes inactive for stop areas.

### DIFF
--- a/tests/bragi_three_cities_test.rs
+++ b/tests/bragi_three_cities_test.rs
@@ -222,6 +222,9 @@ fn three_cities_shape_test(bragi: &mut BragiHandler) {
     // - stop1 visible       (because within shape)
     // - street2 not visible (because outside of shape)
     // - stop2 visible       (because type=stop area)
+    //
+    // We also want to make sure street2 is visible if we don't use shape filtering
+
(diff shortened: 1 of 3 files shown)


tests/bragi_three_cities_test.rs
```diff
@@ -195,4 +205,42 @@ fn three_cities_shape_test(bragi: &mut BragiHandler) {
...
+
+    let geocodings = bragi.post("/autocomplete?q=Stop In&_all_data=true", shape);
+    assert_eq!(geocodings.len(), 1);
+    assert_eq!(get_values(&geocodings, "label"), vec!["Stop In (Melun)"]);
+
+    let geocodings = bragi.post("/autocomplete?q=Stop Out&_all_data=true", shape);
+    assert_eq!(geocodings.len(), 1);
+    assert_eq!(get_values(&geocodings, "label"), vec!["Stop Out (Melun)"]);
+
+    let geocodings = bragi.post("/autocomplete?q=Four", shape);
```
kinnou02: You might want to check that we found it without filter

riendegris: :+1:

PR #339, opened 2019-08-21
Feature: Add explain to bragi

riendegris:
This change adds a query url parameter to bragi (`debug=true`),
which triggers a so called explain query in elasticsearch, and
forwards the results back as json.

By default, when the parameter is not present, the flag debug is false.

This new feature can be used to analyze the score assigned by
elasticsearch to a result.

This change adds an 'explain' object to a geocodejson feature object,
which is not part of the initial specification.

This change addresses issue #214

### DIFF
--- a/libs/bragi/src/routes/autocomplete.rs
+++ b/libs/bragi/src/routes/autocomplete.rs
@@ -80,6 +80,9 @@ pub struct Params {
     #[serde(default, rename = "poi_type")]
     poi_types: Vec<PoiType>,
     lang: Option<String>,
+    // Forwards a request for explanation to Elastic Search.
+    // This parameter is useful to analyze the order in which search results appear.
+    debug: Option<bool>,
 }
 
 impl Params {
@@ -157,6 +160,7 @@ pub fn call_autocomplete(
         &params.poi_types_as_str(),
         &langs,
         rubber,
+        params.debug.unwrap_or(false),
     );
     res.map(|r| Autocomplete::from_with_lang(r, langs.into_iter().next()))
         .map(Json)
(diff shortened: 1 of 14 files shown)


libs/bragi/src/routes/autocomplete.rs
```diff
@@ -80,6 +80,9 @@ pub struct Params {
     #[serde(default, rename = "poi_type")]
     poi_types: Vec<PoiType>,
     lang: Option<String>,
+    // Forwards a request for explanation to Elastic Search.
+    // This parameter is useful to analyze the order in which search results appear.
+    debug: Option<bool>,
```
kinnou02: in the URL i would like to prefix it with an underscore as it isn't a public parameter

libs/mimir/src/objects.rs
```diff
@@ -243,6 +265,8 @@ pub struct Poi {
     /// Not serialized as is because it is returned in the `Feature` object
     #[serde(default, skip)]
     pub distance: Option<u32>,
+
+    pub explanation: Option<Explanation>,
```
antoine-de: it might be worth it to nest it behind a `debug` object, this way it will be easier to add more fields in it in the future (like the `query`, or some computed values

libs/bragi/src/routes/autocomplete.rs
riendegris: :+1:

libs/mimir/src/objects.rs
riendegris: so you're talking about something like:
```
struct Poi {
  [...]
  pub debug: Option<Debug>
  [...]
}

struct Debug {
  explanation: Option<Explanation>,
  [ moaaah debug stuff later in here... ]
}
```

?

antoine-de: yup :+1:

PR #340, opened 2019-08-21
Add administrative regions for admins

GuillaumeGomez:


libs/mimir/src/objects.rs
```diff
@@ -529,6 +529,7 @@ pub struct Admin {
         default
     )]
     pub boundary: Option<geo::MultiPolygon<f64>>,
+    pub administrative_regions: Vec<Arc<Admin>>,
```
amatissart: I think we need to add a default rule to handle deseriailization of older objects for backward compatibility

src/bin/cosmogony2mimir.rs
```diff
@@ -161,12 +188,26 @@ fn index_cosmogony(args: Args) -> Result<(), Error> {
     let max_weight = utils::ADMIN_MAX_WEIGHT;
 
     info!("importing cosmogony into Mimir");
+    let admins_without_boundaries = read_zones(&args.input)?
```
amatissart: maybe it would be possible to merge this "read" with the first one, to use 2 passes instead of 3 ?

PR #341, opened 2019-08-22
Add administrative regions for admins

GuillaumeGomez:
Reopening of #340.

### DIFF
--- a/src/bin/cosmogony2mimir.rs
+++ b/src/bin/cosmogony2mimir.rs
@@ -178,29 +178,30 @@ fn index_cosmogony(args: Args) -> Result<(), Error> {
     use cosmogony::ZoneType::City;
 
     let mut cosmogony_id_to_osm_id = BTreeMap::new();
-    for z in read_zones(&args.input)? {
+    let max_weight = utils::ADMIN_MAX_WEIGHT;
+    let mut zones = Vec::new();
+    for mut z in read_zones(&args.input)? {
         let insee = match z.zone_type {
             Some(City) => admin::read_insee(&z.tags).map(|s| s.to_owned()),
             _ => None,
         };
         cosmogony_id_to_osm_id.insert(z.id.clone(), (z.osm_id.clone(), insee));
+        z.boundary = None; // to prevent too much memory consumption
+        zones.push(z);
     }
-    let cosmogony_id_to_osm_id = cosmogony_id_to_osm_id;
-    let max_weight = utils::ADMIN_MAX_WEIGHT;
+
+    let admins_without_boundaries = zones.into_iter().map(|z| {
+        let admin = z.into_admin(
+            &cosmogony_id_to_osm_id,
+            &args.langs,
+            args.french_id_retrocompatibility,
+            max_weight,
+            None,
+        );
+        (admin.id.clone(), Arc::new(admin))
+    }).collect::<HashMap<_, _>>();
 
     info!("importing cosmogony into Mimir");
-    let admins_without_boundaries = read_zones(&args.input)?
-        .map(|z| {
-            let admin = z.into_admin(
-                &cosmogony_id_to_osm_id,
-                &args.langs,
-                args.french_id_retrocompatibility,
-                max_weight,
-                None,
-            );
-            (admin.id.clone(), Arc::new(admin))
-        })
-        .collect::<HashMap<_, _>>();
 
     let admins = read_zones(&args.input)?.map(|z| {
         z.into_admin(
--- a/src/bin/cosmogony2mimir.rs
+++ b/src/bin/cosmogony2mimir.rs
@@ -190,16 +190,19 @@ fn index_cosmogony(args: Args) -> Result<(), Error> {
         zones.push(z);
     }
 
-    let admins_without_boundaries = zones.into_iter().map(|z| {
-        let admin = z.into_admin(
-            &cosmogony_id_to_osm_id,
-            &args.langs,
-            args.french_id_retrocompatibility,
-            max_weight,
-            None,
-        );
-        (admin.id.clone(), Arc::new(admin))
-    }).collect::<HashMap<_, _>>();
+    let admins_without_boundaries = zones
+        .into_iter()
+        .map(|z| {
+            let admin = z.into_admin(
+                &cosmogony_id_to_osm_id,
+                &args.langs,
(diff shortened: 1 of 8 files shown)


src/bin/cosmogony2mimir.rs
```diff
@@ -104,7 +108,11 @@ impl IntoAdmin for Zone {
             zip_codes: zip_codes,
             weight: utils::normalize_weight(weight, max_weight),
             bbox: self.bbox,
-            boundary: self.boundary,
+            boundary: if all_admins.is_some() {
```
riendegris: isn't this equivalent to `all_admins.and(self.boundary)`

src/bin/cosmogony2mimir.rs
```diff
@@ -121,7 +129,26 @@ impl IntoAdmin for Zone {
                 .filter(|(k, _)| langs.contains(&k))
                 .collect(),
             distance: None,
+            administrative_regions: Vec::new(),
+        };
+        if let Some(ref admins) = all_admins {
+            let mut a = Vec::new();
```
riendegris: may i suggest being a bit more creative than `a`.... like `parent_ids`,

and maybe add a comment, eg
// get a list of encompassing parent ids, which will be used as the get  administrative_regions

src/bin/cosmogony2mimir.rs
```diff
@@ -150,23 +177,41 @@ fn index_cosmogony(args: Args) -> Result<(), Error> {
     use cosmogony::ZoneType::City;
 
     let mut cosmogony_id_to_osm_id = BTreeMap::new();
-    for z in read_zones(&args.input)? {
+    let max_weight = utils::ADMIN_MAX_WEIGHT;
+    let mut zones = Vec::new();
```
riendegris: personnally i find fold to be more readable:
```
let zones = read_zones()(Vec::new(), |zone| { let insee...; zones.push(zone); zones });

File: src/bin/cosmogony2mimir.rs [hunk 4]
```diff
@@ -150,23 +177,41 @@ fn index_cosmogony(args: Args) -> Result<(), Error> {
...
+                args.french_id_retrocompatibility,
+                max_weight,
+                None,
+            );
+            (admin.id.clone(), Arc::new(admin))
+        })
+        .collect::<HashMap<_, _>>();
 
     info!("importing cosmogony into Mimir");
+
```
riendegris: isn't it the second time you read_zones? you've already created a vector of zones, can you merge the two iterations into one?

GuillaumeGomez: If I could I would. Unfortunately, to avoid too much memory usage, the first time we read, we remove the boundaries and generate osm ids. The second time, we build everything, including parents.

src/bin/cosmogony2mimir.rs
GuillaumeGomez: I don't. Not a big fan of all those helper methods which force readers to look at docs...

src/bin/cosmogony2mimir.rs
GuillaumeGomez: ... Yes it is. :laughing:

src/bin/cosmogony2mimir.rs
riendegris: Ok, I won't argue, and I'll let you be the judge.... I, like you maybe, often have to lookup those `map_or_else`, unwrap_or_do_something`, `and_then_do_this`... These are elements of the language and make for a more idiomatic read. The reason I still look at them.... is because I haven't taken the time to learn them. In other words, the code is of a higher standard, and I need to get up there, not get the code down where I am at. This is not just a rust thing, I went down the same path with C++.

As for folds.... well, it's a bit like a design pattern. It's present in most languages. A fold expresses an aggregation mechanism, and it makes the code clearer, because my code expresses my intention, and not more (like how do we iterate over a vec). It's more of a functional programming style

Again you're the judge

GuillaumeGomez: This is the resulting code:

```rust
let zones = read_zones(&args.input)?.fold(Vec::new(), |zones, zone| {
    let insee = match zone.zone_type {
        Some(City) => admin::read_insee(&z.tags).map(|s| s.to_owned()),
        _ => None,
    };
    cosmogony_id_to_osm_id.insert(zone.id.clone(), (zone.osm_id.clone(), insee));
    zone.boundary = None; // to prevent too much memory consumption
    zones.push(z);
    zones
});
```

I see literally no gain in code readibility or anything... It makes sense when working with values you sum up but in here, we just fulfill a vec... So unless other people agree with you, I'll keep the current code.

src/bin/cosmogony2mimir.rs
```diff
@@ -104,7 +108,7 @@ impl IntoAdmin for Zone {
             zip_codes: zip_codes,
             weight: utils::normalize_weight(weight, max_weight),
             bbox: self.bbox,
-            boundary: self.boundary,
+            boundary: all_admins.and(self.boundary),
```
amatissart: is this check still necessary, now that we are loading the zones without boundary on the first pass ?

GuillaumeGomez: Not anymore, indeed. Good catch!

GuillaumeGomez:
travis failed to start some tests.

src/bin/cosmogony2mimir.rs
TeXitoi:
```rust
les zones = read_zones(&args.input)?
    .map(|mut zone| {
        let insee = match zone.zone_type {
            Some(City) => admin::read_insee(&z.tags).map(|s| s.to_owned()),
            _ => None,
        };
        cosmogony_id_to_osm_id.insert(zone.id.clone(), (zone.osm_id.clone(), insee));
        zone.boundary = None;
        zone
    })
    .collect();

Comment (TeXitoi) [reply src/bin/cosmogony2mimir.rs]: Fold in not so idiomatic in rust. It's convenient when you have the exact function at end:

```rust
let s = v.into_iter().fold(0, i32::max);
```

riendegris: that floats my boat :boat:

GuillaumeGomez: Awesome!

GuillaumeGomez:
Updated.

PR #347, opened 2019-08-28
Doc: Fist load admin

### LINKED ISSUE #346: Compoment loading order and missing index
There is an order on component loading?

osm2mimir miss the ES index. But bano2mimir create its own index.

frodrigo:
Better doc for fix #346


kinnou02:
Thanks for the PR!

PR #350, opened 2019-09-03
Greatly improve memory usage in osm2mimir streets computation

GuillaumeGomez:
Depends on https://github.com/TeXitoi/osmpbfreader-rs/pull/22

When running on central america with osm2mimir, the memory consumption goes from 5464 MB to 4790 MB. Considering this is a **very** small file, I expect the difference to be way bigger for the whole world. Only question: should I provide an option for this or not?

### DIFF
--- a/src/osm_reader/osm_utils.rs
+++ b/src/osm_reader/osm_utils.rs
@@ -33,10 +33,10 @@ use osmpbfreader;
 
 use geo::centroid::Centroid;
 use geo::MultiPolygon;
-use std::collections::BTreeMap;
+use osmpbfreader::StoreObjs;
 
-pub fn get_way_coord(
-    obj_map: &BTreeMap<osmpbfreader::OsmId, osmpbfreader::OsmObj>,
+pub fn get_way_coord<T: StoreObjs>(
+    obj_map: &T,
     way: &osmpbfreader::objects::Way,
 ) -> mimir::Coord {
     /*
@@ -49,8 +49,11 @@ pub fn get_way_coord(
         .iter()
         .skip(nb_nodes / 2)
         .filter_map(|node_id| obj_map.get(&(*node_id).into()))
-        .filter_map(|obj| obj.node())
-        .map(|node| mimir::Coord::new(node.lon(), node.lat()))
+        .filter(|obj| obj.node().is_some())
+        .map(|obj| {
+            let node = obj.node().unwrap();
+            mimir::Coord::new(node.lon(), node.lat())
+        })
         .next()
         .unwrap_or_else(mimir::Coord::default)
 }
--- a/src/osm_reader/osm_utils.rs
+++ b/src/osm_reader/osm_utils.rs
@@ -50,11 +50,7 @@ pub fn get_way_coord<T: StoreObjs + Getter>(
         .iter()
         .skip(nb_nodes / 2)
         .filter_map(|node_id| obj_map.get(&(*node_id).into()))
-        .filter(|obj| obj.node().is_some())
-        .map(|obj| {
-            let node = obj.node().unwrap();
-            mimir::Coord::new(node.lon(), node.lat())
-        })
+        .filter_map(|obj| obj.node().map(|node| mimir::Coord::new(node.lon(), node.lat())))
         .next()
         .unwrap_or_else(mimir::Coord::default)
 }
--- a/src/osm_reader/osm_utils.rs
+++ b/src/osm_reader/osm_utils.rs
@@ -34,8 +34,9 @@ use osmpbfreader;
 use geo::centroid::Centroid;
 use geo::MultiPolygon;
 use osmpbfreader::StoreObjs;
+use super::street::Getter;
 
-pub fn get_way_coord<T: StoreObjs>(
+pub fn get_way_coord<T: StoreObjs + Getter>(
     obj_map: &T,
     way: &osmpbfreader::objects::Way,
 ) -> mimir::Coord {
--- a/src/osm_reader/osm_utils.rs
+++ b/src/osm_reader/osm_utils.rs
@@ -31,10 +31,10 @@
 use mimir;
 use osmpbfreader;
 
+use super::street::Getter;
 use geo::centroid::Centroid;
 use geo::MultiPolygon;
 use osmpbfreader::StoreObjs;
-use super::street::Getter;
 
 pub fn get_way_coord<T: StoreObjs + Getter>(
     obj_map: &T,
@@ -50,7 +50,10 @@ pub fn get_way_coord<T: StoreObjs + Getter>(
         .iter()
         .skip(nb_nodes / 2)
         .filter_map(|node_id| obj_map.get(&(*node_id).into()))
-        .filter_map(|obj| obj.node().map(|node| mimir::Coord::new(node.lon(), node.lat())))
+        .filter_map(|obj| {
+            obj.node()
+                .map(|node| mimir::Coord::new(node.lon(), node.lat()))
+        })
         .next()
         .unwrap_or_else(mimir::Coord::default)
 }
--- a/src/osm_reader/street.rs
+++ b/src/osm_reader/street.rs
@@ -34,7 +34,7 @@ use crate::{labels, utils, Error};
 use failure::ResultExt;
 use osmpbfreader::{OsmId, OsmObj, StoreObjs};
 use rusqlite::{Connection, ToSql, NO_PARAMS};
-use serde_json;
+use bincode;
 use slog_scope::{error, info};
 use std::borrow::Cow;
 use std::collections::{BTreeMap, BTreeSet};
@@ -67,11 +67,11 @@ macro_rules! err_logger {
 macro_rules! get_kind {
     ($obj:expr) => {
         if $obj.is_node() {
-            &"node"
+            &0
         } else if $obj.is_way() {
-            &"way"
+            &1
         } else if $obj.is_relation() {
-            &"relation"
+            &2
         } else {
             panic!("Unknown OSM object kind!")
         }
@@ -102,8 +102,8 @@ impl<'a> DB<'a> {
         conn.execute(
             "CREATE TABLE ids (
                 id   INTEGER NOT NULL,
-                obj  TEXT NOT NULL,
-                kind TEXT NOT NULL,
+                obj  BLOB NOT NULL,
+                kind INTEGER NOT NULL,
                 UNIQUE(id, kind)
              )",
             NO_PARAMS,
@@ -123,11 +123,13 @@ impl<'a> DB<'a> {
             "DB::get_from_id: query_map failed"
         );
         while let Some(row) = err_logger!(iter.next(), "DB::get_from_id: next failed") {
-            let obj: String = err_logger!(row.get(0), "DB::get_from_id: failed to get obj field");
-            return err_logger!(
-                serde_json::from_str(&obj),
-                "DB::get_from_id: conversion from string failed"
+            let obj: Vec<u8> = err_logger!(row.get(0), "DB::get_from_id: failed to get obj field");
+            let obj: OsmObj = err_logger!(
+                bincode::deserialize(&obj),
+                "DB::for_each: serde conversion failed",
+                None
             );
+            return Some(obj);
         }
         None
     }
@@ -140,10 +142,10 @@ impl<'a> DB<'a> {
         );
         let mut rows = err_logger!(stmt.query(NO_PARAMS), "DB::for_each: query_map failed", ());
         while let Some(row) = err_logger!(rows.next(), "DB::for_each: next failed", ()) {
-            let obj: String = err_logger!(row.get(0), "DB::for_each: failed to get obj field", ());
+            let obj: Vec<u8> = err_logger!(row.get(0), "DB::for_each: failed to get obj field", ());
 
             let obj: OsmObj = err_logger!(
-                serde_json::from_str(&obj),
+                bincode::deserialize(&obj),
                 "DB::for_each: serde conversion failed",
                 ()
             );
@@ -155,15 +157,15 @@ impl<'a> DB<'a> {
 impl<'a> StoreObjs for DB<'a> {
     fn insert(&mut self, id: OsmId, obj: OsmObj) {
         let kind = get_kind!(obj);
-        let obj = err_logger!(
-            serde_json::to_string(&obj),
+        let ser_obj = err_logger!(
+            bincode::serialize(&obj),
             "DB::insert: failed to convert to json",
             ()
         );
         err_logger!(
             self.conn.execute(
                 "INSERT OR IGNORE INTO ids(id, obj, kind) VALUES (?1, ?2, ?3)",
-                &[&id.inner_id() as &dyn ToSql, &obj, kind]
+                &[&id.inner_id() as &dyn ToSql, &ser_obj, kind]
             ),
             "DB::insert: insert failed",
             ()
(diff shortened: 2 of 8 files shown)


src/osm_reader/street.rs
```diff
@@ -36,6 +36,96 @@ use slog_scope::info;
...
+        }
+    }
+}
+
+impl StoreObjs for DB {
+    fn insert(&mut self, id: OsmId, obj: OsmObj) {
+        let obj = serde_json::to_string(&obj).expect("failed to convert to json");
+        self.conn.execute(
+            "INSERT OR IGNORE INTO ids(id, obj) VALUES (?1, ?2)",
+            &[&id.as_i64() as &dyn ToSql, &obj],
```
TeXitoi: You can have a way, a node and a relation with the same identifier.

src/osm_reader/street.rs
```diff
@@ -36,6 +36,96 @@ use slog_scope::info;
...
+use rusqlite::{Connection, ToSql, NO_PARAMS};
+use serde_json;
+use std::fs;
+
+// TODO: regarder si diesel serait mieux
+struct DB {
+    conn: Connection,
+}
+
+const DB_FILE_PATH: &str = "./cosmogony_db.db3";
```
TeXitoi: That's dirty, better to give an option, with as value the path to the file. Else, you need the working directory to be writable.

src/osm_reader/street.rs
```diff
@@ -36,6 +36,96 @@ use slog_scope::info;
 use std::collections::{BTreeMap, BTreeSet};
 use std::ops::Deref;
 use std::sync::Arc;
+use osmpbfreader::{StoreObjs, OsmId, OsmObj, OsmObjWrapper, NodeId, WayId, RelationId};
+use rusqlite::{Connection, ToSql, NO_PARAMS};
+use serde_json;
+use std::fs;
+
+// TODO: regarder si diesel serait mieux
```
TeXitoi: french

src/osm_reader/street.rs
```diff
@@ -36,6 +36,96 @@ use slog_scope::info;
...
+            .prepare("SELECT obj FROM ids WHERE id=?1").expect("prepare failed");
+        let mut iter = stmt.query(&[&id.as_i64() as &dyn ToSql]).expect("query_map failed");
+        while let Some(row) = iter.next().expect("next failed") {
+            let obj: String = row.get(0).expect("failed to get obj field");
+            return serde_json::from_str(&obj).expect("conversion from string failed")
+        }
+        None
+    }
+
+    fn iter<F: FnMut(OsmId, OsmObj)>(&self, mut f: F) {
```
TeXitoi: An `iter` function should return an iterator. That's more a `for_each` method.

src/osm_reader/street.rs
GuillaumeGomez: Yes and it's handled as such.

src/osm_reader/street.rs
GuillaumeGomez: Oh true. Forgot about it.

src/osm_reader/street.rs
GuillaumeGomez: Damn, forgot to remove this line...

src/osm_reader/street.rs
TeXitoi: You can implement an `iter` function, it would be much better.

GuillaumeGomez: No actually I can't. And believe me I tried for quite a long time. rusqlite has lifetime issues which forced me to make it work this way. But renaming it to `for_each` seems better.

src/osm_reader/street.rs
GuillaumeGomez: And it's handled as such in the getter.

src/osm_reader/street.rs
```diff
@@ -36,6 +36,96 @@ use slog_scope::info;
...
+        }
+    }
+
+    fn get_from_id(&self, id: &OsmId) -> Option<OsmObj> {
+        let mut stmt = self.conn
+            .prepare("SELECT obj FROM ids WHERE id=?1").expect("prepare failed");
+        let mut iter = stmt.query(&[&id.as_i64() as &dyn ToSql]).expect("query_map failed");
+        while let Some(row) = iter.next().expect("next failed") {
+            let obj: String = row.get(0).expect("failed to get obj field");
+            return serde_json::from_str(&obj).expect("conversion from string failed")
```
TeXitoi: Using json is quite dirty and inefficient. Even if still dirty, bson might be much better in performances. And something using the fact that coords are 2 i32 can improve the performances a lot.

src/osm_reader/street.rs
```diff
@@ -36,6 +36,96 @@ use slog_scope::info;
...
+            let obj: String = row.get(1).expect("failed to get obj field");
+
+            let obj: OsmObj = serde_json::from_str(&obj).expect("serde conversion failed");
+            let id = if obj.is_node() {
+                OsmId::Node(NodeId(id))
+            } else if obj.is_way() {
+                OsmId::Way(WayId(id))
+            } else {
+                OsmId::Relation(RelationId(id))
+            };
```
TeXitoi: That's a reimplementation of the `id` method.

src/osm_reader/street.rs
```diff
@@ -36,6 +36,96 @@ use slog_scope::info;
...
+            f(id, obj);
```
TeXitoi: Why providing this `id` parameter when it's already in `obj`?

src/osm_reader/street.rs
```diff
@@ -36,6 +36,96 @@ use slog_scope::info;
...
+        let conn = Connection::open(&DB_FILE_PATH).expect("failed to open SQLITE connection");
+
+        conn.execute(
+            "CREATE TABLE ids (
+                id   INTEGER PRIMARY KEY,
+                obj  TEXT NOT NULL,
+                UNIQUE(id)
+             )",
+            NO_PARAMS,
+        ).expect("failed to create table");
```
TeXitoi: Should not panic as it can fail in normal circumstances.

src/osm_reader/street.rs
```diff
@@ -36,6 +36,96 @@ use slog_scope::info;
...
+    }
+}
+
+impl StoreObjs for DB {
+    fn insert(&mut self, id: OsmId, obj: OsmObj) {
+        let obj = serde_json::to_string(&obj).expect("failed to convert to json");
+        self.conn.execute(
+            "INSERT OR IGNORE INTO ids(id, obj) VALUES (?1, ?2)",
+            &[&id.as_i64() as &dyn ToSql, &obj],
+        ).expect("failed to insert values");
```
TeXitoi: Can panic in normal circumstances (i.e. disk is full).

src/osm_reader/street.rs
GuillaumeGomez: We're kinda forced to keep it as floating numbers. I don't know about bson format, seems like it's worth taking a look.

src/osm_reader/street.rs
```diff
@@ -36,6 +36,96 @@ use slog_scope::info;
...
+struct DB {
+    conn: Connection,
+}
+
+const DB_FILE_PATH: &str = "./cosmogony_db.db3";
+
+impl DB {
+    fn new() -> DB {
+        let _ = fs::remove_file(DB_FILE_PATH); // we ignore any potential error
+        let conn = Connection::open(&DB_FILE_PATH).expect("failed to open SQLITE connection");
```
TeXitoi: Should not panic as it can fail in normal circumstances.

src/osm_reader/street.rs
GuillaumeGomez: I was looking for a `From` implementation on the `OsmId` type but didn't think about looking at it the other way. Thanks!

src/osm_reader/street.rs
GuillaumeGomez: Tried to keep the same API mostly.

src/osm_reader/osm_utils.rs
```diff
@@ -49,8 +49,11 @@ pub fn get_way_coord(
         .iter()
         .skip(nb_nodes / 2)
         .filter_map(|node_id| obj_map.get(&(*node_id).into()))
-        .filter_map(|obj| obj.node())
-        .map(|node| mimir::Coord::new(node.lon(), node.lat()))
+        .filter(|obj| obj.node().is_some())
+        .map(|obj| {
+            let node = obj.node().unwrap();
+            mimir::Coord::new(node.lon(), node.lat())
+        })
```
TeXitoi: But why? That's just more dirty than before...

src/osm_reader/street.rs
```diff
@@ -78,71 +168,75 @@ pub fn streets(
...
-                    weight: 0.,
-                    zip_codes: utils::get_zip_codes_from_admins(&admins),
-                    administrative_regions: admins,
-                    coord: get_way_coord(&objs_map, way),
-                    approx_coord: Some(coord.into()),
-                    distance: None,
-                    country_codes,
-                    context: None,
+    objs_map.iter(|_, obj| {
+        if let Some(rel) = obj.relation() {
```
TeXitoi: This big if is not really readable, better to early return.

src/osm_reader/street.rs
GuillaumeGomez: Indeed. I should have precised it but this is mostly a POC: the global implementation is "correct" but I didn't clean up at all (sorry about that too...).

src/osm_reader/street.rs
GuillaumeGomez: Indeed!

src/osm_reader/osm_utils.rs
GuillaumeGomez: Lifetime issues again... I didn't make those changes for fun you know. :laughing:

GuillaumeGomez: (rusqlite made my life very miserable)

src/osm_reader/street.rs
TeXitoi: If you have `Node(42)` and `Way(42)`, you will not store the 2, and it will not work as expected.

GuillaumeGomez: Oh I see! Then I need to store the type alongside and make the unique above the both of them and remove the primary key. Excellent catch! Didn't think about it at all...

src/osm_reader/street.rs
TeXitoi: I mean osmpbfreader store 2 i32 for the coord, that helps a lot to compact memory: https://github.com/TeXitoi/osmpbfreader-rs/blob/master/src/objects.rs#L179-L182

GuillaumeGomez: Reminds me of my school years when trying to get a bit better performance. Very nice!

src/osm_reader/osm_utils.rs
TeXitoi:
```rust
.filter_map(|obj| obj.node().map(|node| mimir::Coord::new(node.lon(), node.lat())))
```
maybe?

src/osm_reader/street.rs
TeXitoi: I think I'd remove it as it was like that because of BTreeMap, but that's just my personal opinion, that's up to you.

GuillaumeGomez: No, it doesn't make sense to keep it this way. It'll mostly depends if the others want to be able to choose between db or memory storage.

riendegris:
@GuillaumeGomez, would you mind providing a bit more context with this PR? Maybe a title that says what it's about... It will help people reviewing the PR. Thanks

GuillaumeGomez:
I need to update a few more things before but yes I'll!

src/osm_reader/osm_utils.rs
GuillaumeGomez: Awesome! Didn't think about it at all. Thanks!

src/osm_reader/street.rs
```diff
@@ -59,70 +58,109 @@ impl Getter for BTreeMap<OsmId, OsmObj> {
...
+}
+
+macro_rules! get_kind {
+    ($obj:expr) => {
+        if $obj.is_node() {
+            &"node"
+        } else if $obj.is_way() {
+            &"way"
+        } else if $obj.is_relation() {
+            &"relation"
```
TeXitoi: A match would be more appropriate

TeXitoi: And why a macro when you can do a function?

GuillaumeGomez: A function wouldn't work unless we create a new trait since I use it for both OsmId and OsmObj types. A match didn't seem better to me in since we have to rely on functions.

src/osm_reader/street.rs
GuillaumeGomez: @TeXitoi: I wonder about something with the option: I want to be able to enable/disable the DB usage (and come back to btreemap if wanted). However, I have two options here:
 1. make a `db_path` option which, if not set, means we want to use the btreemap
 2. provide both `db_path` and `enable_db` options

What do you think is better?

TeXitoi: I'd prefer 1, but that's my personal opinion, and I'm no more involved in this project officially.

src/osm_reader/street.rs
TeXitoi: OK

src/osm_reader/street.rs
GuillaumeGomez: Even so, your opinion is highly appreciated so let's go for 1!

src/osm_reader/street.rs
```diff
@@ -32,11 +32,172 @@ use super::OsmPbfReader;
...
+            NO_PARAMS,
+        )
+        .map_err(|e| format!("failed to create table: {}", e))?;
+        Ok(DB { conn, db_file })
+    }
+
+    fn get_from_id(&self, id: &OsmId) -> Option<OsmObj> {
+        let mut stmt = err_logger!(
+            self.conn
+                .prepare("SELECT obj FROM ids WHERE id=?1 AND kind=?2"),
```
kinnou02: Minor: I think the prepare should be done only one time and the query reused, [`prepare_cached`](https://docs.rs/rusqlite/0.20.0/rusqlite/struct.Connection.html#method.prepare_cached) can help

GuillaumeGomez:
@kinnou02 Me too but considering it highly depends on the number of streets, I can't say for sure until we try it on the world.

src/osm_reader/street.rs
GuillaumeGomez: I missed it, thanks for the tip!

GuillaumeGomez: And of course, it completely breaks the whole thing because of lifetime issues. I'll send a PR to rusqlite I think...

kinnou02: We can merge the PR as is, and change this latter if needed.

GuillaumeGomez: I have other changes to merge such as using BSON instead of JSON (for example).

GuillaumeGomez:
Just thought about something: instead of storing a string for the object kind, I should store a small integer. Silly me for not thinking about it earlier...

GuillaumeGomez:
Ok, should be ready now.

GuillaumeGomez:
@amatissart Do you have the memory consumption graph on the world import btw?

amatissart:
yep, `osm2mimir` peaked at 31.6 GB during the planet import (compared to more than 105 GB initially)

kinnou02:
Nice!
How much time did it take?

amatissart:
> Nice!
> How much time did it take?

11h50min, compared to ~ 6 hours initially (osm2mimir `--import-way` only)

PR #354, opened 2019-09-05
Add cache header to bragi http response

kinnou02:
This PR make bragi output a cache header for every requests on
/features, /autocomplete, /reverse.
By default cache is enabled and set to 1H.
This header will be used by http clients to cache results, it will also
be used by caching proxy like varnish.

### DIFF
--- a/libs/bragi/src/lib.rs
+++ b/libs/bragi/src/lib.rs
@@ -91,6 +91,16 @@ pub struct Args {
         env = "BRAGI_MAX_ES_FEATURES_TIMEOUT"
     )]
     pub max_es_features_timeout: Option<u64>,
+
+    /// Cache duration for http response served by bragi
+    /// This only set the Cache-control Header, it doesn't enable cache on bragi side
+    /// The duration is in seconds
+    #[structopt(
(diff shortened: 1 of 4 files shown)


libs/bragi/src/lib.rs
```diff
@@ -91,6 +91,15 @@ pub struct Args {
         env = "BRAGI_MAX_ES_FEATURES_TIMEOUT"
     )]
     pub max_es_features_timeout: Option<u64>,
+
+    /// Cache duration for http response served by bragi
+    /// This only set the Cache-control Header, it doesn't enable cahe on bragi side
```
riendegris: s/cahe/cache/

libs/bragi/src/lib.rs
```diff
@@ -91,6 +91,15 @@ pub struct Args {
...
+    #[structopt(
+        long = "http-cache-duration",
+        env = "BRAGI_HTTP_CACHE_DURATION",
+        default_value = "3600"
+    )]
+    pub http_cache_duration: u32,
```
riendegris: add comment for unit (seconds)

PR #359, opened 2019-10-01
Add support for dynamic weight

riendegris:
Weight is now a function of the place type.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ slog-envlogger = "2.1"
 slog-async = "2.3"
 structopt = "0.2"
 csv = "1"
-rs-es = { git = "https://github.com/canaltp/rs-es", features = ["geo"]}
+rs-es = { path = "../rs-es", features = ["geo"]}
 regex = "1"
 osmpbfreader = "0.13.2"
 chrono = "0.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ slog-envlogger = "2.1"
 slog-async = "2.3"
 structopt = "0.2"
 csv = "1"
-rs-es = { path = "../rs-es", features = ["geo"]}
+rs-es = { git = "https://github.com/canaltp/rs-es", features = ["geo"]}
 regex = "1"
 osmpbfreader = "0.13.2"
 chrono = "0.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ slog-envlogger = "2.1"
 slog-async = "2.3"
 structopt = "0.2"
 csv = "1"
-rs-es = { version = "^0.12.0", features = ["geo"]}
+rs-es = { git = "https://github.com/canaltp/rs-es", features = ["geo"]}
 regex = "1"
 osmpbfreader = "0.13.2"
 chrono = "0.4"
--- a/libs/bragi/src/query.rs
+++ b/libs/bragi/src/query.rs
@@ -151,7 +151,7 @@ fn build_with_weight<A: Into<Option<f64>>>(factor: A) -> Query {
 
     Query::build_function_score()
         .with_functions(vec![
-            weighted(Stop::doc_type(), 10.0),
+            weighted(Stop::doc_type(), 100.0),
             weighted(Addr::doc_type(), 1.0),
             weighted(Admin::doc_type(), 1.0),
             weighted(Poi::doc_type(), 1.0),
--- a/libs/bragi/src/query.rs
+++ b/libs/bragi/src/query.rs
@@ -130,8 +130,13 @@ fn build_proximity_with_boost(coord: &Coord, weight: f64) -> Query {
                 .with_decay(0.4f64)
                 .build("coord")
                 .build_exp(),
+                None,
             ),
-            FilteredFunction::build_filtered_function(None, Function::build_weight(weight).build()),
+            FilteredFunction::build_filtered_function(
+                None,
+                Function::build_weight(weight).build(),
+                None,
+            ),
         ])
         .with_boost_mode(BoostMode::Replace)
         .build()
@@ -146,6 +151,7 @@ fn build_with_weight<A: Into<Option<f64>>>(factor: A) -> Query {
                 .with_factor(factor.unwrap_or(0.15) * weight)
                 .with_missing(0.0)
                 .build(),
+            None,
         )
     };
 
@@ -268,10 +274,12 @@ fn build_query<'a>(
                             .with_modifier(Modifier::Log1p)
                             .with_missing(0.)
                             .build(),
+                        None,
                     ),
                     FilteredFunction::build_filtered_function(
                         None,
                         Function::build_weight(admin_weight).build(),
+                        None,
                     ),
                 ])
                 .with_boost_mode(BoostMode::Replace)
--- a/libs/bragi/src/query.rs
+++ b/libs/bragi/src/query.rs
@@ -144,35 +144,35 @@ fn build_with_weight<A: Into<Option<f64>>>(factor: A) -> Query {
             FilteredFunction::build_filtered_function(
                 Query::build_term("_type", Stop::doc_type()).build(),
                 Function::build_field_value_factor("weight")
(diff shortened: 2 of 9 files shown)


kinnou02:
we use a local fork of rs-es while waiting for https://github.com/benashford/rs-es/pull/142

riendegris:
Note that the coefficient I have assigned are identical accross all place types (1.0), except for stop areas (10.0), so __theoretically__ it should only have an impact when stop area are used...

Cargo.toml
```diff
@@ -16,7 +16,7 @@ slog-envlogger = "2.1"
 slog-async = "2.3"
 structopt = "0.2"
 csv = "1"
-rs-es = { version = "^0.12.0", features = ["geo"]}
+rs-es = { git = "https://github.com/canaltp/rs-es", features = ["geo"]}
```
GuillaumeGomez: Is it because the rs-es didn't release the version yet?

libs/bragi/src/query.rs
```diff
@@ -137,12 +140,43 @@ fn build_proximity_with_boost(coord: &Coord, weight: f64) -> Query {
...
-            Function::build_field_value_factor("weight")
-                .with_factor(factor.unwrap_or_else(|| 0.15))
-                .with_missing(0.)
-                .build(),
-        )
+        .with_functions(vec![
+            FilteredFunction::build_filtered_function(
+                Query::build_term("_type", Stop::doc_type()).build(),
+                Function::build_field_value_factor("weight")
+                    .with_factor(factor.unwrap_or_else(|| 0.15) * 10.0)
```
GuillaumeGomez: Why multiplying by 10 in this case?

libs/bragi/src/query.rs
```diff
@@ -137,12 +140,43 @@ fn build_proximity_with_boost(coord: &Coord, weight: f64) -> Query {
...
+                    .with_missing(0.0)
+                    .build(),
+            ),
+            FilteredFunction::build_filtered_function(
+                Query::build_term("_type", Addr::doc_type()).build(),
+                Function::build_field_value_factor("weight")
+                    .with_factor(factor.unwrap_or_else(|| 0.15) * 1.0)
```
GuillaumeGomez: Why multiplying by 1?

libs/bragi/src/query.rs
```diff
@@ -137,12 +140,43 @@ fn build_proximity_with_boost(coord: &Coord, weight: f64) -> Query {
...
+                    .with_missing(0.0)
+                    .build(),
+            ),
+            FilteredFunction::build_filtered_function(
+                Query::build_term("_type", Admin::doc_type()).build(),
+                Function::build_field_value_factor("weight")
+                    .with_factor(factor.unwrap_or_else(|| 0.15) * 1.0)
```
GuillaumeGomez: Why multiplying by 1?

libs/bragi/src/query.rs
```diff
@@ -137,12 +140,43 @@ fn build_proximity_with_boost(coord: &Coord, weight: f64) -> Query {
...
+                    .with_missing(0.0)
+                    .build(),
+            ),
+            FilteredFunction::build_filtered_function(
+                Query::build_term("_type", Poi::doc_type()).build(),
+                Function::build_field_value_factor("weight")
+                    .with_factor(factor.unwrap_or_else(|| 0.15) * 1.0)
```
GuillaumeGomez: Why multiplying by 1?

libs/bragi/src/query.rs
```diff
@@ -137,12 +140,43 @@ fn build_proximity_with_boost(coord: &Coord, weight: f64) -> Query {
...
+                    .with_missing(0.0)
+                    .build(),
+            ),
+            FilteredFunction::build_filtered_function(
+                Query::build_term("_type", Street::doc_type()).build(),
+                Function::build_field_value_factor("weight")
+                    .with_factor(factor.unwrap_or_else(|| 0.15) * 1.0)
```
GuillaumeGomez: Why multiplying by 1?

Cargo.toml
riendegris: I made a pull request a couple days ago, but it hasn't been commented on nor has it been merged. So we temporarily use the canaltp version which is based on the latest, but has the pull request for 'filtered function'

libs/bragi/src/query.rs
riendegris: because we don't want to change the result for others place types.

this is, however, a relevant question, why 1.0 and not 1.1? This will be discussed very soon (TBC)

libs/bragi/src/query.rs
TeXitoi:
```suggestion
                    .with_factor(factor.unwrap_or(|| 0.15) * 10.0)
```

libs/bragi/src/query.rs
TeXitoi:
```suggestion
                    .with_factor(factor.unwrap_or(0.15))
```

riendegris: :+1:

A macro would go a long way to make this a one liner...

TeXitoi: no macro when you can use a function

GuillaumeGomez: > no macro when you can use a function

I strongly approve this comment!

libs/bragi/src/query.rs
```diff
@@ -137,12 +140,43 @@ fn build_proximity_with_boost(coord: &Coord, weight: f64) -> Query {
...
-                .build(),
-        )
+        .with_functions(vec![
+            FilteredFunction::build_filtered_function(
+                Query::build_term("_type", Stop::doc_type()).build(),
+                Function::build_field_value_factor("weight")
+                    .with_factor(factor.unwrap_or(0.15) * 10.0)
+                    .with_missing(0.0)
+                    .build(),
+            ),
```
TeXitoi:
```rust
let weighted = |doc_type, weight| FilteredFunction::build_filtered_function(
    Query::build_term("_type", doc_type).build(),
    Function::build_field_value_factor("weight")
        .with_factor(factor.unwrap_or(0.15) * weight)
        .with_missing(0.0)
        .build(),
);
Query::build_function_score()
    .with_functions(vec![
        weighted(Stop::doc_type(), 10.),
        weighted(Addr::doc_type(), 1.),
        // ...
    )
```

riendegris: That's a good suggestion. I have a question though, out of curiosity, why do you skip the trailing .0 on a float?

TeXitoi: I don't skip the `.0`, I only put a `.` at the end because that's enough to have a float litteral.

riendegris:
There is another suggestion, please let me know what you think....

The `build_with_weight` function will eventually translate to an ES JSON query that looks like

`
[ 
  {
    "filter": { "match": { "term": "_type", "value": "stop" } },
    "field_value_factor": {
       "field": "weight",
       "factor": 1.5,
       "missing": 0.0
    }
  },
  {
    "filter": { "match": { "term": "_type", "value": "admin" } },
    "field_value_factor": {
       "field": "weight",
       "factor": 0.15,
       "missing": 0.0
    }
  },
...
]
`

So we change the factor inside the `field_value_factor` function. The alternative that I haven't implemented would use weight outside the function_score function:

`
{
  filter: ....
  field_value_score...
  weight: 10
},
{
  filter: ....
  field_value_score...
  weight: 1
}
...
`

riendegris:
graph

libs/bragi/src/query.rs
```diff
@@ -120,29 +120,44 @@ fn build_coverage_condition(pt_datasets: &[&str]) -> Query {
...
+    };
+
+    Query::build_function_score()
+        .with_functions(vec![
+            weighted(Stop::doc_type(), 10.0),
+            weighted(Addr::doc_type(), 1.0),
+            weighted(Admin::doc_type(), 1.0),
+            weighted(Poi::doc_type(), 1.0),
+            weighted(Street::doc_type(), 1.0),
+            weighted(Stop::doc_type(), 1.0),
```
riendegris: Working twice on stops.... :-1:

PR #362, opened 2019-10-08
Stop weight use admin

riendegris:
This PR contains two changes:

- the stop weight is now a function of the admin weight

- the function score uses a weight

### DIFF
--- a/src/stops.rs
+++ b/src/stops.rs
@@ -42,7 +42,7 @@ use std::sync::Arc;
 
 const GLOBAL_STOP_INDEX_NAME: &'static str = "munin_global_stops";
 
-pub fn set_weights<'a, It>(stops: It, nb_stop_points: &HashMap<String, u32>)
+pub fn initialize_weights<'a, It>(stops: It, nb_stop_points: &HashMap<String, u32>)
 where
     It: Iterator<Item = &'a mut mimir::Stop>,
 {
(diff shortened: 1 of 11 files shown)


kinnou02:
Results seems to be the same than https://github.com/CanalTP/navitia/pull/2989/commits/5b818239c458a5e510dbdeaaea60ac139332377a I have launched another geocodertester on `master` to be sure.

src/stops.rs
```diff
@@ -70,6 +70,20 @@ pub fn import_stops(
...
+            .map(|adm| adm.weight)
+            .next()
+            .unwrap_or(0.0);
+        // FIXME: 1024, automagic!
+        // It's a factor used to bring the stop weight and the admin weight in the same order of
+        // magnitude...
+        // We then use a log to compress the distance between low admin weight and high ones.
+        admin_weight = admin_weight * 1024.0 + 1.0;
+        admin_weight = admin_weight.log10();
+        stop.weight = (stop.weight + admin_weight) / 2.0;
```
kinnou02: We will have to works on this weight, in most case it almost doesn't play a part in the query, this was quite visible on stops as a lot of stops have almost the same name, and we want to provide the most important ones first.
Weight a mostly between 0 and 1, but stops have their weight with one decimal while others objects are near 10e-6

@riendegris built some nice visualization of the scoring of a response that will help understand what happens here.

PR #367, opened 2019-11-06
reduce memory usage of cosmogony2mimir

kinnou02:
Europe import now take 2go of RAM in place of 20go.
The import took 11 minutes for 411408 admins, it was taking 17 minutes before but I was swapping

### DIFF
--- a/src/bin/cosmogony2mimir.rs
+++ b/src/bin/cosmogony2mimir.rs
@@ -171,21 +171,19 @@ fn index_cosmogony(args: Args) -> Result<(), Error> {
 
     let mut cosmogony_id_to_osm_id = BTreeMap::new();
     let max_weight = utils::ADMIN_MAX_WEIGHT;
-    let zones = read_zones(&args.input)?
-        .map(|mut zone| {
-            let insee = match zone.zone_type {
-                Some(City) => admin::read_insee(&zone.tags).map(|s| s.to_owned()),
-                _ => None,
-            };
-            cosmogony_id_to_osm_id.insert(zone.id.clone(), (zone.osm_id.clone(), insee));
-            zone.boundary = None; // to prevent too much memory consumption
-            zone
-        })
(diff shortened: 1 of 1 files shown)


TeXitoi:
And timing ?

src/bin/cosmogony2mimir.rs
```diff
@@ -171,21 +171,19 @@ fn index_cosmogony(args: Args) -> Result<(), Error> {
...
+            _ => None,
+        };
+        cosmogony_id_to_osm_id.insert(z.id.clone(), (z.osm_id.clone(), insee));
+    }
+    let cosmogony_id_to_osm_id = cosmogony_id_to_osm_id;
 
-    let admins_without_boundaries = zones
-        .into_iter()
-        .map(|z| {
+    info!("building admins");
```
riendegris: nitpick: is this where you build admins, or is it below? / Are these messages info! or debug!

kinnou02: I updated the message but kept them as `info` because I want to see them to know what is happening

kinnou02:
> And timing ?

Updated!

amatissart:
Nice!

Have you investigated to understand why exactly the `Vec<Zone>` (without boundaries) takes so much memory. It seems likely that this is due to `international_labels`.
By default (if  `--filter-langs` is not used), cosmogony will build them for each zone in every language that is present at least once in the hierarchy.

kinnou02:
from the log I got it seems that somethings prevent the indexing to be streamed...
```
Nov 06 10:49:52.234 INFO importing cosmogony into Mimir
Nov 06 10:57:26.398 INFO elastic search host http://localhost:9200
Nov 06 10:57:28.982 INFO creating index munin_admin_france_20191106_095728_770481713
```
the index should be created almost immediately, but if I remember well most memory was allocated before these logs

the `international_labels`  might explain it, I will check that

PR #369, opened 2019-12-17
OSM2Mimir: Exclude certain kind of highway

riendegris:
'bus_platform', for example, should not be indexed.


SGrenet:
some lacks : 
highway: cycleway (some are named)

For public transport purpose (because they're not pedestrian): 
highway:trunk and highway:trunk_link
but these ways could be usefull for a less specific autocomplete

Is reverse impacted by this feature or only autocomplete ?

PR #370, opened 2020-01-08
Typo fix

GuillaumeGomez:


src/bin/openaddresses2mimir.rs
```diff
@@ -45,7 +45,7 @@ lazy_static! {
 
 #[derive(Serialize, Deserialize)]
 #[serde(rename_all = "SCREAMING_SNAKE_CASE")]
-pub struct OpenAddresse {
+pub struct OpenAddresses {
```
amatissart: Wouldn't it make more sense to call this struct `OpenAddress`, as it represents a single address from the dataset ?

GuillaumeGomez: Let's go for it then!

GuillaumeGomez:
Updated.

PR #373, opened 2020-02-03
WIP Replace gst with rstar; Update dependencies.

riendegris:
gst is unmaintained (4 years) and needs to be replaced.
rstar is the obvious contender for replacing gst.

By making use of geo algorithm, this change simplifies the code
for the AdminGeoFinder.

This proposed change has a breaking consequence however. We can no
longer insert admins with boundaries from 'osm2mimir'. We can still
do with 'cosmogony and cosmogony2mimir'. Here is why:

In the AdminGeoFinder, we strip the boundary from the Admin, to save
memory. In 'osm2mimir', we store the admins found in the pbf file in
the AdminGeoFinder, and then iterate over them to index them in
Elasticsearch. With cosmogony2mimir, we simply iterate over all the
admins found in the cosmogony file, and index them (we don't use the
AdminGeoFinder).

### DIFF
--- a/src/admin_geofinder.rs
+++ b/src/admin_geofinder.rs
@@ -27,154 +27,95 @@
 // IRC #navitia on freenode
 // https://groups.google.com/d/forum/navitia
 // www.navitia.io
-use gst::rtree::{RTree, Rect};
+
+use geo::algorithm::bounding_rect::BoundingRect;
+use geo::algorithm::euclidean_distance::EuclideanDistance;
+use geo_types::{MultiPolygon, Point};
 use mimir::Admin;
-use std::collections::{BTreeMap, BTreeSet};
+use rstar::{PointDistance, RTree, RTreeObject, AABB};
+// use std::collections::BTreeMap;
 use std::iter::FromIterator;
 use std::sync::Arc;
 
-/// We want to strip the admin's boundary for the objects referencing it (for performance purpose)
-/// thus in the `AdminGeoFinder` we store an Admin without the boundary (the option is emptied)
-/// and we store the boundary aside
-struct BoundaryAndAdmin(Option<geo_types::MultiPolygon<f64>>, Arc<Admin>);
+pub struct BoundedId {
+    pub boundary: MultiPolygon<f64>,
+    pub admin: Arc<Admin>,
+}
 
-impl BoundaryAndAdmin {
-    fn new(mut admin: Admin) -> BoundaryAndAdmin {
-        let b = std::mem::replace(&mut admin.boundary, None);
-        let minimal_admin = Arc::new(admin);
-        BoundaryAndAdmin(b, minimal_admin)
+impl RTreeObject for BoundedId {
+    type Envelope = AABB<Point<f64>>;
+
+    fn envelope(&self) -> Self::Envelope {
+        let bb = self.boundary.bounding_rect().unwrap();
+        AABB::from_corners(
+            Point::new(bb.min.x, bb.min.y),
+            Point::new(bb.max.x, bb.max.y),
+        )
+    }
+}
+
+impl PointDistance for BoundedId {
+    // This function computes the square of the distance from an Admin to a point.
+    // We compute the distance from the boundary to a point.
+    fn distance_2(&self, point: &Point<f64>) -> f64 {
+        let d = self.boundary.euclidean_distance(point);
+        d * d
     }
 }
 
 pub struct AdminGeoFinder {
-    admins: RTree<BoundaryAndAdmin>,
-    admin_by_id: BTreeMap<String, Arc<Admin>>,
+    rtree: RTree<BoundedId>,
 }
 
 impl AdminGeoFinder {
     pub fn insert(&mut self, admin: Admin) {
-        use ordered_float::OrderedFloat;
-        fn min(a: OrderedFloat<f32>, b: f64) -> f32 {
-            a.0.min(down(b as f32))
+        let mut admin = admin;
+        let boundary = std::mem::replace(&mut admin.boundary, None);
+        if let Some(boundary) = boundary {
+            let bounded_id = BoundedId {
+                boundary,
+                admin: Arc::new(admin),
+            };
+            self.rtree.insert(bounded_id);
         }
-        fn max(a: OrderedFloat<f32>, b: f64) -> f32 {
-            a.0.max(up(b as f32))
-        }
-
-        let rect = {
-            let mut coords = match admin.boundary {
-                Some(ref b) => b.0.iter().flat_map(|poly| (poly.exterior()).0.iter()),
-                None => return,
-            };
-            let first_coord = match coords.next() {
-                Some(c) => c,
-                None => return,
-            };
-            let first_rect: Rect = {
-                let (x, y) = (first_coord.x as f32, first_coord.y as f32);
-                Rect::from_float(down(x), up(x), down(y), up(y))
-            };
-            coords.fold(first_rect, |accu, p| {
-                Rect::from_float(
-                    min(accu.xmin, p.x),
-                    max(accu.xmax, p.x),
-                    min(accu.ymin, p.y),
-                    max(accu.ymax, p.y),
-                )
-            })
-        };
-        let bound_admin = BoundaryAndAdmin::new(admin);
-        self.admin_by_id
-            .insert(bound_admin.1.id.clone(), bound_admin.1.clone());
-        self.admins.insert(rect, bound_admin);
     }
 
-    /// Get all Admins overlapping the coordinate
+    // Get all Admins overlapping the given coordinates.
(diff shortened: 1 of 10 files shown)


riendegris:
We could, if we wanted, add a method to the AdminGeoFinder which would retrieve the admins with their boundary. this method, eg 'admins_with_boundary', would iterate over all the objects stored in the RTree (which are essentially pairs (boundary, Arc<Admin>), and create a new Admin from that pair, and return that.... I'm not sure its necessary though....

riendegris:
This PR is a WIP, and needs more documentation, and I have to remove a lot of commented code...

riendegris:
Le test unitaire qui plante est un test qui utilise justement l'import d'admins depuis osm2mimir.

src/admin_geofinder.rs
```diff
@@ -27,154 +27,95 @@
...
-        let b = std::mem::replace(&mut admin.boundary, None);
-        let minimal_admin = Arc::new(admin);
-        BoundaryAndAdmin(b, minimal_admin)
+impl PointDistance for BoundedId {
+    // This function computes the square of the distance from an Admin to a point.
+    // We compute the distance from the boundary to a point.
+    fn distance_2(&self, point: &Point<f64>) -> f64 {
+        let d = self.boundary.euclidean_distance(point);
+        d * d
     }
```
amatissart: It seems that this method is used internally by rstar to check if `point` contains the boundary.  
Overriding the [`contains_point`](https://docs.rs/rstar/0.7.1/rstar/trait.PointDistance.html#method.contains_point) method to use `boundary.contains()` directly might be more performant ?

src/admin_geofinder.rs
```diff
@@ -27,154 +27,95 @@
...
-                let mut admin_parent_id = admin.parent_id.clone();
-                while let Some(id) = admin_parent_id {
-                    let admin_parent = self.admin_by_id.get(&id);
-                    if let Some(zt) = admin_parent.as_ref().and_then(|a| a.zone_type) {
-                        added_zone_types.insert(zt.clone());
-                    }
-                    if !tested_hierarchy.insert(id) {
-                        break; // stop the exploration of the hierarchy since we have already added this one
-                    }
-                    admin_parent_id = admin_parent.and_then(|a| a.parent_id.clone());
```
amatissart: This was an important optimization (implemented in #240) to avoid calling `contains` over too many zones, especially with complex country geometries. 
For example, after a city has been found, we should reuse its hierarchy, without any other check about parent geometries.

Do you think the migration to rstar makes this optimization obsolete ?

src/admin_geofinder.rs
riendegris: If I remove this trait implementation, I have

```
.locate_all_at_point(&point)`
   |              ^^^^^^^^^^^^^^^^^^^ method not found in `rstar::rtree::RTree<admin_geofinder::BoundedId>`
   |
   = note: the method `locate_all_at_point` exists but the following trait bounds were not satisfied:
           `admin_geofinder::BoundedId : rstar::object::PointDistance`
```

In other words, I am relying on RStar API to find all the objects at a given location, which requires this trait. I rely on RStar to produce an optimal code.

You seem to be hinting there might be a better solution, but I can't understand your proposition.

riendegris:
Here is something I don't understand, as far as optimization / architecture goes (maybe kinnou02 can help), but we seem to be doing the following:

(1) take osm, use cosmogony (and build an rtree probably), then cosmogony2mimir and store the result in ES.
(2) take osm, use osm2mimir, which retrieves all the admins, builds a new RTree, and answer questions like `get(coord)->Admins`, and store the result in ES
(3) rinse and repeat with bano2mimir, ntfs2mimir, poi2mimir, ...

Why isn't AdminGeoFinder like a daemon? That way we don't need to build the RTree and tear it down for each executable

src/admin_geofinder.rs
amatissart: `PointDistance` trait allows us to provide a custom implementation for `contains_point`:

```diff
 impl PointDistance for BoundedId {
     // This function computes the square of the distance from an Admin to a point.
     // We compute the distance from the boundary to a point.
     fn distance_2(&self, point: &Point<f64>) -> f64 {
         let d = self.boundary.euclidean_distance(point);
         d * d
     }
+
+    fn contains_point(&self, point: &Point<f64>) -> bool {
+        self.boundary.contains(point)
+    }
 }
```

It would be closer to the existing implementation, and possibly less expensive than `euclidean_distance`.

riendegris: I see, you are following the documentation's recommendation:

> By default, any point returning a distance_2 less than or equal to zero is considered to be contained within self. Changing this default behavior is advised if calculating the squared distance is more computational expensive a point containment check.

src/admin_geofinder.rs
```diff
@@ -27,154 +27,95 @@
...
-                std::f32::NEG_INFINITY,
-                std::f32::INFINITY,
-                std::f32::NEG_INFINITY,
-                std::f32::INFINITY,
-            ))
-            .into_iter()
-            .map(|(_, a)| a.1.clone())
+    pub fn admins(&self) -> impl Iterator<Item = Arc<Admin>> + '_ {
+        self.rtree.iter().map(|bounded_id| bounded_id.admin.clone())
     }
```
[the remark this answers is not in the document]
riendegris: Because it seems like an expensive operation. The RTree stores Arc<Admin>, so once in there, we can't modify them anymore, and they're boundary less. If we want to have a list of all admins with their boundary, we need to clone them, and add their boundary..... clone clone...

This method is used by osm2mimir --import-admins, so I guess its a one time cost.... I'll leave with it ;-)

src/admin_geofinder.rs
riendegris: An optimization that makes a process run in 7h vs 17h is certainly not to be discarded.

So I want to spend more time looking at it...

When I run this get method on a regional area (ile-de-france), the length of the vector of admins returned by the RTree is usually 3 or 4. I expect the optimization to be negligible (if not counter productive) in that case. But I will run precise measurements.

=> TODO: Decide what dataset represents a good baseline for those tests? France, Europe?
=> TODO: Run two measurements: before change (gst w/ optim) and after change (rstar wo/ optim)

Now that we're talking about optimization.... I am wondering why we have to be build this cache (added_zone_type, tested_hierarchy) _every_time_we_call_get_ ?? How come the cache is not part of the AdminGeoFinder itself?

riendegris:
With this latest patch, the performance is identical to the previous version which used the `gst` crate.
The test I ran was:

1. Setup: insert cosmogony in ES for France
2. Measurement: insert BANO for department 93.

Result: identical result (444s)

src/admin_geofinder.rs
```diff
@@ -27,84 +27,131 @@
...
-        }
-
-        let rect = {
-            let mut coords = match admin.boundary {
-                Some(ref b) => b.0.iter().flat_map(|poly| (poly.exterior()).0.iter()),
-                None => return,
+        let mut admin = admin;
+        let boundary = std::mem::replace(&mut admin.boundary, None);
+        if let Some(boundary) = boundary {
+            let bb = boundary.bounding_rect().unwrap();
```
amatissart: If there is really no reason for `bounding_rect` to return None here, we should use `expect`.  
But I would simply ignore (possibly with a warning message) any zone if its bounding box is not defined.

src/admin_geofinder.rs
```diff
@@ -310,20 +337,21 @@ mod tests {
...
+    // fn test_two_no_zone_type() {
+    //     // a point can be associated to only 1 admin type
+    //     // but a point can be associated to multiple admin without zone_type
+    //     // (for retrocompatibility of the data imported without cosmogony)
+    //     let mut finder = AdminGeoFinder::default();
+    //     finder.insert(make_admin(40., None));
+    //     finder.insert(make_admin(43., None));
+    //     let admins = finder.get(&p(46., 46.).0);
+    //     assert_eq!(admins.len(), 2);
+    // }
```
amatissart: Is this test broken ?

src/admin_geofinder.rs
```diff
@@ -191,35 +218,35 @@ impl FromIterator<Admin> for AdminGeoFinder {
...
+            assert!(
+                down(small_f) as f64 <= f,
+                format!("{} <= {}", down(small_f) as f64, f)
+            );
+            assert!(
+                f <= up(small_f) as f64,
+                format!("{} <= {}", f, up(small_f) as f64)
+            );
+        }
+    }
```
amatissart: is that not obsolete ?

src/admin_geofinder.rs
riendegris: I will add a warning message.

src/admin_geofinder.rs
riendegris: no, I will remove the comments

src/admin_geofinder.rs
riendegris: yes it is, good catch

PR #374, opened 2020-02-06
Mininale Supported Rust Toolchain version 1.39.0

riendegris:
With the upcoming changes, the minimum version of the rust toolchain that compiles the project has changed from 1.36 -> 1.39.


amatissart:
No objection, but I think we should update the Rust version used in Dockerfiles at the same time.

PR #375, opened 2020-02-12
add a gzip option for CSV addresses importers

remi-dupre:
Add a `gzip` parameter for BANO and OpenAddresses importers, this allows to import gziped CSV files. For example:

```bash
cargo run --release --bin openaddresses2mimir -- --gzip --input=luxembourg.csv.gz --connection-string=$ES_ADDRESS
```

Our use-case is that we intend to import a big amount of addresses that were obtained by merging several sources and that we output using OpenAddresses format.

In general it may still be a nice feature, at least for BANO importer as compressed CSV files are distributed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,7 @@ address-formatter = "^0.2.1"
 navitia-poi-model = "0.2.1"
 walkdir = "2"
 rusqlite = "0.20"
+libflate = "0.1"
 
 mimir = { path = "libs/mimir" }
 bragi = { path = "libs/bragi" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ address-formatter = "^0.2.1"
 navitia-poi-model = "0.2.1"
 walkdir = "2"
 rusqlite = "0.20"
-libflate = "0.1"
+flate2 = "1.0"
 
 mimir = { path = "libs/mimir" }
 bragi = { path = "libs/bragi" }
--- a/src/addr_reader.rs
+++ b/src/addr_reader.rs
@@ -1,7 +1,7 @@
 use crate::Error;
 use csv;
 use failure::ResultExt;
-use libflate::gzip;
+use flate2::read::GzDecoder;
 use mimir::rubber::{IndexSettings, IndexVisibility, Rubber};
 use mimir::Addr;
 use par_map::ParMap;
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -42,6 +42,7 @@ address-formatter = "^0.2.1"
 navitia-poi-model = "0.2.1"
 walkdir = "2"
 rusqlite = "0.20"
+libflate = "0.1"
```
amatissart: `flate2`, already used by cosmogony and osmpbfreader could be a better choice

remi-dupre: Nice catch! I fixed this here: 2579f1aa2a035e3aa99fddae2428c1511f45d305.

riendegris:
Hello,

I like the idea of beeing able to read more formats. If I understand correctly, this is to add the ability to read zipped CSVs.

Is a flag `--gzip` really necessary? Shouldn't the file extension be enough to inform you that you may have to deflate the file prior to reading it?

riendegris:
Il y a une autre solution toute simple, et qui consiste a presque ne rien faire ;-)
```
gunzip file.csv.gz | bano2mimir
```
Presque ne rien faire, car il faut tout de meme faire en sorte que le binaire lise `stdin`. On applique la philosophie unix.

En resume, je vois 2 solutions....

1. Chaque executable fait son petit boulot, libre a l'utilisateur de _composer_  ces executables pour accomplir la tache. => Modifier `bano2mimir` pour lire `stdin`. 
2. Faciliter la tache de l'utilisateur au maximum... implementer une logique de reconnaissance des extensions, et implementer la lecture de ces extensions (gz, zip, bz2).

Je suis assez favorable a la version 1, et je sollicite votre opinion

riendegris:
Sorry Remi, I did not comment fast enough, and you implemented a change before I even finished my train of thoughts...

remi-dupre:
Hello,

Thanks for your review!

> Is a flag `--gzip` really necessary? Shouldn't the file extension be enough to inform you that you may have to deflate the file prior to reading it?

I changed the behavior to decide decompression based on the extension as this is indeed lighter to use and implement. My first intention was to take no risk of regression, but I am very paranoid, there might be no plausible scenario where someone actually calls a plain CSV `file.csv.gz` :smile: 

> Sorry Remi, I did not comment fast enough, and you implemented a change before I even finished my train of thoughts...

Oh sorry about that, I'll stay available for your other thoughts ;)

remi-dupre:
> Il y a une autre solution toute simple, et qui consiste a presque ne rien faire ;-)
> 
> ```
> gunzip file.csv.gz | bano2mimir
> ```
> 
> Presque ne rien faire, car il faut tout de meme faire en sorte que le binaire lise `stdin`. On applique la philosophie unix.

I think it would require a few more changes, but it would indeed be interesting (for example it gets the flexibility to work with any compression algorithm), but looses a bit of generality in the case of *openaddresses2mimir* which can handle directories.

I wouldn't mind going for this option if you prefer it. It could even be possible to keep both but I think it would be confusing.

remi-dupre:
I added the possibility to read addresses from standard input as @riendegris suggested as this is indeed more flexible.

riendegris:
Thanks Remi, it looks good to me. Thank you for the extra work of using stdin.

src/addr_reader.rs
```diff
@@ -60,13 +47,89 @@ where
...
+    F: Fn(T) -> Result<Addr, Error> + Send + Sync + 'static,
+    T: DeserializeOwned + Send + 'static,
+{
+    let streams = files.into_iter().filter_map(|path| {
+        info!("importing {:?}...", &path);
+
+        let with_gzip = path
+            .extension()
+            .and_then(|ext| ext.to_str())
+            .map(|ext| ext == "gz" || ext.ends_with(".gz"))
```
amatissart: `ext == "gz"` is not enough ?

remi-dupre: You're right, this indeed works O_o

```rust
fn main() {
    use std::path::Path;
    
    let path = Path::new("foo.csv.gz");
    assert_eq!("gz", path.extension().unwrap());
}
```

([Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=dcaf51505db405b5c605a4a1c0cc991e))

PR #376, opened 2020-02-20
Log addresses count per country

remi-dupre:
This PR only introduces debug information: it counts the number of imported addresses for each country and display the result.

### DIFF
--- a/src/addr_reader.rs
+++ b/src/addr_reader.rs
@@ -7,6 +7,7 @@ use mimir::Addr;
 use par_map::ParMap;
 use serde::de::DeserializeOwned;
 use slog_scope::{error, info, warn};
+use std::collections::HashMap;
 use std::fs::File;
 use std::io::Read;
 use std::marker::{Send, Sync};
@@ -29,6 +30,9 @@ where
         .with_context(|_| format!("Error occurred when making index {}", dataset))?;
 
     info!("Add data in elasticsearch db.");
+
+    let mut country_stats = HashMap::new();
+
     let iter = addresses
         .into_iter()
         .with_nb_threads(nb_threads)
@@ -46,16 +50,37 @@ where
                 warn!("Address Error ignored: {}", err);
                 None
             }
+        })
+        .inspect(|addr| {
+            let country_code = addr
+                .country_codes
+                .first()
+                .map(|string| string.as_str())
+                .unwrap_or("other");
+
+            if let Some(count) = country_stats.get_mut(country_code) {
+                *count += 1;
+            } else {
+                country_stats.insert(country_code.to_string(), 1);
+            }
         });
 
     let nb = rubber
         .bulk_index(&addr_index, iter)
         .with_context(|err| format!("failed to bulk insert: {}", err))?;
     info!("importing addresses: {} addresses added.", nb);
-
     rubber
         .publish_index(dataset, addr_index, IndexVisibility::Public)
         .context("Error while publishing the index")?;
+
+    info!("Addresses imported per country:");
+    let mut country_stats: Vec<_> = country_stats.into_iter().collect();
+    country_stats.sort_unstable_by_key(|(_, count)| *count);
+
+    for (country, count) in country_stats.into_iter().rev() {
+        info!("{:>10} {}", country, count);
+    }
+
     Ok(())
 }
 


remi-dupre:
Note that it wouldn't be more expensive to generalize these statistics to any kind of item: https://github.com/remi-dupre/mimirsbrunn/pull/1/files

Would it be relevant to add this behavior?

libs/mimir/src/objects.rs
```diff
@@ -1008,3 +1008,43 @@ fn test_normalize_id() {
         "stop_area:SIN:SA:ABCDE:1234"
     );
 }
+
+/// Represent an object which is contained inside an admin hierarchy.
```
riendegris: Represents

libs/mimir/src/objects.rs
```diff
@@ -1008,3 +1008,43 @@ fn test_normalize_id() {
...
+pub trait InsideAdmin {
+    fn parent_admins(&self) -> &[Arc<Admin>];
```
riendegris: parents are immediate relative. I think you mean ancestors

riendegris: need a little comment

libs/mimir/src/objects.rs
```diff
@@ -1008,3 +1008,43 @@ fn test_normalize_id() {
...
+
+    /// Return an arbitrary admin containing this object which satisfies a predicate.
```
riendegris: Returns

riendegris: Wouldn't something more generic be more useful:
The idea is not to lock ourselves in the use of `Arc<...>`, nor in the use of `Admin`.
```
/// Represent an object which is contained inside an admin hierarchy.
pub trait Child {
  type Item;

  /// Returns the list of ancestors of that child.
  fn ancestors<T>(&self) -> &[T]
    where
    T: AsRef<Self::Item>;
/// Returns an arbitrary admin containing this object which satisfies a predicate.
fn find_first_ancestor<P, T>(&self, mut predicate: P) -> Option<T>
where
P: Copy + FnMut(T) -> bool,
 T: AsRef<Self::Item>,
{
self.ancestor()
             .iter()
           .filter_map(|parent| {
               if predicate(parent.as_ref()) {
                    Some(parent.as_ref())
              } else {
                   parent.ancestors(predicate)
              }
            })
 .next()
}
}
```

remi-dupre: I had a thought about it, but by returning an iterator instead of a slice (to avoid unecessary allocations), but in that case I think it would require to specify explicitly the type of the iterator for each implementation of the trait, which may be a bit ugly? :(

riendegris: why is predicate in the declaration of the find_admin required to be `mut`?

riendegris: Specifying the type of iterator, you mean something like the following?

```
impl Child for Stop {
    type Item = Admin;
    fn ancestors(&self) -> &[Arc<Self::Item>] {
        &self.administrative_regions
    }
}
```

remi-dupre: I was wondering on how to avoid allocating a new slice, and afraid of writing something like:

```rust
impl Child for Stop {
    type AdminIterator = iter::Map<iter::Slice<Admin>, Fn(Arc<Admin>, &Admin>>;

    fn ancestors(&self) -> Self::AdminIterator {
        self.administrative_regions.map(|admin| admin.as_ref())
    }
}
```

I think I got a solution which is bot elegant and optimal, what do you think of this?

```rust
pub trait AdminChild {
    type AdminContainer: AsRef<Admin>;

    fn parent_admins(&self) -> &[&Admin];

    fn find_admin<P>(&self, mut predicate: P) -> Option<&Admin>
    where
        P: Copy + FnMut(&Admin) -> bool,
    {
        self.parent_admins()
            .iter()
            .map(|admin| admin.as_ref())
            .filter_map(|parent| {
                if predicate(parent) {
                    Some(parent)
                } else {
                    parent.find_admin(predicate)
                }
            })
            .next()
    }
}
```

This still requires very straightforward implementations:

```rust
impl AdminChild for Street {
    type AdminContainer = Arc<Admin>;

    fn parent_admins(&self) -> &[&Admin] {
        &self.administrative_regions
    }
}
```

Am I over-engineering too much something that doesn't need to be efficient?

remi-dupre: > why is predicate in the declaration of the find_admin required to be `mut`?

This is because the predicate is allowed to be a `FnMut` (which is a bit more generic than `Fn`), thus calls to the predicate may mutate its internal state. I'm not actually sure I can see a use case for this but I think this is a free generalization?

TeXitoi:
FnMut is more general and a returning a slice doesn't allocate.

src/addr_reader.rs
```diff
@@ -29,33 +31,69 @@ where
...
+                    .find_admin(|admin| admin.zone_type == Some(ZoneType::Country))
+                    .map(|admin| {
+                        let country_code = admin.country_codes.iter().min();
+                        country_code.unwrap_or_else(|| &admin.name).as_str()
+                    })
+                    .unwrap_or("other")
+                    .to_string();
+                (addr, country_code)
+            })
+        })
```
amatissart: Is there a reason why `addr.country_codes` could not be reused here ?

remi-dupre: You are right, it is almost equivalent as the importers already use Cosmogony data to fill this field: https://github.com/CanalTP/mimirsbrunn/blob/master/src/bin/openaddresses2mimir.rs#L73.

This also deprecates all our conversation about how to implement the trait used to browse the admins as it wouldn't be required anymore.

It also makes this fell terribly true:

> Am I over-engineering too much something that doesn't need to be efficient?

remi-dupre:
Thanks to @amatissart remark it makes the changes tiny-ish, and much less complicated ;)

src/addr_reader.rs
```diff
@@ -46,16 +50,37 @@ where
...
+                .country_codes
+                .first()
+                .map(|string| string.as_str())
+                .unwrap_or("other");
+
+            if let Some(count) = country_stats.get_mut(country_code) {
+                *count += 1;
+            } else {
+                country_stats.insert(country_code.to_string(), 1);
+            }
```
amatissart: FYI, the [`.entry()`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry) API may be useful in such cases:
```rust
let count = country_stats.entry(country_code).or_insert(0);
*count +=1;
```

remi-dupre: The addresses are dropped at the end of the insertion in elasticsearch so I'd have to allocate a new string for every entry (as I can't build my hashmap on references with too short lifetime), this may be a really tiny performance drawback, is it still worth it?

amatissart: ok, it's very minor, let's keep it this way.

PR #379, opened 2020-03-19
bragi: Follow best practices for prometheus metrics endpoint

amatissart:
Two small fixes about prometheus metrics:

* Do not use an arbitrary path as the "handler" label.
Prometheus recommends not to use metric labels to store an unbounded number of values. In the current implementation, any 404 request with an incorrect path would create a new time series in the prometheus response, that could lead to memory exhaustion in the monitoring stack.  
See https://prometheus.io/docs/practices/naming/#labels for more details  
This PR suggests to use an empty string as a fallback value for unknown paths.

* Set a correct `Content-Type` header value on the /metrics response. 
 Bragi currently sets a `Content-Type: application/json` header on this endpoint, which is incorrect as the prometheus metrics are plain text.

### DIFF
--- a/libs/bragi/src/prometheus_middleware.rs
+++ b/libs/bragi/src/prometheus_middleware.rs
@@ -6,7 +6,7 @@
 use actix_service::{Service, Transform};
 use actix_web::{
     dev::{Body, BodySize, MessageBody, ResponseBody, ServiceRequest, ServiceResponse},
-    http::{Method, StatusCode},
+    http::{header, Method, StatusCode},
     web::Bytes,
     Error,
 };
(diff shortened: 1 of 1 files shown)


libs/bragi/src/prometheus_middleware.rs
```diff
@@ -36,11 +36,12 @@ fn get_ressource_name(path: &str) -> String {
     // so we use an hardcoded associated table
     PATH_TO_NAME
         .get(path)
+        .copied()
```
amatissart: This solves a warning from `cargo clippy` about an inefficient call to `.to_string()` on `&&str`, where `&str` could be used instead.

libs/bragi/src/prometheus_middleware.rs
```diff
@@ -36,11 +36,12 @@ fn get_ressource_name(path: &str) -> String {
...
         .unwrap_or_else(|| {
             if path.starts_with("/features") {
                 &FEATURES_ROUTE
             } else {
-                &path
+                ""
```
antoine-de: shouldn't you put a value here (like `unknown_route`)? you don't need this as you can filter the 404 in grafana?

amatissart: Yes, another label `status` is defined to store the status code.  
Also, it seems that Prometheus makes no distinction between empty and unset labels, which kind of makes sense here as there is actually no `handler` that matched the request.  

But I can replace it with a more explicit value, if you find it appropriate.

PR #380, opened 2020-04-03
Add settings for weights and boosts

GuillaumeGomez:
With the usage of bragi becoming more and more specific, we discussed about a way to allow to configure it. After a few tries, I think this way seems to be the more adequate. All "numbers" have been extracted and can now be updated through the `bragi-settings.json` file by default. It can be replaced using the `--weight-config-file` option, which will allow qwant and kisio to continue having their own without impacting each others.

The remaining questions I have are the following:

 * Is the option name good enough or should I change it?
 * Are the json parameters name good enough or should I change them?

cc @amatissart @riendegris


libs/mimir/src/rubber.rs
```diff
@@ -296,32 +297,142 @@ pub fn get_indexes(
...
+        let settings = serde_json::from_str::<serde_json::Value>(&settings)
+            .map_err(|err| format!("Error occurred when reading bragi settings: {}", err))?;
+        Ok(QuerySettings {
+            type_query: TypeQueryBoosts::new(getter!(settings, "/type_query")?)?,
+            string_query: StringQueryBoosts::new(getter!(settings, "/string_query")?)?,
+            importance_query: ImportanceQueryBoosts::new(getter!(settings, "/string_query")?)?,
+        })
+    }
+}
+
```
remi-dupre, line 408: All this logic could probably be condensed using `serde`'s derive macro ?

libs/mimir/src/rubber.rs
```diff
@@ -296,32 +297,142 @@ pub fn get_indexes(
...
+        info!("elastic search host {} ", cnx);
+
         let timeout = timeout.into();
         Rubber {
             es_client: rs_es::Client::init_with_timeout(&cnx, timeout).unwrap(),
             http_client: reqwest::Client::builder().timeout(timeout).build().unwrap(),
             cnx_string: cnx.to_owned(),
             timeout,
+            query_settings: QuerySettings::new(include_str!("../../../json/bragi-settings.json"))
+                .expect("failed to build query settings"),
```
remi-dupre, line 428: As you said it would be better to have a setting to choose which file to use, and to load it during runtime :)

libs/bragi/src/query.rs
```diff
@@ -186,13 +187,13 @@ fn build_query<'a>(
...
-            match_type_with_boost::<Street>(1.),
+            match_type_with_boost::<Addr>(query_settings.type_query.address),
+            match_type_with_boost::<Admin>(query_settings.type_query.admin),
+            match_type_with_boost::<Stop>(query_settings.type_query.stop),
+            match_type_with_boost::<Poi>(query_settings.type_query.poi),
+            match_type_with_boost::<Street>(query_settings.type_query.street),
         ])
-        .with_boost(30.)
+        .with_boost(query_settings.type_query.global)
         .build();
```
amatissart, line 197: there are also boost coefficients and other hard-coded value defined in the helpers functions `build_with_weight` and `build_proximity_with_boost`. They should be configurable too.

riendegris:
Hello, thanks for addressing this issue.

A little header for the issue stating what the goal is would probably not hurt.... ;-)

If I recall correctly, the primary goal is to be able to have multiple behaviors (read ranking of results) depending on which set of parameters we choose.
A secondary goal would be to extract all these coefficients so that they can more easily be tweaked.

I haven't look into the performance aspect, but I wonder if there is a penalty to be paid for each client if we're reading from file every time. What about using a set of constants? I haven't given much thought, but what if we introduced features like 'qwant', 'kisio', ... and each would provide a different set of constants (probably with defaults one if no feature is choosen) optimized for each particular needs. So coefficients would be configurable at compile time and baked in the code at runtime...

GuillaumeGomez:
Are we reading from the file every time? :o If so, I can use a static instead (didn't want to because I assumed `Context` was instantiated just once).

Also sorry, for now it's mostly experimentations, so once I'm happy with the current content. I'll add a description and everything missing.

libs/mimir/src/rubber.rs
GuillaumeGomez: As said in a conversation: I prefer this way in here because `serde_json` error messages aren't that great. It's even more true in case you have a file modified by users.

libs/mimir/src/rubber.rs
GuillaumeGomez: There is now an option. :)

libs/bragi/src/query.rs
GuillaumeGomez: They are now as well.

GuillaumeGomez:
I think I'm now done with the base, so as promised, I updated the PR description. :)

libs/bragi/src/lib.rs
```diff
@@ -128,6 +139,7 @@ impl From<&Args> for Context {
                 .or_else(|| max_es_timeout.clone())
         };
 
+        let content = read_to_string(&args.weight_config_file).expect("cannot read file");
```
riendegris, line 142: Ideally a trait TryFrom returning a result would improve error handling
At the very least if I see 'cannot read file' when I start bragi... I won't know what file is missing. I think it should be more specific

libs/bragi/src/lib.rs
```diff
@@ -143,6 +155,7 @@ impl From<&Args> for Context {
             ),
             cnx_string: args.connection_string.clone(),
             http_cache_duration: args.http_cache_duration.clone(),
+            query_settings: QuerySettings::new(&content).expect("failed to build query settings"),
```
riendegris, line 158: Same thing as previously mention... a TryFrom trait is probably better

libs/bragi/src/lib.rs
```diff
@@ -157,6 +170,9 @@ impl Context {
     pub fn get_rubber_for_autocomplete(&self, timeout: Option<Duration>) -> Rubber {
         clone_or_create(&self.autocomplete_rubber, timeout)
     }
+    pub fn get_query_settings(&self) -> &QuerySettings {
+        &self.query_settings
+    }
```
riendegris, line 175: That does seem a bit superfluous... I will have to write
`self.get_query_settings();` vs `self.query_settings;`

json/bragi-settings.json
```diff
@@ -0,0 +1,55 @@
+{
```
riendegris, line 1: json... hmmm, json
Not sure it's the most appropriate format here, but then it depends how it's going to be used
For manual update, I'd say know because you can't put comment, so it's difficult to know what knob you're touching
For automatic update, why not...

libs/mimir/src/rubber.rs
```diff
@@ -299,20 +299,15 @@ pub fn get_indexes(
...
-            cnx_string: cnx.to_owned(),
-        }
+        Rubber::new_with_timeout(cnx, None)
     }
 
     pub fn new_with_timeout<T>(cnx: &str, timeout: T) -> Rubber
     where
         T: Into<Option<time::Duration>>,
     {
+        info!("elastic search host {} ", cnx);
```
riendegris, line 309: Can I start bikeshedding about levels of logging?

GuillaumeGomez: Absolutely!

libs/bragi/src/lib.rs
GuillaumeGomez: Oh, well I did actually before this comment haha.

libs/bragi/src/lib.rs
GuillaumeGomez: Same answer. :)

libs/bragi/src/lib.rs
GuillaumeGomez: I like to keep inner fields hidden. But I can make it public if you prefer.

json/bragi-settings.json
GuillaumeGomez: We can give a try to toml as well?

libs/mimir/src/rubber.rs
amatissart: This message actually annoyed me when inspecting some import logs.  
It that's what you implied @riendegris, I would agree to change it to `debug!`

GuillaumeGomez: I changed it.

libs/bragi/src/query.rs
```diff
@@ -117,51 +118,58 @@ fn build_coverage_condition(pt_datasets: &[&str]) -> Query {
...
-fn build_with_weight<A: Into<Option<f64>>>(factor: A) -> Query {
-    let factor = factor.into();
+fn build_with_weight(build_weight: &BuildWeight, types: &Types) -> Query {
     let weighted = |doc_type, weight| {
         FilteredFunction::build_filtered_function(
             Query::build_term("_type", doc_type).build(),
             Function::build_field_value_factor("weight")
-                .with_factor(factor.unwrap_or(0.75))
-                .with_missing(0.0)
+                .with_factor(build_weight.build)
```
amatissart, line 159: This name could be made more explicit.  
What about `build_weight.factor` or `build_weight.weight_factor` ?

GuillaumeGomez: I'm fine with both? Do you have a preferred one by any chance?

libs/bragi/src/query_settings.rs
```diff
@@ -0,0 +1,170 @@
+macro_rules! getter {
+    ($settings:expr, $to_get:expr, $extra:ident) => {{
+        let tmp = getter!($settings, $to_get)?;
+        tmp.$extra().ok_or(format!(
+            "Failed to call `{}` on `{}` entry from bragi-settings.json",
```
remi-dupre, line 5: `bragi-settings.json` can actually be changed now (same at line 12)

GuillaumeGomez: Absolutely!

libs/bragi/src/query.rs
amatissart: `build_weight.factor` looks good to me

PR #383, opened 2020-04-21
Increment version => 1.15

riendegris:
increment versions across all projects within mimirsbrunn


libs/tools/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "tools"
-version = "0.0.1"
+version = "0.15.0"
```
amatissart, line 3: Is that intentional to keep "tools" version as 0.x ?  
Anyway it's very minor as it's used only in tests.

PR #384, opened 2020-04-22
Fix and extends Gitlab CI

riendegris:
This extend the CI pipeline to include
several checks:
- format
- lint
- coverage

It also includes the changes to pass the first
two checks (format and lint). Some of these changes
are temporary by allowing clippy to disregard the warning.


.gitlab-ci.yml
```diff
@@ -0,0 +1,68 @@
...
+  script:
+    - rustup component add rustfmt
+    - cargo fmt --all -- --check
+
+lint:
+  stage: check
+  image: rust:latest
+  script:
+    - rustup component add clippy
+    - cargo clippy -- -D warnings
```
woshilapin, line 38: You can also use `--all-targets` to use `clippy` on tests, examples, benches, and most of the other Rust files.

Think also of `--workspace` for building all the crates in the repository.

And finally, think about `--all-features` to really test all the code.

libs/bragi/src/lib.rs
```diff
@@ -28,6 +28,7 @@
 // https://groups.google.com/d/forum/navitia
 // www.navitia.io
 
+#![allow(clippy::option_as_ref_deref)]
```
woshilapin, line 31: You should be able to remove that line and replace in the code with
```diff
@@ -165,8 +164,7 @@ impl TryFrom<&Args> for Context {
                 format!(
                     "failed to parse `{}`: {}",
                     args.weight_config_file
-                        .as_ref()
-                        .map(|x| x.as_str())
+                        .as_deref()
                         .unwrap_or_else(|| "config/bragi-settings.toml"),
                     e
                 )
```

libs/bragi/src/model.rs
```diff
@@ -27,7 +27,7 @@
 // IRC #navitia on freenode
 // https://groups.google.com/d/forum/navitia
 // www.navitia.io
-
+#![allow(dead_code)]
```
woshilapin, line 30: Why not removing this line and also removing the variant `BragiError::IndexNotFound`? `BragiError` is not exposed outside of the crate and the variant `IndexNotFound` is not used inside the crate...

libs/bragi/src/query.rs
```diff
@@ -27,6 +27,7 @@
 // IRC #navitia on freenode
 // https://groups.google.com/d/forum/navitia
 // www.navitia.io
+#![allow(clippy::too_many_arguments)]
```
woshilapin, line 30: This can be done at each function level which is more fine-grained (note the syntax is slightly different, there is `!` removed).

```diff
--- a/libs/bragi/src/query.rs
+++ b/libs/bragi/src/query.rs
@@ -27,7 +27,6 @@
 // IRC #navitia on freenode
 // https://groups.google.com/d/forum/navitia
 // www.navitia.io
-#![allow(clippy::too_many_arguments)]
 use super::model::{self, BragiError};
 use crate::query_settings::{BuildWeight, Proximity, QuerySettings, Types};
 use geojson::Geometry;
@@ -176,6 +175,7 @@ fn build_with_weight(build_weight: &BuildWeight, types: &Types) -> Query {
         .build()
 }
 
+#[allow(clippy::too_many_arguments)]
 fn build_query<'a>(
     q: &str,
     match_type: MatchType,
@@ -417,6 +417,7 @@ fn build_query<'a>(
     query.build()
 }
 
+#[allow(clippy::too_many_arguments)]
 fn query(
     q: &str,
     pt_datasets: &[&str],
@@ -565,6 +566,7 @@ pub fn features(
     }
 }
 
+#[allow(clippy::too_many_arguments)]
 pub fn autocomplete(
     q: &str,
     pt_datasets: &[&str],
```

libs/bragi/src/query.rs
```diff
@@ -482,17 +483,20 @@ fn query(
...
     };
 
     if let Some(timeout) = &timeout {
         search_query.with_timeout(timeout.as_str());
     }
     let result = search_query.send()?;
 
-    timer.map(|t| t.observe_duration());
+    if let Some(t) = timer {
+        t.observe_duration()
```
woshilapin, line 498: You can probably add a `;` since the returned value is not used.

```suggestion
        t.observe_duration();
```

libs/mimir/src/lib.rs
```diff
@@ -83,6 +83,6 @@ where
 
     let log = slog::Logger::root(drain.fuse(), slog_o!());
     let scope_guard = slog_scope::set_global_logger(log);
-    let log_guard = slog_stdlog::init().unwrap();
-    (scope_guard, log_guard)
+    slog_stdlog::init().unwrap();
+    (scope_guard, ())
```
woshilapin, line 87: Maybe you can change the signature of the function to only return `slog_scope::GlobalLoggerGuard` instead of a tuple.

libs/mimir/src/objects.rs
```diff
@@ -27,6 +27,7 @@
 // IRC #navitia on freenode
 // https://groups.google.com/d/forum/navitia
 // www.navitia.io
+#![allow(clippy::large_enum_variant, clippy::needless_lifetimes)]
```
woshilapin, line 30: I believe this line can be removed. See

```diff
diff --git a/libs/mimir/src/objects.rs b/libs/mimir/src/objects.rs
index b4ca5b6..ac9d616 100644
--- a/libs/mimir/src/objects.rs
+++ b/libs/mimir/src/objects.rs
@@ -27,7 +27,6 @@
 // IRC #navitia on freenode
 // https://groups.google.com/d/forum/navitia
 // www.navitia.io
-#![allow(clippy::large_enum_variant, clippy::needless_lifetimes)]
 use cosmogony::ZoneType;
 use geo_types::{Coordinate, MultiPolygon, Rect};
 use geojson::Geometry;
@@ -52,6 +51,7 @@ pub trait Incr: Clone {
 
 /// Object stored in elastic search
 #[derive(Serialize, Deserialize, Debug, Clone)]
+#[allow(clippy::large_enum_variant)]
 pub enum Place {
     Admin(Admin),
     Street(Street),
@@ -61,6 +61,7 @@ pub enum Place {
 }
 
 /// Object stored in elastic search
+#[allow(clippy::large_enum_variant)]
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[serde(rename_all = "snake_case", tag = "type")]
 pub enum Address {
@@ -646,7 +647,7 @@ where
     })
 }
 
-pub fn serialize_rect<'a, S>(bbox: &'a Option<Rect<f64>>, serializer: S) -> Result<S::Ok, S::Error>
+pub fn serialize_rect<S>(bbox: &Option<Rect<f64>>, serializer: S) -> Result<S::Ok, S::Error>
 where
     S: serde::Serializer,
 {
```

src/osm_reader/street.rs
```diff
@@ -27,6 +27,12 @@
 // IRC #navitia on freenode
 // https://groups.google.com/d/forum/navitia
 // www.navitia.io
+#![allow(
```
woshilapin, line 30: Most of these are due to the macro `err_logger`. I took a quick glance at it and one part of the problem might be that this macro is returning different objects depending of if it is called with 2 (returns `Option<T>` and makes use of `?`) or 3 parameters (returns a `T`, type of `$ret`). Maybe this needs 2 macros?

.gitlab-ci.yml
riendegris: :+1:

PR #385, opened 2020-04-23
Gitlab CI/CD and clippy suggestions

riendegris:
This PR (re)introduces a Gitlab CI/CD pipeline, in which I include formatting, linting, and code coverage. A large portion of this PR includes the changes suggested by running clippy.

Cases that require deeper changes have been left out, and an appropriate `#[allow(clippy::XXX)]` marker has been inserted to prevent clippy from returning an error. These cases will have to be addressed later.


libs/bragi/src/model.rs
```diff
@@ -28,6 +28,8 @@
 // https://groups.google.com/d/forum/navitia
 // www.navitia.io
 
+// #![allow(dead_code)]
```
GuillaumeGomez, line 31: If not used, please remove it.

libs/mimir/src/objects.rs
```diff
@@ -50,21 +50,23 @@ pub trait Incr: Clone {
...
 
 /// Object stored in elastic search
+#[allow(clippy::large_enum_variant)]
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub enum Place {
     Admin(Admin),
     Street(Street),
     Addr(Addr),
-    Poi(Poi),
+    Poi(Poi), // Boxing for avoiding large enum variants
```
GuillaumeGomez, line 59: I discovered about this lint yesterday. So in here, we should definitely do it instead of just commenting, don't you think?

libs/mimir/src/objects.rs
```diff
@@ -50,21 +50,23 @@ pub trait Incr: Clone {
...
 }
 
 /// Object stored in elastic search
+#[allow(clippy::large_enum_variant)]
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[serde(rename_all = "snake_case", tag = "type")]
 pub enum Address {
     Street(Street),
-    Addr(Addr),
+    Addr(Addr), // Boxing for avoiding large enum variants
```
GuillaumeGomez, line 69: Same.

libs/mimir/src/objects.rs
```diff
@@ -648,6 +647,7 @@ where
     })
 }
 
+#[allow(clippy::needless_lifetimes)]
 pub fn serialize_rect<'a, S>(bbox: &'a Option<Rect<f64>>, serializer: S) -> Result<S::Ok, S::Error>
```
GuillaumeGomez, line 651: This is actually a good lint: why do we use a named lifetime in here?

libs/tools/src/lib.rs
```diff
@@ -267,10 +267,10 @@ impl BragiHandler {
         let (status, s) = self.raw_post(q, shape);
         assert!(status.is_success(), "invalid status: {}", status);
 
-        self.to_json(s)
+        self.json(s)
     }
 
-    pub fn to_json(&mut self, b: bytes::Bytes) -> Value {
+    pub fn json(&mut self, b: bytes::Bytes) -> Value {
```
GuillaumeGomez, line 273: Why this renaming?

GuillaumeGomez: Eventually (if we still want to rename it), I'd pick `as_json` just like `as_str` for `String`.

src/osm_reader/street.rs
```diff
@@ -27,6 +27,12 @@
 // IRC #navitia on freenode
 // https://groups.google.com/d/forum/navitia
 // www.navitia.io
+#![allow(
+    clippy::unused_unit,
+    clippy::needless_return,
```
GuillaumeGomez, line 32: This one might be interesting to fix, no?

src/osm_reader/street.rs
```diff
@@ -461,7 +465,7 @@ pub fn streets(
                     .collect();
                 name_admin_map
                     .entry(StreetKey { name, admins })
-                    .or_insert(vec![])
+                    .or_insert_with(|| vec![])
```
GuillaumeGomez, line 468: In fact, in the current case it's basically an option `Vec` versus a closure (+ eventually an empty `Vec`). Not sure if it's that useful...

libs/mimir/src/objects.rs
riendegris: I agree we should do it... soon

libs/bragi/src/model.rs
riendegris: yes

libs/tools/src/lib.rs
riendegris: The lint came in because of the naming convention (i forgot the exact name of the rule). But basically if your function is called 'to_xxx', it should not take &mut self as argument but self.
This is why I choose `json`. `as_json` crossed my mind, but that usually implies a conversion, which it is not.

src/osm_reader/street.rs
riendegris: I think the whole macro needs to be reevaluated.... soon. Remember we can grep for `clippy` and quickly find all those places left behind. I think there is only a handful

libs/tools/src/lib.rs
riendegris: using `as_json`

libs/mimir/src/objects.rs
```diff
@@ -648,6 +647,7 @@ where
     })
 }
 
+#[allow(clippy::needless_lifetimes)]
```
GuillaumeGomez, line 650: Still there.

riendegris: done, actually, wasn't terribly complicated...

libs/mimir/src/objects.rs
riendegris: See #388

src/osm_reader/street.rs
riendegris: See #389

PR #390, opened 2020-05-04
Change hardcoded -broken- bash path for *nixes

berkes:
For portability /usr/bin/bash should not be used: it does not exist on
most *nixes (instead it is /bin/bash on most Linuxes). /usr/bin/env bash
can be run on most *nixes including BSD families and Linuxes.

See: https://stackoverflow.com/a/10383546/73673

### DIFF
--- a/scripts/import2mimir.sh
+++ b/scripts/import2mimir.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/bash
+#!/usr/bin/env bash
 
 set -o errexit
 set -o nounset


berkes:
For search indexes and future reference. The error (loosely translated from my locale:)

```
./scripts/import2mimir.sh 
bash: ./scripts/import2mimir.sh: /usr/bin/bash: invalid interpreter: File or directory does not exist
```

riendegris:
Thank you, I didn't know that, even after using bash for years!

PR #391, opened 2020-05-04
Move a hardcoded europe/france reference into config.

berkes:
This replaces a hardcoded /europe/france/ link to a configurable option. It allows any region to be downloaded from geofabrik instead of only French regions.


berkes:
This change causes issues with the use of ${OSM_REGION}-latest down the line. 

I'm splitting the config into $OSM_DOWNLOAD_URL and $OSM_REGION that fixes this. Currently testing and re-running with various settings.

berkes:
After commit:df05852 we can now set a `OSM_DOWNLOAD_URL` as well as a `OSM_REGION`.

The latter is not strictly nessecary, but is used all over the script so removing that requires a giant refactoring.

This change is not backwards compatible: the script now requires a `OSM_DOWNLOAD_URL` and will fail if that is not yet set.

riendegris:
Hello @berkes,

I am reluctant to accept this PR. I really would like the rc file to just contain region definition, and then the script to figure out, based on these values, what URL should be used. I think this is what you refer to in your comment as a 'giant refactoring'. I'll try my hands at it maybe today. Don't you think it would be better to limit the number of variables to define in the rc file?

berkes:
> Don't you think it would be better to limit the number of variables to define in the rc file?

Yes. 

And in my current fork (which I need to import the Netherlands on production, so no docker, openaddress instead of pano and more) I've worked it out to have the config file define `$OSM_DOWNLOAD_URL` and have the script figure out `$OSM_REGION` from that.

So exactly the other way around then you suggest.

The reason is simple: I want to both allow `http://download.geofabrik.de/europe/cyprus-latest.osm.pbf` or `http://download.geofabrik.de/europe/germany/nordrhein-westfalen/koeln-regbez-latest.osm.pbf` or even `https://planet.openstreetmap.org/pbf/planet-latest.osm.pbf`.
And, for some weird requirement, I also need to allow importing dated osm files, like `http://download.geofabrik.de/europe/germany/berlin-200502.osm.pbf`

As you can see, all URLS are structured differently: the first has europe/country-latest, the second `europe/country/region/subregion-latest` the third is and entirely different host and the last is rather arbitrary.

Since the script does not *need* a region at all, other than to download/name/find a temporary file, the refactoring could be easy: download the file into a predictable file: 
`wget --output-document=${DATA_DIR}/osm/data.pbf` which, coincidentally, fixes a bug in the current script where wget now writes `osm/ile-de-france.pbf.1` files on second run. which are then ignored.

berkes:
My fork is here: https://github.com/berkes/mimirsbrunn/blob/feature/deploy/scripts/import2mimir.sh

This is from where I backported several features that I deemed candidates for upstream. Most changes are only needed for my own import so I keep them there.

PR #400, opened 2020-05-25
Round coordinates used in ID for OpenAddresses

remi-dupre:
Add an option to specify the max precision of of coordinate values in addresses ID.

This option can be disabled. In our use case if an addresses has unnecessary precise coordinates, we end up with inconvenient URLs.

~~**NOTE**: I am not fully satisfied with how I implemented this as I relied on float arithmetics to perform the rounding: I don't know floats well enough to ensure there can't be any corner case where round(10 * x) / 10 has tons of floating points. A more reliable approach would be to perform this rounding when the coordinates are parsed to a string before being sent to ES, however I couldn't think of a way to implement that without making the code unnecessarily complicated (current implementation just relies on the `Serialize` trait auto implemented by serde, which is pretty compact).~~

### DIFF
--- a/src/bin/openaddresses2mimir.rs
+++ b/src/bin/openaddresses2mimir.rs
@@ -39,6 +39,7 @@ use std::io::stdin;
 use std::ops::Deref;
 use std::path::PathBuf;
 use structopt::StructOpt;
+use tools::round_at;
 
 lazy_static! {
     static ref DEFAULT_NB_THREADS: String = num_cpus::get().to_string();
@@ -64,6 +65,7 @@ impl OpenAddress {
         self,
         admins_geofinder: &AdminGeoFinder,
         use_old_index_format: bool,
+        coord_precision: u32,
     ) -> Result<mimir::Addr, mimirsbrunn::Error> {
         let street_id = format!("street:{}", self.id); // TODO check if thats ok
         let admins = admins_geofinder.get(&geo::Coordinate {
@@ -87,7 +89,11 @@ impl OpenAddress {
             &country_codes,
         );
 
-        let coord = mimir::Coord::new(self.lon, self.lat);
+        let coord = mimir::Coord::new(
+            round_at(self.lon, coord_precision),
+            round_at(self.lat, coord_precision),
+        );
+
         let street = mimir::Street {
             id: street_id,
             name: self.street,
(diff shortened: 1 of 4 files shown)


src/bin/openaddresses2mimir.rs
```diff
@@ -64,12 +65,24 @@ impl OpenAddress {
...
+                    x: self.lon,
+                    y: self.lat,
+                }
+            } else {
+                geo::Coordinate {
+                    x: round_at(self.lon, coord_precision),
+                    y: round_at(self.lat, coord_precision),
+                }
+            }
         });
```
amatissart, line 84: This change only applies on coordinates used to find admins where the address is located. 
But the `id` is still the same: https://github.com/CanalTP/mimirsbrunn/pull/400/files#diff-cbe83f968989226ba55b268b952379efR120
Am I missing something ?

remi-dupre: It made no sense indeed (I got lost when I reverted my attempts to cleanly tweak the serializer), I fixed this and added a little test for this feature.

src/bin/openaddresses2mimir.rs
```diff
@@ -105,8 +111,8 @@ impl OpenAddress {
         Ok(mimir::Addr {
             id: format!(
                 "addr:{};{}{}",
-                self.lon,
-                self.lat,
+                coord.lon(),
+                coord.lat(),
```
amatissart, line 115: Are we sure that something like  `0.1234559999999999999` can't be rendered in the string here ?  
I think it would be more robust to use the precision options built-in with `format!`:
https://doc.rust-lang.org/std/fmt/#precision

remi-dupre: I also think it could be safer but in this case should we not affect actual coordinates of the address and only the ID? Do you think it would be strange?

amatissart: As you wish. I'm ok with updating the id only.
The main feature of the id is to be as stable as possible. I don't think there is any problem if the actual coordinates in the document don't correspond exactly with the document id. 
The name of the command-line option should make the choice clear though.

remi-dupre: I just fixed this, the new parameter name `id-precision` affects the ID in a similar way without affecting actual coordinates.

amatissart:
@riendegris Do you have any opinion about this change ?
The default behavior would be to round lat/lon values used in document ids imported with `openaddresses2mimir` to 6 decimals.

PR #401, opened 2020-06-02
Move synonyms into a textfile for easier overrides.

berkes:
This PR extracts the hardcoded synonyms into a compile-time-included text file.

This makes it a bit easier for users to translate, or otherwise alter the synonyms.

It  is still compiled in, like all the other index-settings, so recompilation is needed in order to change synonyms.


riendegris:
Hello Ber,

Thanks for this PR, which makes a lot of sense. Do we know what happens when the config file is not present? What should happen if it not formatted properly? A warning, and an empty list of synonyms, or an error?

Also, it seems like a json file array of strings, so maybe the extension should be '.json'.

Looking a bit ahead, how do you think we should make it a configurable option, so that in eg the netherlands you have a different set of synonyms than in France?

berkes:
Since it is parsed into the JSON, it is not valid JSON, hence I went for .txt.

I can make it valid JSON, though, and then extract the relevant part from it, in the code. Will do that in a follow-up commit in this branch/PR.

It is included on compile-time, so a missing file etc will throw an error on compiling or checking. The same as with all the other .json files. 

`error: couldn't read libs/mimir/src/../../../config/synonyms.txt: No such file or directory (os error 2)```

Formatting the file wrong, is nearly impossible, as it simply parses the lines and includes those. I don't know any trick to break this. The compiler checks if it is valid UTF8 text:

`error: libs/mimir/src/../../../config/synonyms.txt wasn't a utf-8 file`

Formatting wrong for ES is another thing, elasticsearch has its own synonym sintax and will cause an error (or maybe ignore it, I don't know ES2 that well?) (probably causing a panic?) if one or more lines are invalid synonym-syntax. This is no different from the old behaviour.

About configurability: synonyms are defined as filters on indexes and having locale-aware synonyms is not possible. Well, one could create a per-language filter, e.g. `synonyms_en`, `synonyms_ru` etc, and then configure to use those on the index and fields depending on the language setting when creating the index. This requires the import to pick the file from disk based on language.

However, the current index-settings are all tuned for france already. Not just the synonyms, but also stemmer, and boosts depend on locale and language. So if we were to implement this, I'd go one step further and implement a cascading locale-aware config: `config/default/`, `config/fr_FR/`, config/nl_NL/` etc. The files in `config/fr_FR/` would then override the ones in `config/default/`. 
That is a different PR, though.

berkes:
When turning the synonyms into a json-file, I realized that the current and old situation cuold just as well have the synonyms in the settings.json file. Since they are merged in there, regardless.

This simplifies the code a lot more.

PR #402, opened 2020-06-02
WIP: Let both GET and POST autocomplete handle a json shape in the body.

berkes:
GET can handle a body in HTTP, as is seen in e.g. Elasticsearch which we
use a lot. Some client libs have difficulty with this, so we should
leave the POST in place as well.

This makes both behave the same. Making the code a little easier by
removing some wrapper function callsLet both GET and POST autocomplete
handle a json shape in the body.

GET can handle a body in HTTP, as is seen in e.g. Elasticsearch which we
use a lot. Some client libs have difficulty with this, so we should
leave the POST in place as well.

This makes both behave the same. Making the code a little easier by
removing some wrapper function calls.

### DIFF
--- a/libs/bragi/src/routes/autocomplete.rs
+++ b/libs/bragi/src/routes/autocomplete.rs
@@ -134,13 +134,24 @@ impl JsonParams {
     }
 }
 
-pub fn call_autocomplete(
-    params: &Params,
-    state: &Context,
-    shape: Option<Geometry>,
+pub fn autocomplete(
+    params: BragiQuery<Params>,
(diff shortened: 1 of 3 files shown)


berkes:
The rebase on master went wrong, tests are -rightfully- failing. I'm fixing that now.

berkes:
I cannot find a way to have the extractor validations trigger properly for an Option<T>.
For some reason, somewhere, actix (extractors?) turn this into a None if they cannot parse it, rather than returning the error. 

I'm giving up on this.

PR #403, opened 2020-06-03
send batches from several threads in Rubber's bulk_index

remi-dupre:
Send `bulk_index`'s batches in parallel. The number of threads can be defined in a distinct CLI parameter `-T`.

Note that when the number of threads is too high (eg. `-T=16` on my local tests), Elasticsearch may crash during the import (the max number of threads is probably related to the number of shards?). Thus, the default number of threads is set to 1 in order to make sure that the caller sets the right amount.

Performances
---

Here are some benchmarks obtained by importing OpenAddresses data for Denmark (~3M8 addresses) using openaddresses2mimir.

##### ES on the same machine as openaddresses2mimir

(my CPU already struggles at `--nb-insert-threads=2`, I can't get any better result).

|    branch   |      |
|:-----------:|-----:|
|    master   | 5:29 |
| HEAD (-T 1) | 5:35 |
| HEAD (-T 2) | 4:33 |

##### ES on a separate machine

I tried to distribute the CPU stress of the import using a second machine in order to get more room for improvement. The issue here is that it seems that my networking is introducing a lot of variations.

|    branch   |      |
|:-----------:|-----:|
|    master   | 12:39 |
| HEAD (-T 1) | 13:04 |
| HEAD (-T 8) | 6:29 |

*(edit)*

We just imported BANO in our development pipeline using this branch and 5 threads to communicate with ES. We registered a x3.5 improvement speedup:

|    branch   |      |
|:-----------:|-----:|
|    master   | ~2 h 33 min |
| HEAD (-T 5) | ~42 min |

### DIFF
--- a/libs/mimir/src/rubber.rs
+++ b/libs/mimir/src/rubber.rs
@@ -93,6 +93,7 @@ pub struct Rubber {
     // Note: The timeout is used for the http client AND for the ES internal query
     pub timeout: Option<time::Duration>,
     pub cnx_string: String,
+    pub nb_insert_threads: usize,
 }
 
 #[derive(Clone, Debug)]
@@ -314,7 +315,16 @@ impl Rubber {
             http_client: reqwest::Client::builder().timeout(timeout).build().unwrap(),
             cnx_string: cnx.to_owned(),
             timeout,
+            nb_insert_threads: 1,
         }
+    }
+
+    /// Change the number of parallel threads used to insert data into Elasticsearch
+    /// during `bulk_index`. Note that if too many insertions are performed in parallel,
+    /// Elasticsearch may raise an error.
+    pub fn with_nb_insert_threads(mut self, value: usize) -> Self {
+        self.nb_insert_threads = value;
+        self
     }
 
     pub fn get(&self, path: &str) -> Result<reqwest::Response, EsError> {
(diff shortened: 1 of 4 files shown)


libs/mimir/src/rubber.rs
```diff
@@ -632,27 +642,34 @@ impl Rubber {
...
+                            .into_iter()
+                            .fold(Action::index(v), |action, id| action.with_id(id))
+                    })
+                    .collect::<Vec<_>>()
+            })
+            .with_nb_threads(self.nb_insert_threads)
+            .par_map(move |chunk| {
+                rs_es::Client::init_with_timeout(&cnx_string, timeout)
+                    .expect("invalid ES connection string")
+                    .bulk(&chunk)
```
amatissart, line 665: Instead of duplicating the initialization, we could simply clone the existing  `es_client`.
(a bit ugly, admittedly)
```suggestion
            .par_map({
                let client = self.es_client.clone();
                move |chunk| {
                    client
                        .clone()
                        .bulk(&chunk)
                        ...
```

remi-dupre: That's actually better, for some reason I was convinced that the ES client wasn't `Sync`

remi-dupre:
I'll mark this as WIP while we run more extensive tests

remi-dupre:
We just imported BANO in our development pipeline using this branch and 5 threads to communicate with ES. We registered a x3.5 improvement speedup:

|    branch   |      |
|:-----------:|-----:|
|    master   | ~2 h 33 min |
| HEAD (-T 5) | ~42 min |

PR #410, opened 2020-07-02
Forbid to search by house number only

remi-dupre:
When a query only consists of a single word, this PR will prevent it from matching the house number.

### Motivation

We noticed that when importing addresses from countries with potentially large house numbers (such as the US), some house numbers can have the same value as a postcode for a city. In such a case and when querying for the postcode of the city, the result will be likely to be polluted with addresses:

```
http://<BRAGI_URL>/autocomplete?q=92220

 1. 92220 OVERSEAS HWY (Key Largo)
 2. 92220 YOUNGS RIVER RD
 3. 92220 LEWIS & CLARK RD
 4. 92220 MARCOLA RD (Marcola)
 5. 92220 WHITMORE ST (Marcola)
 6. Bagneux (92220), Hauts-de-Seine, Île-de-France, France
```

This behavior is sometime less problematic when focus is enabled, however with some smaller cities it can be really hard to find a city from its postcode.

While this problem could probably be solved (at least partially ?) by tweaking coefficient, I don't think there may be a use case where a user expects to find a complete address by typing only the house number. Thus I think we can safely avoid the issue by removing any address result when a single word is searched.

Results from this PR:

```
http://<BRAGI_URL>/autocomplete?q=92220

 1. Bagneux (92220), Hauts-de-Seine, Île-de-France, France
 2. Олександропіль (92220), Білокуракинський район, Луганська область, Україна
 3. Ляшківка (92220), Білокуракинський район, Луганська область, Україна
 4. Cimetière parisien de Bagneux (Bagneux)
 5. Église Saint-Hermeland (Bagneux)
 6. Banning Municipal Airport (Banning)
```

### Tests

Running `geocoder-tester` on an instance of Bragi using an ES database populated with OA and OSM data for the planet didn't highlight any regression (21 new passing, 4 new failures).


riendegris:
The PR looks good to me. @remi-dupre, do you think its necessary to add tests for situation where eg a street name would start with a number (like '11 novembre'). I don't think it would cause a problem with your code

PR #411, opened 2020-07-15
Stop overriding package version at compile time

remi-dupre:
We are currently using a build hook for Bragi that overrides the version number specified in `Cargo.toml`. The same behavior is implemented for the top-level package but I don't think we actually take benefit of it.

I think it is a rather heavy behavior to keep as it implies that cargo will rebuild all binaries at each invocation, which is rather slow.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,4 +1,3 @@
-use git_version;
 use json;
 
 use std::ffi::OsStr;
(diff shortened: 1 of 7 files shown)


remi-dupre:
I added [this commit](https://github.com/CanalTP/mimirsbrunn/pull/411/commits/5ec660fd1cce72e3b466e73a18dc0041c59a5c8b) which implements the same behavior as what is done on `master`, but without passing git version through `CARGO_PKG_VERSION` during a "pre-build".

*FYI*, version given by `<BRAGI_URL>/status` is formatted as follows (on both master and this branch):

 ```
v{last github tag}-{commit number}-{commit hash}{"-modified" if some changes have been done since last commit}
```

File: build.rs [hunk 1]
```diff
@@ -13,6 +12,11 @@ fn validate_json_files(dir: &Path) -> io::Result<()> {
         if path.is_dir() {
             validate_json_files(&path)?;
         } else if path.extension() == Some(OsStr::new("json")) {
+            println!(
```
riendegris, line 15: Maybe add a reference to https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed to explain why we do this... (I had to look it up, I initially thought that it was a comment)

PR #417, opened 2020-08-19
allow to change the radius of proximity queries

remi-dupre:
Add query parameters `proximity_scale`, `proximity_offset` and `proximity_decay` which allows to tweak the weight computed from the distance in proximity queries.

These parameters reflect the parameters used by Elasticsearch itself. See [this](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html#function-decay) for more details.

## Current behavior

The score of such a query is computed from the distance of documents to input `lat`/`lon` based on a gaussian distribution:

[](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html#_supported_decay_functions)

Currently, this score function is the same for all queries, only tweakable through settings, [by default](https://github.com/CanalTP/mimirsbrunn/blob/master/config/bragi-settings.toml#L49-L51) we currently have:

 - offset = 20km (any document less than 20km of the center of the query have a score of 1)
 - scale = 130km
 - decay = 0.4 (any document less than 20 + 130 km if the center of the query have a score greater than 0.4)

## Motivation

Our motivation is to have a way of using the proximity filter to boost results at larger scale (like a region or a country) based on the zoom level on our application. Note that this could probably be used to make queries over smaller zones but we won't try to use it in this purpose, at least for now.

### DIFF
--- a/config/bragi-settings.toml
+++ b/config/bragi-settings.toml
@@ -46,6 +46,6 @@ global = 1.0
     [importance_query.proximity]
     weight = 0.4
     weight_fuzzy = 0.4
-    offset_distance = 20.0  # This can be overwritten by the parameter `radius`
-    decay_distance_ratio = 6.5 # The decay distance will be offset_distance * decay_ratio
-    decay = 0.4
+    decay_distance = 130.0 # can be overwritten by query parameter
+    offset_distance = 20.0 # can be overwritten by query parameter
+    decay = 0.4 # can be overwritten by query parameter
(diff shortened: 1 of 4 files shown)


remi-dupre:
@crocme10, I just simplified my PR to allow tweaking the Gaussian used for proximity queries, the parameters already present in `bragi-settings.toml` are kept but only act as defaults now.

PR #418, opened 2020-08-26
Prevent unwanted alias removal

crocme10:
This PR fixes an issue we have in production where updating a coverage may accidentally delete another unrelated coverage.

The name of the coverage is embedded in the alias. If you have a coverage 'foo', then the alias will start with 'munin_foo'.

Now, if you have another coverage 'foo_bar', it will start with 'munin_foo_bar'. When you update 'foo', it will probably delete
'munin_foo' and 'munin_foo_bar'.

Warning (this PR is based on a branch 'update_dependencies')

### DIFF
--- a/libs/mimir/src/rubber.rs
+++ b/libs/mimir/src/rubber.rs
@@ -456,7 +456,7 @@ impl Rubber {
         base_index: &str,
     ) -> Result<BTreeMap<String, Vec<String>>, Error> {
         let res = self
-            .get(&format!("{}*/_aliases", base_index))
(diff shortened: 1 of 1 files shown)


amatissart (approved):
Thanks!  
As you pointed, the problem persists for datasets `foo` and `foo_bar`: the deletion of all indices concerning `foo` will also delete indices about `foo_bar`.
In addition to this change, maybe we would like to prevent  dataset/coverage name from including the character `_` to avoid any accidental removal in this case too ?

PR #419, opened 2020-08-26
Split osm_store, introduce db-storage feature

crocme10:
The file `src/osm_reader/street.rs` contained a large chunk of code
related to storage that was not related to streets per se. This part of
the code was moved to a separate file, `src/osm_reader/osm_store.rs`.

In addition, the dependency on a database storage (Sqlite) was made
optional with the help of a feature 'db-storage'.

This PR is based on the regex_alias branch


src/osm_reader/osm_store.rs
```diff
@@ -0,0 +1,424 @@
...
+        }
+    }
+}
+
+#[cfg(not(feature = "db-storage"))]
+impl ObjWrapper {
+    pub fn new(_db_file: &Option<PathBuf>, _db_buffer_size: usize) -> Result<ObjWrapper, Error> {
+        info!("Running with BTreeMap (RAM) storage");
+        Ok(ObjWrapper::Map(BTreeMap::new()))
+    }
```
amatissart, line 376: Here, if a db_file is passed it will be ignored silently.

amatissart (commented):
Thanks for moving the storage logic to another file, it's definitely cleaner this way :+1:

However I fear that the new `db-storage` feature could break some existing workflows.  
How the `db-storage` feature is supposed to be used with the docker image that is currently built ? As far as I understand, the feature is not be enabled in the `osm2mimir` binary, and the `--db_file` option would be silently ignored.

src/osm_reader/osm_store.rs
crocme10: Good catch!

J'ai mis un warning et une error quand l'utilisateur utilise l'option alors que la feature n'a pas ete utilisee.
J'ai ajoute aussi un commentaire sur les options en ligne de commande, donc le `osm2mimir --help` ressemble a ca:

```
--db-file <db-file>
     If you use this option by providing a filename, then we will use a SQlite database that will be persisted.
     You can only do that if osm2mimir was compiled with the 'db-storage' feature. If you don't provide a value,
     then we will use in memory storage
```

PR #421, opened 2020-09-01
Split storage

crocme10:
Continuation de la PR #419


amatissart:
As discussed, the new feature needs to be enabled in the docker image where importers are built to preserve all existing options:

In `docker/Dockerfile_import`:
```diff
-RUN cargo build --release
+RUN cargo build --release --features db-storage
```

amatissart (approved):

PR #423, opened 2020-09-17
Progressive focus

remi-dupre:
https://github.com/CanalTP/mimirsbrunn/pull/417 introduced a way to apply a very slight proximity boost for a very large region. However it kept the previous behavior of using different weights in a proximity context and out of a proximity context, this makes it hard to benefit from the feature as some proximity-specific behaviors would raise for quite generic requests (eg. decreasing the importance of the weight consequently).

This PR addresses the issue by computing a smooth transitions of the weights used to build ES queries when the radius changes:

 - the new setting `radius_range` gives an interval for this behavior of "transition" base on the value of `curve.offset + curve.scale`
 - if `curve.offset + curve.scale` is above `radius_range`, we use weights from `max_radius_fuzzy` or `max_radius_prefix` like we did before when proximity was disabled
 - if `curve.offset + curve.scale` is below `radius_range`, we use weights from `min_radius` like we did before when proximity was enabled
- if `curve.offset + curve.scale` is inside `radius_range`, we compute weights as a combination of the two previous cases. When the radius increases it converges logarithmically to weights from `max_radius_fuzzy`. The choice of the logarithm is influenced by the observation that cartographic tools typically increases the scale exponentially when the user zooms in.

tt

### DIFF
--- a/libs/bragi/src/query.rs
+++ b/libs/bragi/src/query.rs
@@ -250,30 +250,31 @@ fn build_query<'a>(
         .build();
 
     let weights = {
-        // Weights for minimal level of zoom
-        let min_weights = query_settings.importance_query.weights.min_zoom.weight;
+        let settings = &query_settings.importance_query.weights;
 
-        // Weights for maximal level of zoom
+        // Weights for maximal radius
         let max_weights = match match_type {
-            MatchType::Prefix => query_settings.importance_query.weights.max_zoom.weight,
-            MatchType::Fuzzy => query_settings.importance_query.weights.coords_fuzzy,
+            MatchType::Prefix => settings.max_radius_prefix,
+            MatchType::Fuzzy => settings.max_radius_fuzzy,
         };
+
+        // Weights for minimal radius
+        let min_weights = settings.min_radius;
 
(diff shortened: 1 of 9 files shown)


libs/bragi/src/query.rs
```diff
@@ -249,40 +249,55 @@ fn build_query<'a>(
...
+        &weights,
+        &query_settings.importance_query.weights.types,
+    )];
+
+    if let Some(ref coord) = coord {
+        importance_queries.push(build_proximity_with_boost(
+            coord,
+            &query_settings.importance_query.proximity,
+            match_type == MatchType::Fuzzy,
+        ))
```
amatissart, line 298: I wonder whether this proximity boost, multiplied by the factor `importance_query.proximity.weight` or `importance_query.proximity.weight_fuzzy`, should not be applied gradually too ?

remi-dupre: I can't catch any difference introduced by [this change](https://github.com/CanalTP/mimirsbrunn/pull/423/commits/52bb42271f8c26a49b18eacfb1788c7611d04bf3) but it actually looks like a nice and safe idea

crocme10 (approved):

PR #424, opened 2020-09-30
Streets at borders

remi-dupre:
Currently, when a street is at the border between two admins, it won't be imported into elasticsearch. This is a rather common case as streets often define the border between two cities.

This PR mainly introduces two changes:

- `admin_geofinder` can be spawned with two separate behaviors: `CompareStrict` for search of admins that strictly contain a point and `CompareInclusive` to search for admins that can contain input point only at their border
- when searching for streets, we will create one document for each city it belongs to, this is generalized for any admin that is not parent of another

~*Note*: I added [this PR](https://github.com/georust/geo/pull/514) to `geo` to avoiding having to implement missing geometry operations on our side, we should probably wait for the next release to update this crate (probably update `geojson` too) and merge this PR .~


src/admin_geofinder.rs
```diff
@@ -94,19 +128,19 @@ impl PointDistance for SplitAdmin {
     // the geo algorithms for performance, as suggested in rstar documentation.
     fn contains_point(&self, point: &[f64; 2]) -> bool {
         let p = Point::new(point[0], point[1]);
-        self.boundary.contains(&p)
+        C::compare(&self.boundary, &p)
```
amatissart, line 131: I am not sure this trait `PointDistance` is actually used at all in our use case :upside_down_face: 
All interactions with the RTree use the envelope explicitly (see `PointInEnvelopeSelectionFunction`), so I wonder if this `contains_point` function is even called.

remi-dupre: I'll run a profiler on this X)

amatissart (commented):
A few details to fix or clarify, but it looks good overall.

src/bin/osm2mimir.rs
```diff
@@ -122,12 +122,12 @@ fn run(args: Args) -> Result<(), mimirsbrunn::Error> {
     } else {
         rubber.get_all_admins()?
     };
-    let admins_geofinder = admins.into_iter().collect::<AdminGeoFinder>();
+
```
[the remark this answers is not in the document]
remi-dupre: Right, I managed to forget removing the most important line when I simplified everything :/

src/admin_geofinder.rs
```diff
@@ -103,7 +104,7 @@ impl PointDistance for SplitAdmin {
 // Using an id
 pub struct AdminGeoFinder {
     rtree: RTree<SplitAdmin>,
-    admin_by_id: BTreeMap<String, Arc<Admin>>,
+    admin_by_id: HashMap<String, Arc<Admin>>,
```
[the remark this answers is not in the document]
remi-dupre: Well actually this is mainly because HashMaps are the [recommended default](https://doc.rust-lang.org/std/collections/#use-a-hashmap-when) (mostly for performance reasons as it has a better locality). I was a bit confused to see a BTree and tried to understand if there was a specific motivation behind it.

I agree that this is probably not where our performance hurts and we probably won't notice any consequent improvement. Here is the time spent for a few runs over *ile-de-france*:

 - BTreeMap + BTreeSet (1): 4m35
 - BTreeMap + BTreeSet (2): 4m33
 - BTreeMap + BTreeSet (3): 4m37
 - BTreeMap + HashSet:     4m30
 -  HashMap + HashSet (1):  4m24
 -  HashMap + HashSet  (2):  4m33
 -  HashMap + HashSet (3):  4m33

src/osm_reader/street.rs
```diff
@@ -89,108 +87,129 @@ pub fn streets(
...
-                    coord: get_way_coord(&objs_map, way),
-                    approx_coord: Some(coord.into()),
-                    distance: None,
-                    country_codes,
-                    context: None,
-                })
+                let coord = get_way_coord(&objs_map, &way);
+                let name = rel_name
+                    .or_else(|| way.tags.get("name").map(String::as_str))
+                    .unwrap_or("");
```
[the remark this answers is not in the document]
remi-dupre: Nice catch! I was also able to remove the inconvenient `.map(String::as_str)` :tada:

src/osm_reader/street.rs
```diff
@@ -89,108 +87,129 @@ pub fn streets(
...
+
+                // Add osmid of all the relation members in the set
+                // We don't create any street for all the osmid present in street_rel
+                street_in_relation.insert(ref_obj.member);
```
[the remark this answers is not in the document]
remi-dupre: :+1:
That's a dumb mistake as the comment explicitly described a different behavior than the code it refers to...

src/osm_reader/street.rs
```diff
@@ -89,108 +87,129 @@ pub fn streets(
...
-            context: None,
-        })
-    });
-    street_list.extend(streets);
+    street_list.extend(
+        name_admin_map
+            .values()
+            .filter_map(|min_id| {
+                let obj = objs_map.get(&min_id)?;
+                let way = obj.way()?.clone();
```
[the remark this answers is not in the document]
remi-dupre: Nope, nice catch! (I wonder if we could make a clippy lint out of this)

src/admin_geofinder.rs
```diff
@@ -134,6 +136,61 @@ impl AdminGeoFinder {
...
 
+    // Get all admins overlapping the given coordinates of zone type at most `granularity`.
+    // Each element of the returned array contains a leaf of the admin hierarchy that overlap
+    // input coordinates together with its parents.
+    pub fn get_with_granularity(
+        &self,
+        coord: &geo_types::Coordinate<f64>,
+        granularity: ZoneType,
+    ) -> Vec<Vec<Arc<Admin>>> {
+        // Get a list of overlapping admins...
```
crocme10, line 147: // Get a list of overlapping admins whose zone_type is less than granularity

src/admin_geofinder.rs
```diff
@@ -134,6 +136,61 @@ impl AdminGeoFinder {
...
+            })
+            .filter(|cand| {
+                cand.admin
+                    .zone_type
+                    .map(|zt| zt <= granularity)
+                    .unwrap_or(false)
+            })
+            .collect::<Vec<_>>();
+
+        let mut visited: HashSet<&str> = HashSet::new();
```
crocme10, line 161: Either call it visited_ids, or put a comment to say what it is, thx

src/admin_geofinder.rs
crocme10: and, while we're at it, maybe calling it max_zone_type might be clearer to the reader as to the purpose of the variable...

remi-dupre: It may still be confusing as we return admins with higher `ZoneType` are part of the output. I'll try to go for something like `max_split_zone_type`?

src/osm_reader/street.rs
```diff
@@ -95,126 +94,163 @@ pub fn streets(
...
-        let way_name = rel.tags.get("name");
-        rel.refs
+        let rel = obj.relation().expect("invalid relation filter");
+        let rel_name = rel.tags.get("name");
+
+        // Add osmid of all the relation members in the set.
+        // Then, we won't create any street for the ways that belong to this relation.
+        for ref_obj in &rel.refs {
+            street_in_relation.insert(ref_obj.member);
+        }
```
amatissart, line 153: The previous implementation inserted way members only. Technically only ways with role `street` could be considered.  

I think that we should keep this filter to avoid saving too many ids unnecessarily, especially `house` roles that are probably more abundant than `street` objects in each relation.

src/osm_reader/street.rs
```diff
@@ -95,126 +94,163 @@ pub fn streets(
...
+                approx_coord: Some(coord.into()),
+                distance: None,
+                country_codes,
+                context: None,
+            }
+        };
+
+    // Return an iterator giving documents that will be inserted for a given street: one for each
+    // hierarchy of admins
+    let pois_with_admin = move |name: String, id, kind, mut admins: Vec<Vec<_>>, coord| {
```
amatissart, line 120: This should also be renamed.   
`streets_with_admin` ? or `build_streets_for_admins` ?

remi-dupre: I went for `build_streets_for_admins` which seem perfect :ok_hand:

remi-dupre: I went for `build_streets_for_admins` which seem perfect :ok_hand:

PR #426, opened 2020-10-07
[FEAT]: Add custom settings for osm2mimir

crocme10:
This patch introduces 2 new CLI arguments for osm2mimir:
-  `--config-dir` to point to a directory where configuration are stored
-  `--settings` to identify custom settings.

The motivation is to have different settings depending on the context.

It uses these new settings to expand the list of invalid highways.

It introduces a new dependency, [config](https://docs.rs/config/0.10.1/config/), to read a `.toml` file.
This config crate has possibly many functionalities, like merging configuration files (like specific settings, on top of default ones),
overwriting values with environment variables, ...

### DIFF
--- a/src/osm_reader/street.rs
+++ b/src/osm_reader/street.rs
@@ -67,7 +67,7 @@ pub fn streets(
     let invalid_highways = &settings
         .street
         .clone()
-        .map(|street| street.exclusion.highways.unwrap_or(Vec::new()))
+        .map(|street| street.exclusion.highways.unwrap_or_else(Vec::new))
         .unwrap_or(Vec::new());
 
     let is_valid_highway = |tag: &str| -> bool { !invalid_highways.iter().any(|k| k == tag) };
--- a/src/osm_reader/street.rs
+++ b/src/osm_reader/street.rs
@@ -36,7 +36,7 @@
 use super::osm_utils::get_way_coord;
 use super::OsmPbfReader;
 use crate::admin_geofinder::AdminGeoFinder;
-use crate::{labels, utils, Error};
+use crate::{labels, settings, utils, Error};
 use failure::ResultExt;
 use osmpbfreader::StoreObjs;
 use slog_scope::info;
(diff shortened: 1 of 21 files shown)


src/settings.rs
```diff
@@ -0,0 +1,82 @@
...
+                    "Could not merge default configuration from file {}: {}",
+                    path, e
+                )
+            })?;
+        } else {
+            return Err(failure::err_msg(format!(
+                "Could not read default settings in '{}'",
+                default_path.display()
+            )));
+        }
```
amatissart, line 50: Another possibility would be to read the default file at compile time:
```rust
config.merge(File::from_str(
    include_str!("../config/default.toml"),
    FileFormat::Toml,
))?;
```

Then, `--config-dir` could be made optional, and be required only if `--settings` is used.

crocme10: Excellent suggestion! Implemented in my next commit

src/osm_reader/street.rs
```diff
@@ -62,28 +62,34 @@ pub fn streets(
...
 ) -> Result<StreetsVec, Error> {
-    // This is the list of highway that we don't want to index
-    // See [OSM Key Highway](https://wiki.openstreetmap.org/wiki/Key:highway) for background.
-    const INVALID_HIGHWAY: &[&str] =
-        &["bus_guideway", "escape", "bus_stop", "elevator", "platform"];
+    let invalid_highways = &settings
+        .street
+        .clone()
+        .map(|street| street.exclusion.highways.unwrap_or(Vec::new()))
+        .unwrap_or(Vec::new());
```
remi-dupre, line 71:
```suggestion
        .map(|street| street.exclusion.highways.unwrap_or_else(Vec::new))
        .unwrap_or_else(Vec::new);
```

That's very minor: Clippy will be likely to suggest avoiding unnecessary function calls here

src/settings.rs
```diff
@@ -0,0 +1,110 @@
...
+                            "Could not merge default configuration from file {}: {}",
+                            path, e
+                        )
+                    })?;
+                } else {
+                    return Err(failure::err_msg(format!(
+                        "Could not read default settings in '{}'",
+                        dir.display()
+                    )));
+                }
```
remi-dupre, line 53: Maybe it would be simpler to also use compile-time settings for this case? I suppose you are afraid that it can bring confusion to the user if the actual `default.toml` is exposed to the user but not used?

amatissart: And this would allow to modify the `default.toml` directly, which is probably a good thing if we want it to be exposed.

src/osm_reader/street.rs
crocme10: Good call!

src/settings.rs
crocme10: This configuration thing is a bit of a layered approach... There is a base configuration, and customizing layers on top (and possibly command line argument, and environment variables on top). But they should all be kept dynamic. If you change the directory, then you should read the default.toml in it.

amatissart (approved):

PR #428, opened 2020-10-16
[DEPS]: Update Transit Model

crocme10:
This PR updates transit model to the current version, and introduces the chrono tz crate, because timezone are no longer just `strings` but `structs`, so `chrono_tz` is used to get a `string` representation suitable for mimirsbrunn.


crocme10:
This code leaves introduces an `unwrap()`, so the alternative would be to use an unwrap_or_default, to have an empty comment. I'm not sure what comments are used for...

src/bin/ntfs2mimir.rs
```diff
@@ -116,11 +116,11 @@ fn to_mimir(
...
-    let comments = navitia
-        .comments
-        .iter_from(&stop_area.comment_links)
-        .map(|comment| mimir::Comment {
-            name: comment.name.clone(),
+    let comments = stop_area
+        .comment_links
+        .iter()
+        .map(|com_id| mimir::Comment {
+            name: navitia.comments.get(com_id).unwrap().name.clone(),
```
datanel, line 123: normally `transit_model` ensures that objects do not reference other non-existent objects.
So the `unwrap` is safe here. but who knows?

If you want the code to be safer you can omit the comments that don't exist (and log)
```suggestion
    let comments = stop_area
        .comment_links
        .iter()
        .filter_map(|com_id| navitia.comments.get(com_id))
        .map(|c| mimir::Comment {
            name: c.name.clone(),
```

woshilapin (approved):

src/bin/ntfs2mimir.rs
crocme10: Great. I will do that. Although I'm not sure of the consequences of having empty comments.... But no real programmer should be worried by a lack of comment ;-)

PR #430, opened 2020-10-23
bragi: perform elision in prefix match

remi-dupre:


config/addr_settings.json
```diff
@@ -25,12 +29,24 @@
                     "filter": [ "lowercase", "asciifolding" ],
                     "char_filter" : [ ]
                 },
+                "word_elision": {
+                    "type": "custom",
+                    "tokenizer": "standard",
+                    "filter": [ "elision_filter", "lowercase", "asciifolding" ],
```
amatissart, line 35: Shouldn't the `elision_filter` be applied _after_ `lowercase` ?   
Or is the elision function provided by Elasticsearch not case-sensitive ?

libs/bragi/src/query.rs
```diff
@@ -349,8 +349,12 @@ fn build_query<'a>(
...
         // these are not present in label, for instance, the zip_code.
         // The field full_label contains all of them and will do the trick.
-        MatchType::Prefix => Query::build_match("full_label.prefix".to_string(), q.to_string())
-            .with_operator("and")
+        // The query must at least match with elision activated, matching without elision will
+        // provide extra score bellow.
+        MatchType::Prefix => Query::build_bool()
+            .with_must(vec![Query::build_match("full_label.prefix_elision", q)
+                .with_operator("and")
+                .build()])
```
amatissart, line 357: To be sure, the wrapping `build_bool` has no effect, right ?

config/addr_settings.json
remi-dupre: I think you're right, it doesn't cost much to perform this changed anyway.

As it is quite painful apply the change to all types of documents I delayed and then forgot about it :|

libs/bragi/src/query.rs
```diff
@@ -349,6 +349,8 @@ fn build_query<'a>(
         // When the match type is Prefix, we want to use every possible information even though
         // these are not present in label, for instance, the zip_code.
         // The field full_label contains all of them and will do the trick.
+        // The query must at least match with elision activated, matching without elision will
+        // provide extra score bellow.
         MatchType::Prefix => Query::build_match("full_label.prefix".to_string(), q.to_string())
```
amatissart, line 354:
```suggestion
        MatchType::Prefix => Query::build_match("full_label.prefix_elision".to_string(), q.to_string())
```

config/addr_settings.json
```diff
@@ -129,11 +145,11 @@
                     "index_options": "docs",
                     "analyzer": "word",
                     "fields": {
-                        "prefix": {
+                        "prefix_elision": {
```
amatissart, line 148: When preparing the testing environment for this branch, I realized that renaming this field is actually a breaking change that would require to deploy a new version of Bragi _after_ the new mapping is used. This is impractical as several indices will be updated at different times.

To preserve the compatibility with the previous mapping, it is much simpler to preserve this field name `prefix`. And then no change is necessary in `bragi/`.

remi-dupre: That's indeed not worth the pain, I reverted the change. (Elision is actually probably not important enough to be part of the field name)

amatissart (approved):

PR #431, opened 2020-10-29
[FEAT]: Make use of BulkResult

crocme10:
This patch adds a warning in the logs for each stop which result in an insertion error from Elasticsearch (and for which Elasticsearch returns error information)

### DIFF
--- a/libs/mimir/src/rubber.rs
+++ b/libs/mimir/src/rubber.rs
@@ -690,8 +690,8 @@ impl Rubber {
                                     .unwrap_or_else(|_| String::from("Could not serialize error"));
 
                                 warn!(
-                                    "An error occured while importing stop '{}': {}",
-                                    action_res.inner.id, error
+                                    "An error occured while importing {} '{}': {}",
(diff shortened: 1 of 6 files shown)


libs/mimir/src/rubber.rs
```diff
@@ -671,13 +671,31 @@ impl Rubber {
...
+                        .for_each(|action_res| {
+                            // We only display a warning if it brings some information, otherwise
+                            // the log is distracting
+                            if let Some(ref error) = action_res.inner.error {
+                                let error = serde_json::to_string(error)
+                                    .unwrap_or_else(|_| String::from("Could not serialize error"));
+
+                                warn!(
+                                    "An error occured while importing stop '{}': {}",
+                                    action_res.inner.id, error
```
amatissart, line 694: This error is not specific to `stop` documents. 
```suggestion
                                    "An error occured while importing {} '{}': {}",
                                    T::doc_type(), action_res.inner.id, error
```

PR #433, opened 2020-11-05
config: share "settings" field for index templates

remi-dupre:
Template settings all share the exact same "settings" field, this can be quite painful when altering the [ES schema for all kind of documents](https://github.com/CanalTP/mimirsbrunn/pull/430).

This PR moves the shared field "settings" into its own file. There is still a lot of redundancies for "mappings" but it might be better to tackle the issue in a separate PR as it would be more tricky.


remi-dupre:
Deprecated code

PR #434, opened 2020-11-06
street: skip added ways in relations

remi-dupre:
When computing streets, always skip ways that are already part of another relation. The previous behavior was not to check for it but this can lead to some duplicates when two relations share the exact same ways:

 - https://www.openstreetmap.org/relation/11824084
 - https://www.openstreetmap.org/relation/4251394


SGrenet:
Is there a control on the relation name?
An interesting exemple : 
https://www.openstreetmap.org/relation/2891001
https://www.openstreetmap.org/relation/3363458

A use case that we need to keep : 
https://www.openstreetmap.org/relation/4429947
https://www.openstreetmap.org/relation/3554704
There are hundreds of duplicated street names in Lille (lvl10 Lille, Lille-Hellemme, Lille-Lomme), sometimes 
- in 2 differents relations
- 1-2 relations and ways (in case of 3 distinct streets)
- 1 relation and ways (case 2 streets)
Unavoidable defect if only ways: we group them

amatissart:
So the best solution would be to skip ways from a street relation that are already part of another relation, only if the name of the second relation is the same as the first one. Does it sound right ?

A few implementation details that come to mind: 
 * the name comparison should probably be case-insensitive. 
 * storing the name with each way id could be memory expensive and redundant. Some compression mechanism could be useful, or simply using the relation id as an indirection.

remi-dupre:
>     * storing the name with each way id could be memory expensive and redundant. Some compression mechanism could be useful, or simply using the relation id as an indirection.

As I understand it, `street_in_relation` would now contain the list of names associated with a way id, it may not be much more expensive as names would not have to be cloned ? (The most expensive part would eventually be Vec, which is the size of 3 pointers, maybe something like [tinyvec](https://crates.io/crates/tinyvec) would be helpful)

remi-dupre:
Deprecated code

PR #435, opened 2020-11-16
Update docker base to buster

crocme10:
This PR proposes to update the base for building docker images from *stretch* to *buster*.
I recently used the binaries from `Dockerfile_import` inside another docker image which was based on docker,
and it failed because our Docker image was stretch based.

Now I actually think we should have both leaving side by side. If you have an older Debian you should still be able to use Mimirsbrunn. It seems a way to do that would be to have two images, ie `Dockerfile_stretch_import` and `Dockerfile_buster_import`. What do you think?

### DIFF
--- a/docker/Dockerfile_bragi
+++ b/docker/Dockerfile_bragi
@@ -1,10 +1,10 @@
-FROM rust:1.44.1-stretch as builder
+FROM rust:1.47-buster as builder
 
 WORKDIR /srv/mimirsbrunn
 
 ENV DEBIAN_FRONTEND noninteractive
 RUN apt-get update \
-    && apt-get install -y make libssl1.0-dev git \
+    && apt-get install -y make libssl-dev git \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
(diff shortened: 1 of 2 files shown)


amatissart (approved):
Also note that the official [`rust` images based on stretch](https://hub.docker.com/_/rust?tab=tags&page=1&name=stretch) are no longer updated. (Last updated 3 months ago, based on Rust 1.45.2). So using buster as the base image is probably the way to go.

I see no objection to maintain images based on stretch, but is there really a need for it ?  
And if there are other workflows that cannot use the docker image directly, I would tend to think that the binaries should be packaged differently for them.

crocme10:
For example, if we look at (part of) the tags maintained by rust, we have the following:
1-buster, 1.47-buster, 1.47.0-buster, buster, 1, 1.47, 1.47.0, latest
1-slim-buster, 1.47-slim-buster, 1.47.0-slim-buster, slim-buster, 1-slim, 1.47-slim, 1.47.0-slim, slim

So I propose to do something a bit similar:
v1.17, v1.17-buster, v1.17-stretch
v1.17.0, v1.17.0-buster, v1.17.0-stretch

Is that OK (knowing, that at some point, we'll drop completely drop stretch, and include a new version)?

PR #436, opened 2020-11-17
osm2mimir configuration

crocme10:
Moving poi type configuration inside the (big) configuration file. Removing the command line argument which enabled to read a separate file.

So I moved the internal default configuration to a separate file, because it's not a good idea to have configuration inside the code.
and I used the `include_str!` macro.

I used a `toml` format, which is less readable than json, but it does allow comments, which i think is really important for configuration files.


src/settings/osm2mimir.rs
```diff
@@ -0,0 +1,380 @@
...
+                    db_file.to_str().expect("valid utf-8 filename"),
+                )
+                .with_context(|e| format!("Could not set database file: {}", e))?;
+        }
+
+        if let Some(db_buffer_size) = args.db_buffer_size {
+            config
+                .set("database.buffer_size", i64::try_from(db_buffer_size)?)
+                .with_context(|e| format!("Could not set database buffer size: {}", e))?;
+        }
```
amatissart, line 295: Passing only one of `--db-buffer-size` or `--db-file` without the other will fail with a not very explicit message:
```
ERRO Could not generate settings from configuration: missing field `file`
```

Also, no default value for the buffer size is defined anymore.
Maybe only the database file should be defined as an Option, instead of using `Option<Database>` in the main struct.

src/settings/osm2mimir.rs
```diff
@@ -0,0 +1,380 @@
...
+                .set("admin.import", args.import_admin)
+                .with_context(|e| format!("Could not set import_admin: {}", e))?;
+        }
+
+        if let Some(city_level) = args.city_level {
+            config
+                .set("admin.city_level", i64::try_from(city_level)?)
+                .with_context(|e| format!("Could not set city level: {}", e))?;
+        }
+
```
amatissart, line 180: `args.level` is not read here to update `admin.levels`. As far as I understand, that's a reason why tests are currently failing.  

I wonder if it could be cleaner to mutate the `Settings` instance directly after it has been deserialized, instead of using `config.set` here, as it requires unnecessary conversions, and does not validate keys at compile time.

crocme10: Implementing the trait [Source](https://docs.rs/config/0.10.1/config/trait.Source.html) for Args, so that I can call in something like config.merge(args)

src/settings/osm2mimir.rs
crocme10: Yes, I noticed that too, I propose to do the following:
1. Maybe do a better job at handling the feature db-storage: The use should not even have the option to use --db-file and such if the feature hasn't been compiled in.
2. prior to using arguments, if db-storage is used, make sure both options db-file and db-buffer-size are set. Propose a useful error message for the missing db-buffer-size value (how do you choose a value?)

crocme10: I think leaving a default value for db-buffer-size is not such a good idea, because that value depends a lot on the context (the size of the osm file you're working with). I'd rather give suggestions, in the form of comments in the config file (yeah, \0/ toml allows comments!) and in the command line help and error messages.

amatissart: All that sounds reasonable enough to me :+1:

amatissart (commented):
Very minor comments.

config/osm2mimir-default.toml
```diff
@@ -0,0 +1,134 @@
...
+  pois_replicas = 1
+
+[admin]
+  import = false
+  city_level = 8
+  levels = []
+
+[way]
+  import = false
+  [way.exclusion]
```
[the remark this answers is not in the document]
crocme10: I agree that street feels more natural... I changed it because I saw it was called 'way' in some places. I guess the naming is inconsistent across. I moved it back to street for the PR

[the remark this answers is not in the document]
crocme10: (par exemple, on utilise `--import-way` dans la ligne de commande...

PR #438, opened 2020-11-20
allow for a match of POIs by their address

remi-dupre:
Index the address related to a POI, this behavior is still quite restrictive to avoid regressions.

For example in our use case in Qwant Maps:

Screenshot_2020-11-20_09-35-14


remi-dupre:
Deprecated code

PR #439, opened 2020-11-20
[FEAT]: Add Makefile and parametric dockerfiles

crocme10:
To support multiple platforms (like debian buster and stretch), we
introduce build arguments (`--build-arg`) when calling Dockerfile, such
that one Dockerfile can be used to generate one image per debian
platform.

In addition we introduce a Makefile and support script to facilitate
releases.

Building a docker image, you now have to use a line like:

docker build --build-arg DEBIAN_VERSION=buster --build-arg RUST_VERSION=1.47 ...

Plus de details sur le Makefile [ici](https://github.com/crocme10/test-makefile)

### DIFF
--- a/.make-release-support
+++ b/.make-release-support
@@ -22,19 +22,19 @@ readonly SCRIPT_NAME=$(basename "$0")
 # Uses git status in short format to see if there are outstanding files that
 # have not been commited
 function hasChanges() {
-	test -n "$(git status -s .)"
+  test -n "$(git status -s .)"
 }
 
 # This returns the version contained in Cargo.tom l
 function getRelease() {
-	# awk -F= '/^release=/{print $2}' .release
+  # awk -F= '/^release=/{print $2}' .release
   cat Cargo.toml | grep '^version' | grep '\([0-9]\+\.\?\)\{3\}' -o
 }
 
 # The tag can have suffixes, like the commit for when the directory is not equal to
 # the tagged content.... This function return the base part of the tag
 function getBaseTag() {
-	# sed -n -e "s/^tag=\(.*\)$(getRelease)\$/\1/p" .release
+  # sed -n -e "s/^tag=\(.*\)$(getRelease)\$/\1/p" .release
   getRelease
 }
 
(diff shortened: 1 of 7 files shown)


crocme10:
There is already something that's not going to work, is that we have multiple dockerfiles. And the Makefile only handles one in the `deploy.env` file. I should turn the variable `DOCKER_FILE_PATH` to an array, and iterate.

amatissart:
This principle looks good to me.  
Have you considered how it interacts with the builds on Docker Cloud though ? As far as I know this repository is currently configured by Kisio to build docker images automatically, and push them to https://hub.docker.com/u/navitia. Do you plan to update this configuration to build these images for both Buster and Stretch ?

crocme10:
I haven't upgraded the script which deploys the docker images to docker hub. I will indeed have to do that. I don't think you will have to do anything with [docker_mimir](https://github.com/QwantResearch/docker_mimir).

.make-release-support
```diff
@@ -0,0 +1,190 @@
...
+# Uses git status in short format to see if there are outstanding files that
+# have not been commited
+function hasChanges() {
+	test -n "$(git status -s .)"
+}
+
+# This returns the version contained in Cargo.tom l
+function getRelease() {
+	# awk -F= '/^release=/{print $2}' .release
+  cat Cargo.toml | grep '^version' | grep '\([0-9]\+\.\?\)\{3\}' -o
```
remi-dupre, line 31:
```suggestion
	# awk -F= '/^release=/{print $2}' .release
	cat Cargo.toml | grep '^version' | grep '\([0-9]\+\.\?\)\{3\}' -o
```

There are some tab/space indent inconsistencies throughout this file

remi-dupre (commented):
That sounds nice :)

I think we would need to update the build documentation [in the README](https://github.com/CanalTP/mimirsbrunn#build)? The new preferred way would be using `make build` as I understand?

crocme10:
While we're at building parametric docker images, should we add a build argument for the `db-feature` ?

amatissart:
> While we're at building parametric docker images, should we add a build argument for the `db-feature` ?

Just to spare a few megabytes from images that are very small already ? I don't see a need for it.  
But if you think it's worthwhile and easy enough, go for it :smiley:

crocme10:
I don't understand your comment. If someone wants to use a docker image, today, oh, yes, we do enable that feature. So its enabled by default, and they can use. Ok, forget it then ;-)

remi-dupre:
@crocme10, this is completely out of scope : @amatissart and I where kicked from you slack server, we are not sure why but it happened a few hours after it asked us to setup a double authentication (which we both did).

Now, when we try to connect we receive this email:

>Vous avez récemment essayé de vous connecter à Kisio Digital, mais il semble que votre compte pour cet espace de travail soit désactivé. Nous en sommes désolés !
>
> Si vous pensez qu’il s’agit d’une erreur, veuillez demander à un administrateur de l’espace de travail de réactiver le compte.

Do you have a clue on what happens?

crocme10:
yes, we have new security measures. I will find out how you can log back in.

.make-release-support
crocme10: done!

PR #443, opened 2020-12-02
Build deb package in gihub actions

pbench:
so that packages are easier to download than from the CI

### DIFF
--- a/.github/workflows/build_package.yml
+++ b/.github/workflows/build_package.yml
@@ -1,6 +1,8 @@
 name: Build packages 
 
-on: push
+on: 
+  push:
+    branches: [master]
 
 
 jobs:
--- a/.github/workflows/build_package.yml
+++ b/.github/workflows/build_package.yml
@@ -29,7 +29,7 @@ jobs:
         . $HOME/.cargo/env
         cargo deb
     - name: upload debian packages
-      uses: actions/upload-artifact@v1
+      uses: actions/upload-artifact@v2
       with:
-        name: archive.zip
-        path: "target/debian/*.deb"
+        name: archive
+        path: target/debian/*.deb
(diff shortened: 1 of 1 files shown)


crocme10:
Is this debian able to handle the `db-storage` feature (support sqlite)?

.github/workflows/build_package.yml
```diff
@@ -0,0 +1,37 @@
...
+        apt update
+        apt install -y curl
+        curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs > get_rust.sh
+        sh get_rust.sh -y
+        . $HOME/.cargo/env
+        rustup update
+    - name: install packaging tool
+      run: |
+        . $HOME/.cargo/env
+        cargo install --force cargo-deb
```
nonifier, line 28: Maybe in the future, it would be nice to move all the tool-chain dependencies inside a container, the same way `navitia/debian8_dev` does. It would save us time and compute, and prevent any breaking change from a new version of rust-compiler. 
But, I think it's fine for now as long as we can move forward with the CI :)

nonifier (approved):

.github/workflows/build_package.yml
crocme10: For this PR, I'd like

- to support the db-storage feature (we're not using it at kisio, but someone might)
  i don't know if the packages have to be installed (eg `apt install sqlite3`)
  that means changing the build stage to `cargo build --all-features --release`
- to have a fixed version of rust (not just `rustup update`, but also rustup default 1.47 for example)

crocme10: In https://github.com/CanalTP/mimirsbrunn/blob/master/docker/Dockerfile_bragi this file, to build a docker image of bragi, we support multiple debian versions and rust versions. I don't know if we need multiple debian packages, one for buster, one for stretch

PR #446, opened 2020-12-20
[Changed] Makefile for docker images

crocme10:
This makes a few changes to the Makefile and its supporting script.
It moves the Dockerfile for mimir and import in different folders so
that they can be easily targeted by the Makefile.


amatissart:
It seems that the new Dockerfiles have not been committed.

PR #447, opened 2020-12-20
[Changed] github workflow files

crocme10:
This contains now two parallel workflows, one to generate
debian packages, and the other to do validity checks.

It also changes the script for the production of debian packages:
- The script no longer depends on fpm, which does not seem to add much value
- The workflow uses a matrix and can produce packages for different versions of debian.


crocme10:
Sorry abouth the validation / check failure, which is due to a missing `make check` command, made available in another PR.

PR #449, opened 2020-12-23
Platform

crocme10:
This PR does two little things:
1. It extends street exclusion to public_transport { platform, hub }
2. It changes the key for highways to highway (to match what's done in OSM)

### DIFF
--- a/config/osm2mimir-default.toml
+++ b/config/osm2mimir-default.toml
@@ -23,13 +23,8 @@ dataset = "fr"
   import = false
   [street.exclusion]
     # See [OSM Key Highway](https://wiki.openstreetmap.org/wiki/Key:highway) for background.
-    highways = [
-      "bus_guideway",
-      "bus_stop",
-      "elevator",
-      "escape",
-      "platform"
-    ]
+    highway = [ "bus_guideway", "bus_stop", "elevator", "escape", "platform" ]
+    public_transport = [ "platform", "hub" ]
 
 [poi]
   import = false
--- a/src/osm_reader/street.rs
+++ b/src/osm_reader/street.rs
@@ -63,20 +63,35 @@ pub fn streets(
     let invalid_highways = settings
         .street
         .as_ref()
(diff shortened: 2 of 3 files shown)


amatissart:
As far as I can see, the new condition on "public_transport" tag would include train stations in the `street` index.
Is it really what is expected here ?  
https://overpass-turbo.eu/s/12jP

SGrenet:
> 
> 
> As far as I can see, the new condition on "public_transport" tag would include train stations in the `street` index.
> Is it really what is expected here ?
> https://overpass-turbo.eu/s/12jP

I don't think, none of them are highways
https://overpass-turbo.eu/s/12vA

We seek to avoid import of highway=pedestrian&public_transport=platform

BTW, I haven't seen any of these cases (public_transport=station) in previous indexes

crocme10:
There are 2 things to lookup in this PR:
1. Does the rust code correctly interpret the configuration file
2. The configuration file itself.

The configuration can be adjusted later at runtime, it should just provide sensible / minimal defaults.

amatissart:
> > As far as I can see, the new condition on "public_transport" tag would include train stations in the `street` index.
> > Is it really what is expected here ?
> > https://overpass-turbo.eu/s/12jP
> 
> I don't think, none of them are highways
> https://overpass-turbo.eu/s/12vA

That's my point. Unless I miss something, the new condition allows a way with no `highway` tag to be imported.
```rust
has_valid_name_tag && (has_valid_highway_tag || has_valid_public_transport_tag)
```
A way with a `name` and `public_transport=station` will pass this condition, even if no `highway` tag is present.

> We seek to avoid import of highway=pedestrian&public_transport=platform

Then I think that `has_valid_public_transport_tag` should be rewritten to allow elements with no `public_transport` tag.
Maybe something like:
```rust
                let has_no_excluded_public_transport_tag = way
                    .tags
                    .get("public_transport")
                    .map_or(true, is_valid_public_transport(v));
                has_valid_name_tag && has_valid_highway_tag && has_no_excluded_public_transport_tag
```

PR #453, opened 2021-01-13
[Changed] Add negative flags

crocme10:
We introduce 3 command line argument to let the user specify that
he/she does not want to import a certain kind of place. Prior to this,
the absence of the flag was meant to signify the user did not want to
import a certain kind of place, which interfered with the configuration.


src/settings/osm2mimir.rs
```diff
@@ -189,13 +189,22 @@ pub struct Args {
     connection_string: Option<String>,
     /// Import ways.
     #[structopt(short = "w", long = "import-way")]
-    import_way: bool,
+    import_way: Option<bool>,
+    /// Don't import ways.
+    #[structopt(long = "no-import-way")]
+    no_import_way: Option<bool>,
```
amatissart, line 195: Using `Option<bool>` causes current tests to break.  
As far as I understand, in this case `--import-way` cannot longer be used as a flag without value.  
So to keep the current behavior, we would need to define both flags as `bool` (without Option). Then, they will be `false` by default when they are absent from the command line, and we would update the configuration only if one of them is `true`.

But as you suggested, it could actually make sense to keep a single `Option<bool>`, that would require an explicit "true" or "false" value, to override the current configuration. 
That would be a breaking change, but it may be less confusing.

crocme10: Thanks @amatissart for noticing this.

It seems that the most logical solution is to remove this `--no-import-way`, and keep the `--import-way` as an `Option<bool>`. That way, we have the following:

| input                           |                             result |      wrt config                  |
|------------------------------|--------------------------------|---------------------------------|
| `--import-way true`     | `import-way: Some(true)` | Overrides configuration |
| `--import-way false`   | `import-way: Some(false)` | Overrides configuration |
| nothing                      | `import-way: None`            | Configuration decides   |
| `--import-way`           | Error  missing value           |                                       |

Are we good with this behavior?

amatissart: This approach looks good to me :+1:

PR #458, opened 2021-02-04
Fix id poi format

azime:
**Why this PR** : To standardize the POI identifiers accross tools at Kisio


SGrenet (approved):

crocme10:
Thanks @azime , 
Would you mind providing more details on the reason why the POI id needs fixing.
Also, I will remove the reference to the JIRA COR, which is not accessible from outside of Kisio. We can, however, refer to this Pull Request in the ticket.

amatissart (approved):

PR #459, opened 2021-02-05
[Added] Multiple docker targets

crocme10:
Slight extension to the Makefile to support building multiple images. The image name are the name of the directories found directly under the docker directory. This change iterates through all the directories under the docker directory, and build an image for each of them.
This change is to support building images for both mimir binaries, and bragi.


azime:
when i run : `make snapshot`

image

PR #460, opened 2021-02-08
Multiple docker repo

crocme10:
This is a significant change in the Makefile. It should now support:

* Multiple docker images: Each image is a subdirectory under the docker directory at the root.
* Multiple docker repositories: One for each of the three main target: snapshot, prerelease, and release. These repositories are configured in the deploy.env file, with corresponding variable names: SNAPSHOT_REPO, RELEASE_REPO, ...

### DIFF
--- a/.make-release-support
+++ b/.make-release-support
@@ -38,12 +38,6 @@ function getBaseTag() {
   getRelease
 }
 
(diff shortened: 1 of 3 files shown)


crocme10:
Obviously missing, the documentation that supports the use of this great Makefile ;-)

PR #466, opened 2021-03-09
[Changed] Add request_id to autocomplete param

crocme10:
This PR introduces a new `request-id` parameter in the autocomplete route, which is used in logging so that we can easily trace client's requests.

### DIFF
--- a/libs/bragi/src/query.rs
+++ b/libs/bragi/src/query.rs
@@ -622,9 +622,7 @@ pub fn autocomplete(
     query_settings: &QuerySettings,
     request_id: Option<&str>,
 ) -> Result<Vec<mimir::Place>, BragiError> {
-    info!("Info id: {:?}", request_id);
-    debug!("Debug id: {:?}", request_id);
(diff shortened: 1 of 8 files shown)


libs/bragi/src/query.rs
```diff
@@ -600,7 +620,11 @@ pub fn autocomplete(
     mut rubber: Rubber,
     debug: bool,
     query_settings: &QuerySettings,
+    request_id: Option<&str>,
 ) -> Result<Vec<mimir::Place>, BragiError> {
+    info!("Info id: {:?}", request_id);
+    debug!("Debug id: {:?}", request_id);
+    trace!("Trace id: {:?}", request_id);
```
azime, line 627: Why info, debug and trace for the same

benoit-bst (approved):

libs/bragi/src/query.rs
amatissart: and is that not redundant with the logs added in `query` function ?

crocme10:
complete git mess

PR #467, opened 2021-03-15
[CI] Fix release workflow

azime:


.github/workflows/release.yml
```diff
@@ -57,7 +58,10 @@ jobs:
     needs: package
     steps:
       - name: install  dependency
-        run: apt update && apt install -y httpie
+        run: |
+          apt-get update
+        apt update && apt install -y httpie
```
pbench, line 63: I guess that just one of `apt-get update` and `apt update` is enough

pbench (approved):

.github/workflows/release.yml
benoit-bst: Yep there are a double update

benoit-bst (approved):

.github/workflows/release.yml
azime: :+1:

benoit-bst (approved):

crocme10 (approved):

PR #469, opened 2021-03-26
[CI] Use sudo

azime:


benoit-bst (commented):
It is mandatory ? Why a sudo is necessary ?

azime:
See: [CI](https://github.com/CanalTP/mimirsbrunn/runs/2200119320?check_suite_focus=true)

benoit-bst:
Okay, but I don't understand why. 
We code GA flow, from the beginning, without sudo. We can see examples inside Navitia project.
I don't see the difference between this 2 projects and why Sudo is necessary for imirsbrunn

azime:
In Navitia, we use containers

benoit-bst (approved):

pbench (approved):

PR #471, opened 2021-04-08
request-id

crocme10:
This PR fixes some lints (clippy), and adds an optional `request_id` parameter, which can be used to trace requests. This probably is a moot issue if we overhaul tracing in a distributed environment with more efficient tools (eg jaeger, ....)

### DIFF
--- a/libs/bragi/src/query.rs
+++ b/libs/bragi/src/query.rs
@@ -622,9 +622,7 @@ pub fn autocomplete(
     query_settings: &QuerySettings,
     request_id: Option<&str>,
 ) -> Result<Vec<mimir::Place>, BragiError> {
-    info!("Info id: {:?}", request_id);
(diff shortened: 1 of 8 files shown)


azime (approved):

libs/bragi/src/query.rs
```diff
@@ -600,7 +620,9 @@ pub fn autocomplete(
     mut rubber: Rubber,
     debug: bool,
     query_settings: &QuerySettings,
+    request_id: Option<&str>,
 ) -> Result<Vec<mimir::Place>, BragiError> {
+    info!("Request Id: {:?}", request_id);
```
amatissart, line 625: Is that not redundant with the other message in `call_autocomplete` ?  
And is it intentional to log even if request_id is None ?

crocme10: sure, removing it

PR #473, opened 2021-04-08
Shape

crocme10:
This PR adds a query parameter 'shape_scope' to define the domain of application of the shape parameter. The shape parameter defines a finite domain and the shape_scope defines the list of places to which the shape is applied.


tests/bragi_three_cities_test.rs
```diff
@@ -157,7 +157,7 @@ fn three_cities_shape_test(bragi: &mut BragiHandler) {
         "coordinates": [[[2.656546, 48.537227],
         [2.657608, 48.537244],[2.657340, 48.536602],[2.656476, 48.536545],[2.656546, 48.537227]]]}}}"#;
 
-    let geocodings = bragi.post("/autocomplete?q=Rue du Port", shape);
+    let geocodings = bragi.post("/autocomplete?q=Rue du Port&shape_scope[]=street", shape);
```
azime, line 160: Is it necessary to add this argument?

tests/bragi_three_cities_test.rs
```diff
@@ -189,7 +189,10 @@ fn three_cities_shape_test(bragi: &mut BragiHandler) {
         "coordinates":[[[2.656546, 48.538927]
         ,[2.670816, 48.538927],[2.676476, 48.546545],[2.656546, 48.546545],[2.656546, 48.538927]]]}}}"#;
 
-    let geocodings = bragi.post("/autocomplete?q=Melun", shape);
+    let geocodings = bragi.post(
+        "/autocomplete?q=Melun&shape_scope[]=admin&shape_scope[]=street",
+        shape,
+    );
```
azime, line 195: Is it necessary to add this argument?

azime (commented):
What does that do? 
`
bragi.post("/autocomplete?q=Rue du Port&shape_scope[]=blabla", shape);
`

tests/bragi_three_cities_test.rs
crocme10: yes

tests/bragi_three_cities_test.rs
crocme10: yes

crocme10:
With an unrecognized shape_scope parameter, eg `shape_scope[]=foo`, there is no validation. So the shape filter is applied to places of type `foo`, and there are none..... So it's as if the shape and shape_scope parameter were ignored

libs/bragi/src/routes/autocomplete.rs
```diff
@@ -88,6 +88,10 @@ pub struct Params {
     #[serde(default, rename = "poi_type")]
     poi_types: Vec<PoiType>,
     lang: Option<String>,
+    // The scope is a list of place types on which we apply the shape filter.
+    // Places found in this list are restricted to the shape.
+    #[serde(default)]
+    shape_scope: Vec<String>,
```
antoine-de, line 94: shouldn't this be a `PlaceType`? (can't remember the right type)

crocme10: There doesn't seem to be an easy way to do that:
I am using the term `_type` to match the document. The value for this `_type` are the values given by the trait `mimir::MimirObject::doc_type()`.
What you were probably thinking about was the `bragi::routes::autocomplete::Type`, which is an enum serialized to `city`, `house`, ... but that's of no use to me.

antoine-de: hum maybe you need to add a new enum for it? Since it can be a fixed list, I think rust's enums are a good fit for this, the possible values will be more explicit

crocme10: I've implemented your suggestion.

azime (approved):

libs/bragi/src/query.rs
```diff
@@ -177,6 +177,7 @@ fn build_query<'a>(
     match_type: MatchType,
     coord: Option<Coord>,
     shape: Option<Geometry>,
+    shape_scope: &[&str],
```
antoine-de, line 180: shouldn't you also type this and convert it as late as possible into a string?

tests/bragi_three_cities_test.rs
antoine-de, line 160: maybe it can be worth it to add a test with `shape_scope[]=unknown_doc_type`?

tests/bragi_bano_test.rs
```diff
@@ -182,7 +231,14 @@ fn simple_bano_shape_filter_test(bragi: &mut BragiHandler) {
...
         )
-    )
+    );
+
+    // Searching with an invalid shape_scope parameter... should yield an error.
+    let (status, _) = bragi.raw_post(
+        "/autocomplete?q=18 Rue Hector Malot, (Paris)&shape_scope[]=add",
+        shape,
+    );
+    assert_eq!(status, actix_web::http::StatusCode::BAD_REQUEST);
```
antoine-de, line 241: by curiosity what error do you get on this query?

crocme10: {"short":"validation error","long":"invalid argument: failed with reason: unknown variant `foo`, expected one of `admin`, `street`, `addr`, `poi`, `stop`"}

libs/bragi/src/query.rs
crocme10: done

tests/bragi_bano_test.rs
antoine-de: It's a bid of a shame not to know with parameter is wrong, but if it's difficult to do, that's fine for me

antoine-de (approved):

PR #483, opened 2021-07-13
Fix CI workflow

crocme10:
This version fixes the previous version which was not overriding the toolchain version.
It does not call the makefile, but rather relies solely on the github actions.


crocme10:
The errors in this PR checks will be fixed with #482

PR #485, opened 2021-08-16
Adapt cosmogony2mimir_test with new docker wrapper and es client

amatissart:
To run this test:  
```
ELASTICSEARCH_TEST_URL=http://localhost:9201 cargo test --test cosmogony2mimir_test
```


amatissart:
Replaced with #496

PR #496, opened 2021-08-20
Additional tests about documents indexed by cosmogony2mimir

### LINKED ISSUE #487: Split and remove redundencies in cosmogony2mimir_test (integration test)

### LINKED ISSUE #488: Use `search_documents` in cosmogony2mimir_test

amatissart:
Closes #487
Closes #488


Cargo.toml
```diff
@@ -72,7 +72,7 @@ tracing = "0.1.26"
 tracing-appender = "0.1.2"
 tracing-futures = "0.2.5"
 tracing-subscriber = "0.2.17"
-tracing-bunyan-formatter = "0.1.7"
+tracing-bunyan-formatter = { version = "0.2", default-features = false }
```
amatissart, line 75: To work around https://github.com/serde-rs/json/issues/768

src/bin/cosmogony2mimir.rs
```diff
@@ -161,17 +164,23 @@ mod tests {
...
-        let guard = mtx.lock().unwrap();
+    fn get_es_test_url() -> String {
+        std::env::var(elasticsearch::remote::ES_TEST_KEY).expect("env var")
+    }
+
+    async fn es_initialize() -> std::sync::MutexGuard<'static, ()> {
+        let guard = docker::AVAILABLE.lock().unwrap();
         docker::initialize().await.expect("initialization");
+        guard
+    }
```
amatissart, line 179: Running the tests sequentially would be preferable.  
Maybe with https://github.com/palfrey/serial_test/

crocme10: We're effectively running the test sequentially, each test needs to hold a mutex to run. I see that you've removed the `(drop guard)` lines. But I think they are actually needed. If a thread panics while holding the mutex, it 'poisons' the mutex, and then each thread trying to acquire it will result in an error:
thread 'tests::should_return_an_error_when_given_an_invalid_path_for_input' panicked at 'called `Result::unwrap()` on an `Err` value: PoisonError { .. }', src/bin/cosmogony2mimir.rs:176:46
When you remove the drop, then we're in a situation where we test something using an assert while still holding the mutex.
So I'll merge you're PR, and put the drops back in. We can revisit this decision later. 
We can be looking at the crate serial_test, although i'd like to balance the cost of having a new dependency vs having a mutex around.

PR #498, opened 2021-08-26
Simply usecases

remi-dupre:
The main goal of this PR is to remove the trait `UseCase` to just generic function calls. As discussed before in Slack, it requires less context to use and understand it. Using functions generics we keep the same expressiveness (and even a bit more since some things can't be done with traits yet).

I also replaced some uses of `dyn` with generic calls over `Stream`, this allowed to remove the `ErasedStorage` trait which I find confusing as very redundant with its twin `Storage`.


libs/mimir2/src/domain/ports/primary/explain_query.rs
```diff
@@ -0,0 +1,20 @@
+use crate::domain::model::query::Query;
+use crate::domain::ports::secondary::explain::{Error, Explain, Parameters};
+
+pub async fn explain_document<B, D>(
```
crocme10, line 4: I think we should keep ports as traits

libs/mimir2/src/domain/ports/mod.rs
```diff
@@ -1,7 +1,2 @@
-pub mod explain;
-pub mod import;
-pub mod list;
-pub mod query;
-pub mod remote;
-pub mod search;
-pub mod storage;
+pub mod primary;
```
crocme10, line 1: That's a good reorganization. It makes the code a bit move verbose in the import section, but much easier to discover for
a newcomer

crocme10:
If we can get away with the simpler version you suggest lets do it! I'd like, however, to keep a few things:

# Ports

For one, I think we should keep ports as traits. When you have a trait, you can have several implementations, so you can have test doubles

# Usecase or Service or ...?

We need something for what used to be called usecase prior to your PR.
Different sources of information use different names for it. In [Structuring Code for Better Testability](https://betterprogramming.pub/structuring-rust-project-for-testability-18207b5d0243), there is a primary port `AuthService`. Then he describes, _in the domain_, a structure `AuthServiceImpl` which implements the trait `AuthService`. This implementation is directly used in the [main.rs](https://github.com/eckyputrady/testable-rust-web-project/blob/main/src/main.rs).

```rust
    let service = AuthServiceImpl {
        credential_repo: PostgresCredentialRepoImpl {
            pg_pool: pg_pool.clone()
        },
        token_repo: RedisTokenRepoImpl {
            redis_client: redis_client.clone()
        }
    };
    rest_auth_controller::configure(web::Data::new(service), cfg);
```

So what do you have in this code? A port implementation which is given to a primary adapter (rest_auth_controller)

This is mentioned differently by [Herberto Graca](https://herbertograca.com/2017/11/16/explicit-architecture-01-ddd-hexagonal-onion-clean-cqrs-how-i-put-it-all-together/#connecting-the-tools-and-the-application-core) when he writes:

> In other words, our Driving Adapters are Controllers or Console Commands who are injected in their constructor with some object whose class implements the interface (Port) that the controller or console command requires.

PR #500, opened 2021-09-02
Fix docker connectivity

crocme10:
Try to detect early if there is a connection problem when starting an elasticsearch docker. This improves the testing situation since we won't get the mutex guard if elasticsearch is not ready.

### DIFF
--- a/libs/mimir2/src/adapters/secondary/elasticsearch/mod.rs
+++ b/libs/mimir2/src/adapters/secondary/elasticsearch/mod.rs
@@ -8,7 +8,7 @@ pub mod remote;
 pub mod storage;
 
 #[derive(Clone)]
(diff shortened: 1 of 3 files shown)


libs/mimir2/src/adapters/secondary/elasticsearch/mod.rs
```diff
@@ -8,7 +8,7 @@ pub mod remote;
 pub mod storage;
 
 #[derive(Clone)]
-pub struct ElasticsearchStorage(Elasticsearch);
+pub struct ElasticsearchStorage(pub Elasticsearch);
```
remi-dupre, line 11: Couldn't we manage to only allow access from `mimir2`?

```suggestion
pub struct ElasticsearchStorage(pub(crate) Elasticsearch);
```

PR #505, opened 2021-09-03
Move `Document`/`ContainerDocument` in a new lib `common`

remi-dupre:
Noticeable changes:
 - moving those traits in a new module allow to implement it on structures from `places` which allows to avoid the use of odd wrappers like `AdminDoc`
 - the "legacy" trait `MimirObject` has been removed and total replaced by `Document` / `ContainerDocument`
 - the spawned elasticsearch container is now constrained to 1GB memory (which I think is enough?), on my personal computer it was taking >8GB which seems very overkill and led to intensive swapping when having a browser + ~2 rust analyzer instances running on parallel


libs/common/src/document.rs
```diff
@@ -0,0 +1,80 @@
+use places::{addr::Addr, admin::Admin, poi::Poi, stop::Stop, street::Street};
```
crocme10, line 1: Should we have dependencies from common -> places, or rather from places -> common. I think the latter, and maybe those impl should be where you define the types instead of where you define the trait? What do you think?

libs/common/src/document.rs
```diff
@@ -0,0 +1,80 @@
...
+use serde::Serialize;
+
+/// Generic document.
+pub trait Document: Serialize {
+    /// Unique identifier for the document.
+    fn id(&self) -> String;
```
crocme10, line 7: Maybe add a TODO to see if we can just return a &'str to avoid all those clone()

crocme10 (commented):
Great job

crocme10 (approved):

PR #509, opened 2021-09-09
model: replace IndexConfiguration by generic config

remi-dupre:
This is an attempt to change how we interact with ES configuration to address several issues:

 - callers have to load configuration from files, even though we usually use default settings
 - it gave no clue on how to structure it from primary ports
 - we had to often deserialize/re-serialize data before just sending it to ES (useless computations put apart, it is mostly painful because it requires a few lines of code just to change a single value)
 - `IndexConfiguration` was the name used in Elasticsearch but also in ports (were we should talk of containers)

This [new interface](https://github.com/CanalTP/mimirsbrunn/pull/509/files#diff-5e8d5d8e5db6aa78e26ae1f27b8bcf29f67e21d0a97694eb4b3f9dcd50b28fd2R25-R52) remains quite generic, a default configuration is loaded into the binary and is used by default, but the caller can still specify a path which will be loaded from the elasticsearch adapter (to avoid code duplication and retain ES-specific implementations).

A few settings can also directly be overridden from this new `ContainerConfiguration`, but you may object that they are not specific to the ES backend ? (I'm think of `nb_replicas` and mostly of `nb_shards`).


libs/mimir2/src/domain/ports/secondary/import.rs
```diff
@@ -1,69 +0,0 @@
-use async_trait::async_trait;
```
remi-dupre, line 1: That's not really related to my PR, this is just not used anymore and would be redundant with `generate_index.rs`.

libs/common/src/container_config.rs
```diff
@@ -0,0 +1,17 @@
...
+///
+/// Such a configuration is structured as follows:
+///  - name: index name (string)
+///  - settings: raw index settings, see
+///    <https://www.elastic.co/guide/en/elasticsearch/reference/7.9/indices-create-index.html#create-index-settings>
+///  - mappings: raw index mappings, see
+///    <https://www.elastic.co/guide/en/elasticsearch/reference/7.9/mapping.html>
+///  - parameters: query parameters at index creation, see
+///    <https://www.elastic.co/guide/en/elasticsearch/reference/7.9/indices-create-index.html#indices-create-api-query-params>
+pub trait DefaultEsContainerConfig: ContainerDocument {
```
crocme10, line 15: Est-ce que tu as besoin d'un nouveau trait, pourquoi ne pas mettre ca directument dans `ContainerDocument`, avec une methode `config() -> Config`, ou `static_config()...`

libs/mimir2/src/adapters/secondary/elasticsearch/configuration.rs
```diff
@@ -30,14 +38,22 @@ pub struct IndexConfiguration {
 
 // FIXME A lot of work needs to go in there to type everything
```
crocme10, line 39: can probably remove this comment

libs/mimir2/src/adapters/secondary/elasticsearch/configuration.rs
```diff
@@ -30,14 +38,22 @@ pub struct IndexConfiguration {
...
-    pub value: serde_json::Value,
+pub struct IndexSettings(serde_json::Value);
+
+impl std::fmt::Display for IndexSettings {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::result::Result<(), std::fmt::Error> {
+        self.0.fmt(f)
+    }
 }
 
 // FIXME A lot of work needs to go in there to type everything
```
crocme10, line 49: Idem

libs/mimir2/src/adapters/secondary/elasticsearch/internal.rs
```diff
@@ -185,16 +189,7 @@ impl ElasticsearchStorage {
...
-            r#"{{ "mappings": {mappings}, "settings": {settings} }}"#,
-            mappings = config.mappings.value,
-            settings = config.settings.value,
-        );
-
-        let body: serde_json::Value =
-            serde_json::from_str(&body_str).context(JsonDeserialization {
-                details: String::from("could not deserialize index configuration"),
-            })?;
+        let body = json!({ "mappings": config.mappings, "settings": config.settings });
```
crocme10, line 192: luckily you're here ;-)

libs/places/src/utils.rs
```diff
@@ -0,0 +1,33 @@
+/// Build default configuration for given place type. By convention this will look in
```
crocme10, line 1: nice move! :+1:

libs/mimir2/src/domain/ports/secondary/storage.rs
```diff
@@ -1,9 +1,9 @@
 use async_trait::async_trait;
+use config::Config;
```
crocme10, line 2: Not sure this is a good idea to get read of the `Configuration`... It's a small cost to hide the actual implementation `Config` behind a newtype or a struct. But we can do that later

libs/common/src/container_config.rs
remi-dupre: Non effectivement on peut mettre ça dans le même trait, ou même directement dans les implémentations individuelles dans places.

PR #510, opened 2021-09-10
Update mappings for "addr" documents, adapt and test "openaddresses2mimir"

amatissart:
* "config/addr/mappings.json" is updated to be consistant with "config/admin/mappings.json" .  
The field names have also been ordered to make the comparison easier.

* "openaddresses2mimir" is now usable again, with a test case adapted from the old "tests/openaddresses2mimir_test.rs"


remi-dupre:
Looks good :tada: 

FYI I quickly built a rebased version of this branch and mine with new config mechanism, it is really quick to put together: https://github.com/remi-dupre/mimirsbrunn/tree/es7-addr-and-config

remi-dupre (approved):

libs/mimir2/src/adapters/secondary/elasticsearch/internal.rs
```diff
@@ -900,36 +900,43 @@ impl ElasticsearchStorage {
...
-    // pub(super) fn kjjjkkve_all_documents<D>(
-    pub(super) fn list_documents<D>(
+    pub(super) async fn list_documents<D>(
         &self,
         index: String,
     ) -> Result<Pin<Box<dyn Stream<Item = D> + Send>>, Error>
     where
         D: DeserializeOwned + Send + Sync + 'static,
     {
+        let client = &self.0;
```
crocme10, line 910: What motivated this change, moving the pit out of the state machine? This change then forces you to make 'list-document' async,

amatissart: I made this change to be able to return an `Error` when the request index does not exist. 
If this function is not async, it's not possible to return an `Error` after the request hits the ES server, and to handle the error from the "pit" response.

I am not sure there is a better way. Would you prefer to make this function return an empty stream instead and simply log a warning?

libs/mimir2/src/adapters/secondary/elasticsearch/internal.rs
```diff
@@ -900,36 +900,43 @@ impl ElasticsearchStorage {
...
+                .await
+                .unwrap()
+                .error_for_status_code()
+                .context(ElasticsearchClient {
+                    details: format!("cannot list index {}", index),
+                })?;
+            let response_body = response.json::<Value>().await.unwrap();
+            response_body
+                .get("id")
+                .expect("response has id")
```
crocme10, line 925: isn't that expect a bit brutal here? if Elasticsearch API changes, we'll put the program in panic mode.... shouldn't it be an 'ok_or(Err....)' ? I realize, of course, that the 'expect' is what i had written in the first place, but you can do better than me ;-). It also ties in with:
- a need for unit tests in that file
- the possible introduction of a 'State::Error' for clean termination

crocme10 (commented):
Hello, I've left a few open ended questions in there, which you can either handle now, or open an issue so that we remember to address them at a later point in time ;-)

PR #511, opened 2021-09-13
Add poi_type to poi's poi_type

crocme10:
POI have a field 'poi_type', and the poi_type has in turn two fields,
'id' and 'name'. This change makes sure that poi_type.id is prefixed
with 'poi_type'.


remi-dupre (approved):

SGrenet:
Add link with https://jira.kisio.org/browse/NAVITIAII-3603 in description please

PR #512, opened 2021-09-14
remove trimming of insee code

crocme10:
Code INSEE must be 5 characters. It is not clear why the code was previously trimmed when it started with zeros.


remi-dupre (approved):

SGrenet:
Add link with https://jira.kisio.org/browse/NAVITIAII-3597 in comment please

PR #517, opened 2021-09-17
Bragi API REST

crocme10:
This restores Bragi's original REST API.

* It uses the new mimir API to query Elasticsearch
* It uses warp instead of actix-web

TODO:

- [ ] Bragi integration tests
- [x] Warp filters test
- [x] reverse route
- [x] status route
- [x] Document how bragi works (architecture), how to use it, and how to configure it.
- [ ] Prometheus ??


crocme10:
I don't know why i pulled in all the documentation changes in there. Please disregard this part in your comments, I'll cleanup the PR later.

config/bragi/default.toml
```diff
@@ -0,0 +1,3 @@
+debug = false
+testing = false
+mode = "default"
```
amatissart, line 3: The purpose of these 2 values is not clear to me.  
Is `mode` the name of the config file that will be applied in addition to the default values?
But it seems that `mode = "testing"` is a special value, regarding Elasticsearch URL configuration. 

And is the `testing` flag still used?

src/bragi/settings.rs
```diff
@@ -0,0 +1,174 @@
...
+
+        builder = builder.add_source(File::from(settings_path).required(true));
+
+        // Add in a local configuration file
+        // This file shouldn't be checked in to git
+        builder = builder.add_source(File::with_name("config/local").required(false));
+
+        // Add in settings from the environment (with a prefix of APP)
+        // Eg.. `APP_DEBUG=1 ./target/app` would set the `debug` key
+        builder = builder.add_source(Environment::with_prefix("app"));
```
amatissart, line 127: Maybe `BRAGI_` could be used as a prefix?

src/bragi/settings.rs
```diff
@@ -0,0 +1,174 @@
...
+            builder
+                .set_override("elasticsearch.url", es_url)
+                .context(ConfigExtract {
+                    msg: String::from("Could not set elasticsearch url from environment variable"),
+                })?
+        } else {
+            builder
+        };
+
+        builder = if let Ok(es_url) = env::var("BRAGI_SERVICE_PORT") {
```
remi-dupre, line 145: Shouldn't we delegate all environment-defined settings to the `Environment::with_prefix("app")` defined earlier?

That taken apart I'm not sure there are many bragi specific stuff left here, and it is quite redundant with what seems to be done in `osm2mimir`, we may be fine with a generic construct to build a config from a recursive walk over a directory?

libs/mimir2/src/adapters/secondary/elasticsearch/status.rs
```diff
@@ -0,0 +1,30 @@
+use async_trait::async_trait;
+
+use super::ElasticsearchStorage;
+use crate::domain::model::status::StorageStatus;
+use crate::domain::ports::secondary::status::{Error as StatusError, Status};
+
+#[async_trait]
+impl Status for ElasticsearchStorage {
+    // This function delegates to elasticsearch the creation of the index. But since this
+    // function returns nothing, we follow with a find index to return some details to the caller.
```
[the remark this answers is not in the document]
crocme10: yes, gone

src/bragi/settings.rs
crocme10: correct, done

src/bragi/settings.rs
crocme10: I have removed the BRAGI_SERVICE_PORT specify code like you mentioned, and delegated to the Environment::with_prefix like you mentioned. Note that it needed to add a `.separator("_")` to work properly.

remi-dupre (approved):

PR #518, opened 2021-09-17
Doc

crocme10:
Adds 3 new documents

- Mimir Architecture: docs/architecture.md
- Bragi: docs/bragi.md
- Architecture Decision Records


remi-dupre:
I think you only intended to include fa5dc6c (change dir name, adds hexagonal architecture) here? Maybe you also forgot to track `bragi.md` in git? (happens all the time :wink:)

crocme10:
@remi-dupre you are correct, I will cleanup this mess shortly

PR #521, opened 2021-09-18
Update elasticsearch test with new configuration.

crocme10:
Update the `make check` command so that it takes into account all the workspaces.

Provide a bit more context in some error cases, especially during the creation of the configuration of an elasticsearch index.

Modify elasticsearch adapter tests to make sure they pass.

### DIFF
--- a/libs/mimir2/src/adapters/secondary/elasticsearch/configuration.rs
+++ b/libs/mimir2/src/adapters/secondary/elasticsearch/configuration.rs
@@ -45,6 +45,12 @@ impl std::fmt::Display for IndexSettings {
     }
 }
 
+impl IndexSettings {
+    pub fn new(value: serde_json::Value) -> IndexSettings {
+        IndexSettings(value)
+    }
+}
+
 #[derive(Debug, Clone, Default, Serialize, Deserialize)]
 pub struct IndexMappings(serde_json::Value);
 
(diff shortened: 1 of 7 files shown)


libs/mimir2/src/adapters/secondary/elasticsearch/configuration.rs
```diff
@@ -54,9 +60,15 @@ impl std::fmt::Display for IndexMappings {
...
+    }
+}
+
 #[derive(Debug, Clone, Default, Serialize, Deserialize)]
 #[serde(rename = "snake_case")]
 pub struct IndexParameters {
-    pub timeout: String,
-    pub wait_for_active_shards: String,
+    pub timeout: String,                // TODO How should we set this value
+    pub wait_for_active_shards: String, // TODO How should we set this value
```
remi-dupre, line 73: These values are set from [here](https://github.com/CanalTP/mimirsbrunn/blob/es7/libs/mimir2/src/adapters/secondary/elasticsearch/storage.rs#L51), `Config` makes use of the `Deserialize` crate to build `IndexConfiguration` and its fields.

These fields also don't have any default value which ensures they can't be omitted by in the configuration (they are defined here for default configurations: https://github.com/CanalTP/mimirsbrunn/blob/es7/libs/places/src/utils.rs#L17-L20)

remi-dupre (approved):

libs/mimir2/src/adapters/secondary/elasticsearch/configuration.rs
crocme10: What I meant was it's not clear what kind of values elasticsearch expects.... eg `wait_for_active_shards` expects a number, yet  we have a string. Intuitively, the variable name is phrased as a question 'wait_...' so I'd expect a boolean. I think it's confusing enough and I left a TODO so we probably need to
1. put in a usize instead of a String,
2. document with the description of elasticsearch: `This setting determines the number of shard copies that must be active before the client returns. The default value is 1, which implies only the primary shards.`
3. And yes, there is a default value, as mentioned in (2). So maybe should be `Option<...>`

That's why I left a TODO

PR #522, opened 2021-09-21
ES7: add bano2mimir

remi-dupre:
This just adds a binary for bano similar to `openaddresses2mimir` which was added in https://github.com/CanalTP/mimirsbrunn/pull/510.

I deduplicated the config loading logic but there is still a lot of similarities between implementations of the importers.

### DIFF
--- a/libs/common/src/config.rs
+++ b/libs/common/src/config.rs
@@ -1,6 +1,8 @@
+use crate::document::ContainerDocument;
 use config::Config;
+use std::path::PathBuf;
 
-pub fn config_from_args(
+fn config_from_args(
     args: impl IntoIterator<Item = String>,
 ) -> Result<Config, Box<dyn std::error::Error>> {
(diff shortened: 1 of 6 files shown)


libs/common/src/config.rs
```diff
@@ -25,3 +27,26 @@ pub fn config_from_args(
...
+        cfg_builder = cfg_builder.add_source(config::File::from(settings));
+    }
+
+    Ok(cfg_builder
+        .add_source(
+            config_from_args(args_override)
+                .map_err(|err| format!("couldn't override settings: {}", err))?,
+        )
+        .build()?)
+}
```
amatissart, line 52: Nice refactoring, that will be useful in #524 too.

amatissart (approved):

crocme10:
Looks good, but I think we're building a bit of legacy for lack of tests.... But who am I to say this ;-)

PR #524, opened 2021-09-21
Update mappings for Street and Poi, adapt osm2mimir CLI to new configuration

amatissart:
A few breaking changes are introduced in this PR:
 * `osm2mimir` can no longer import `Admin` documents (only `Street` and `Poi`). Importing 'Admin' can only be done with cosmogony2mimir.
 * The field `properties` in `Poi` (which typically contains OSM tags) is now indexed as a "flattened" object, instead of a list of key/value. It should make the filtering in ES queries easier and more performant.


src/settings/osm2mimir.rs
```diff
@@ -432,3 +295,55 @@ impl Source for Args {
         Ok(m)
     }
 }
+
+impl Args {
```
crocme10, line 299: This is probably overlapping with `common::config::load_es_config_for`

amatissart: Yes, I will rebase on "es7" and use the new `load_es_config_for`.

amatissart: The branch has now been updated to use `load_es_config_for`.

remi-dupre (approved):
LGTM, that actually feels nice to see all this setting handling code naturally removed from `osm2mimir` :smile:

libs/places/src/lib.rs
```diff
@@ -68,6 +68,16 @@ impl Place {
         }
     }
 
+    pub fn id(&self) -> &str {
```
crocme10, line 71: Do you need this function.... if the trait `common::document::Document` is in scope, do you still need it?

src/bin/osm2mimir.rs
```diff
@@ -0,0 +1,351 @@
...
+            .unwrap(),
+            client,
+        )
+        .await
+        .unwrap()
+    }
+
+    #[tokio::test]
+    async fn should_correctly_index_osm_streets_and_pois() {
+        let _guard = docker::initialize()
```
crocme10, line 216: You have to be careful with the guard. If an assert fails while you're holding it (in effect a mutex), it will poison it, and all other tests that depend on that mutex will fail (i think). What I usually try to do is get all the data in and out of elasticsearch, drop the guard, and then assert. Sometimes its difficult and I have to squeeze an assert! before dropping the guard.

libs/places/src/lib.rs
amatissart: `Place` does not implement Document. But I can add this `impl` while we are at it.
```rust
impl common::document::Document for Place {
    fn id(&self) -> String {
        match *self {
            Place::Admin(ref o) => o.id.clone(),
            Place::Street(ref o) => o.id.clone(),
            Place::Addr(ref o) => o.id.clone(),
            Place::Poi(ref o) => o.id.clone(),
            Place::Stop(ref o) => o.id.clone(),
        }
    }
}
```

src/bin/osm2mimir.rs
amatissart: That's a good point. In this case, would you recommend to run all search requests before any assert? Like this:
```rust
let results_rue_des_pres = search("Rue des Près").await;
let results_rue_du_four = search("Rue du Four à Chaux").await;
let results_rue_marcel = search("Rue Marcel Houdet").await;
let results_chatelet = search("Grand Châtelet").await;
// ...  and other searches.

drop(_guard);

// call assert! here.
// ...
```

Arguably, it looks a bit harder to read and maintain.   

Wouldn't this mutex problem be solved using [serial_test](https://crates.io/crates/serial_test) that implements its own mutex that should be cleaned up properly in case of failure ?

crocme10: I agree with you, it does not as readable.

So, yes, maybe you can evaluate the serial_test crate. Obviously the think we have to look for is if it improves on our predicament: what happens when an assert fails while you're in a serial test....

PR #525, opened 2021-09-23
WIP: add generic synonyms in elasticsearch filters

remi-dupre:
Add a filter in Elasticsearch to allow common synonyms, the `config` directory started  to be a bit messy so I moved specific Elasticsearch templates in `config/elasticsearch`. I didn't update the analyzers as it will probably relate to how Bragi is built in #517.

Note that this is not tested *yet*: my intention is to add end-to-end tests in a following PR like prototyped in [acceptance.feature][https://github.com/CanalTP/mimirsbrunn/blob/es7/features/acceptance.feature#L30-L37), but as existing tests for admins are very specific to admins it will need a bit of refactoring to generalize it to POIs / addresses.


crocme10 (approved):

crocme10:
Merged in #529

PR #533, opened 2021-09-28
bring some changes in master to es7

crocme10:
Since es7 and master started diverging, a few changes had been
introduced on master. This commit ports them over to es7.

Note that it would probably have been better to make it into small
dedicated commits... alas.


Makefile
```diff
@@ -4,7 +4,6 @@ SHELL=/bin/bash
 # Configuration
 .PHONY: check docker-build-bragi-release docker-build-bragi-master dockerhub-login push-bragi-image-master push-bragi-image-release wipe-useless-images help
 .DEFAULT_GOAL := help
-ELASTICSEARCH_TEST_URL=http://localhost:9201
```
amatissart, line 7: This change is probably the reason why tests depending on ES are failing.

PR #534, opened 2021-09-28
Autocomplete POST requests

crocme10:
There is still some work on Bragi, and how the shape and other query parameters are taken into account

There is also some documentation on bragi, and lots of tests on bragi's query parameters.


libs/mimir2/src/adapters/primary/bragi/routes/mod.rs
```diff
@@ -1,7 +1,10 @@
+use geojson::{GeoJson, Geometry};
```
remi-dupre, line 1: I notice that there are many modules that are built around a directory containing a single file `mod.rs`. Is there a reason for that? It seems easier to navigate without creating the directory (eg. `routes.rs` instead of `routes/mod.rs`).

docs/bragi.md
```diff
@@ -104,41 +104,63 @@ TODO How to specify negative long lat ?
...
 <table>
 <colgroup>
-<col style="width: 17%" />
-<col style="width: 29%" />
-<col style="width: 28%" />
-<col style="width: 25%" />
+<col style="width: 12%" />
+<col style="width: 18%" />
+<col style="width: 41%" />
+<col style="width: 27%" />
```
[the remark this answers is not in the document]
crocme10: I do use pandoc.... The problem with github markdown table is that its very limited. One of the feature I use here is to have multiple lines in a single cell.

So there is a directory docs/tbl in which I store the table source. They look like [this](https://raw.githubusercontent.com/CanalTP/mimirsbrunn/ad7798859b4308015a605ba459bde57d161dea20/docs/tbl/autocomplete-query-param.md), and then I use pandoc to turn this table into html and embed it in the markdown document.

I think the result looks ok, but I agree with you that the process is not that great. I don't know how else to do that.

libs/mimir2/src/adapters/primary/bragi/routes/mod.rs
crocme10: I agree with you. I did this because originally I intended to have `routes/autocomplete.rs`, `routes/reverse.rs`, .... one for each endpoint. The files are not that big yet, so I'll follow your recommendation.

crocme10:
Since I deleted the branch update-master after merging it, the pull request #534 closed itself

PR #535, opened 2021-09-29
ES aliases: Use atomic update instead of successive add and remove

### LINKED ISSUE #494: Atomic update of index aliases, to prevent any downtime
The Aliases API describes how to achieve atomic operations:
https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-aliases.html

amatissart:
Fixes #494 

`add_alias` and `remove_alias` methods are replaced with `update_alias`, relying on the Aliases API:  
https://www.elastic.co/guide/en/elasticsearch/reference/7.14/indices-aliases.html#indices-aliases

### DIFF
--- a/libs/mimir2/src/adapters/secondary/elasticsearch/internal.rs
+++ b/libs/mimir2/src/adapters/secondary/elasticsearch/internal.rs
@@ -2,8 +2,7 @@ use elasticsearch::cat::CatIndicesParts;
 use elasticsearch::cluster::ClusterHealthParts;
 use elasticsearch::http::response::Exception;
 use elasticsearch::indices::{
-    IndicesCreateParts, IndicesDeleteAliasParts, IndicesDeleteParts, IndicesGetAliasParts,
-    IndicesPutAliasParts, IndicesRefreshParts,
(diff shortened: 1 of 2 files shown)


remi-dupre (approved):
Neat!

crocme10:
This looks good.

I notice you've removed a lot of the error handling code that I had written, and which was admittedly heavy handed. But is relying on `error_for_status_code` enough? For example, could elasticsearch return a `acknowledged = false` but still have a response ok? I'm not sure.... My approach was to have a very narrow path for success, and anything else was an error. What do you think?

amatissart:
> This looks good.
> 
> I notice you've removed a lot of the error handling code that I had written, and which was admittedly heavy handed. But is relying on `error_for_status_code` enough? For example, could elasticsearch return a `acknowledged = false` but still have a response ok? I'm not sure.... My approach was to have a very narrow path for success, and anything else was an error. What do you think?

I updated the PR to check for `"acknowledged": true`. It's true that ES documentation is not very explicit on this point, so let's keep cautious.

Nevertheless I tried to simplify the error handling code a bit more :wink:

PR #536, opened 2021-09-29
Use 'serial_test' to run tests sequentially, remove mutex on docker init

amatissart:
`serial-test` is a light-weight dependency that will take in charge the mutex and avoid poisoning on failure.

Tests with `#[serial]` will run sequentially.

It's also possible to use `#[serial(something)]` to define independent sequences:
https://docs.rs/serial_test/0.5.1/serial_test/attr.serial.html

### DIFF
--- a/.github/workflows/create_publish_bragi_docker.yml
+++ b/.github/workflows/create_publish_bragi_docker.yml
@@ -11,6 +11,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
       - name: Get branch name
         shell: bash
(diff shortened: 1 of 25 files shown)


remi-dupre:
:+1: 

If we want to avoid a new dependency and continue after failing tests we could also ignore poisoning altogether by using something like that in `docker::initialize()`:

```rust
    let guard = AVAILABLE
        .lock()
        .unwrap_or_else(|poisoned| poisoned.into_inner());
```

But I don't think we are using this locking mechanism for anything else than those tests, so `serial_test` may deliver a better understanding of its intention?

crocme10 (approved):

PR #537, opened 2021-09-30
ES7: refactor end to end tests, add address related tests

remi-dupre:
The goal of this PR was to generalize end-to-end cucumber tests to add tests for other types of documents as a follow-up of #529.

I this regard, I renamed the test target from `admin` to `end_to_end` and split steps into separates files to make it clear what would be shared behavior (search.rs, admins.rs, download.rs, ...). While all this is quite straightforward, the potentially controversial part of my PR would probably be the rewriting of the tests states that I'll introduce bellow.

Now all kind of state data is stored in a unique structure holding the entire history of executed steps. Previously we were constrained to only know the state of the previously executed step, which was not convenient for example to know whether admins have been indexed at some point before. Also, this allows to ensure that if you need to index the cosmogony file for a specific region, it has been built for this specific region (before there was no way to have keep this region information in the context. This is implemented in a quite weird way using the `Any`trait, so I wouldn't the surprised if you object this design choice, but I'm pretty happy with it as it only takes a few line and avoid having to define a huge enum for of all kinds of steps.

I haven't been super clean with error handling, using `.expect("...")` by default, but I'm not sure whether maintaining a big `Error` type is worth the pain in tests? It may even be less convenient by hiding at which line the error was triggered instead of just panicking immediately.

PS: I allowed myself to remove some deprecated files in `tests/`, I think the branch starts to get closer to its complete state, so it may be time to start this kind of cleanup, git is here to keep track of legacy files anyway :smile:


tests/data/bano/limousin.csv
```diff
@@ -0,0 +1,2000 @@
+190050001T-10,10,Place Allègre,19240,Allassac,C+O,45.257780,1.475811
```
remi-dupre, line 1: Indexing addresses for a whole region was too slow, so I stuck with an arbitrary selection of 2000 addresses.

config/elasticsearch/synonyms_settings.json
```diff
@@ -17,6 +17,9 @@
                             "pt,pont",
                             "rle,ruelle",
                             "rte,route",
+                            "pl,place",
+                            "r,rue",
+                            "av,avenue",
```
remi-dupre, line 22: As Adrien just mentioned to me, these weren't there before they are prefixes of their respective alias and still found. It makes sense to add these as we boost perfect matches, we could be more conservative by using mapping like `"r=>rue"` instead (cf https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-synonym-tokenfilter.html#_solr_synonyms).

amatissart: We also noticed that synonyms are currently used in "prefix" and "ngram" analyzers only, and not on perfect matches. So these can be ignored for now. And perhaps a more advanced strategy is to be found to handle these synonyms more finely.
For example by using them on search-time only.  

See also this blog post about the different trade-offs in how synonyms can be used:  
https://www.elastic.co/blog/boosting-the-power-of-elasticsearch-with-synonyms

remi-dupre: :+1: 

I just removed it for now, it is unrelated to this PR anyway as those synonyms are already handled by prefix.

tests/steps/admin.rs
```diff
@@ -1,329 +1,166 @@
...
+            .expect("can't generate cosmogony file without downloading from OSM first");
+
+        // Check if the admin index already exists
+        let index = client
+            .find_container(String::from("munin_admin"))
+            .await
+            .expect("Looking up munin_admin");
+
+        // TODO: change this logic to check immutably what appends?
+        // If the previous step has been skipped, then we don't need to index the cosmogony file.
```
amatissart, line 139: I am not sure to understand this comment.  

Another approach could be to use a different "dataset" for each region, and check for the existence of `munin_admin_<region>` in Elasticsearch ?

tests/steps/address.rs
```diff
@@ -0,0 +1,106 @@
...
+            .expect("You must index admins before indexing addresses");
+
+        // Check if the address index already exists
+        let index = client
+            .find_container(String::from("munin_addr"))
+            .await
+            .expect("Looking up munin_address");
+
+        // TODO: change this logic to check immutably what appends?
+        // If the previous step has been skipped, then we don't need to index BANO file.
```
amatissart, line 62: This seems to be a copy/paste from "steps/admin.rs", but there is no previous step here, right?

amatissart (commented):
Looks good to me. Just a few questions.

tests/steps/address.rs
remi-dupre: Yeah, this "previous step" check may even not be required in admins as we are erasing data before the test, so there is no risk that the dataset was updated.

tests/steps/admin.rs
remi-dupre: Good idea :smile:

crocme10:
This looks great, it brings some clarity to these tests.
So I started working to add `DownloadBano` next to `DownloadOsm`, and then, once compiled, wanted to test .... just the new bano code. Could I run the end to end test just for bano? I know that it would imply that the admins are correctly indexed first, so that can go in the feature file, but how would i go about just requesting bano tests? I noticed that the Cucumber object has a `cli()` function where you can read arguments.... what do you think?

tests/steps/admin.rs
```diff
@@ -1,329 +1,166 @@
...
+        let Self(region) = self;
+        let client: &ElasticsearchStorage = ctx.get().expect("could not get ES client");
+
+        let gen_status = state
+            .status_of(&GenerateCosmogony(region.to_string()))
+            .expect("can't generate cosmogony file without downloading from OSM first");
+
+        // Check if the admin index already exists
+        let index = client
+            .find_container(String::from("munin_admin"))
```
crocme10, line 134: munin_admin, that's automagick, I think you should use `domain/model/configuration/root_doctype`

tests/steps/admin.rs
crocme10: I don't know what I meant by this comment. I need to comment my comments now... ;-)

remi-dupre: Now these tests create two indices: `munin_addr_limousin` and `munin_admin_limousin`.

tests/steps/admin.rs
remi-dupre: Yeah, I intended to work on indices names in the end and finally forgot about it, now these steps create indices with the name `munin_{type}_{dataset}`.

remi-dupre:
> This looks great, it brings some clarity to these tests. So I started working to add `DownloadBano` next to `DownloadOsm`, and then, once compiled, wanted to test .... just the new bano code. Could I run the end to end test just for bano? I know that it would imply that the admins are correctly indexed first, so that can go in the feature file, but how would i go about just requesting bano tests? I noticed that the Cucumber object has a `cli()` function where you can read arguments.... what do you think?

I'm sure I had answered but I must have included in a never submitted review :see_no_evil:.

Nice catch for the `cli()` function, it allows to filter on the scenario outline, so I made those outlines a bit more verbose (we may require to extend it later) in order to use it:

```bash
ELASTICSEARCH_TEST_URL=http://localhost:9200 cargo test --test end_to_end -- -e "Admin"
ELASTICSEARCH_TEST_URL=http://localhost:9200 cargo test --test end_to_end -- -e "Address"
```

We can provide any regex so I suppose this is flexible enough to cover any usage we may require.

crocme10 (approved):

PR #541, opened 2021-09-30
Post autocomplete

crocme10:
This pull request ends up being a mix bag of
- fixes to the import script 'import2mimir.sh`
- changes to the way the elasticsearch endpoint is specified in the configuration, using host and port rather than a whole url
- renaming of files for bragi: `routes/mod.rs` for instance becomes `routes.rs`


crocme10:
Supersedes #534

crocme10:
Since I  deleted the branch update-master after merging it,  the pull request #534 closed itself

src/bragi/server.rs
```diff
@@ -63,12 +67,12 @@ pub async fn run_server(settings: Settings) -> Result<(), Error> {
...
         .to_socket_addrs()
-        .context(SockAddr)?
+        .context(SockAddr { host, port })?
         .next()
         .ok_or(Error::AddrResolution {
-            msg: String::from("Cannot resolve elasticsearch addr"),
+            msg: String::from("Cannot resolve elasticsearch addr."),
         })?;
-    let elasticsearch_url = addr.to_string();
+    let elasticsearch_url = format!("http://{}", addr.to_string());
```
amatissart, line 75: Maybe it would be easier to have a single parameter `url`, instead of "host" and "port".  
Currently, it would be impossible to connect to ES via https, or to use an authenticated connection 
(eg: "https://user:password@elasticsearch.example.com")

amatissart (approved):
Looks good to me.  

I have a question about the connection to ES (see comment).

It does not cover all changes present in #534, is it deliberate?

src/bragi/server.rs
crocme10: i sort of agree. i thought it was nice, in the testing.toml, to just have to override the port.

PR #546, opened 2021-10-05
Addr multifile

crocme10:
This change enables bano2mimir to use a directory of csv files rather than just a single file.

It also brings an important change to the generate_index port by relaxing the constraint on the input stream of document beeing `Unpin`.

This change lacks:
* tests
* the ability to filter input documents in a directory based on an extension (eg, read all the .csv files)
* we need to synchronize bano2mimir and openaddress2mimir to make sure they use the same function.


Cargo.toml
```diff
@@ -64,6 +64,8 @@ rusqlite = { version = "0.23", optional = true }
 flate2 = { version = "1.0", features = [ "tokio" ] }
 toml = "0.5"
 tokio = { version = "1.6.0", features = [ "sync", "rt-multi-thread", "macros", "process" ] }
+tokio-stream = { version = "0.1.7", features = [ "fs" ] }
+tokio-test = "0.4.2"
```
remi-dupre, line 68: I think this dependency is not used here?

src/addr_reader.rs
```diff
@@ -50,15 +51,15 @@ where
 }
 
 pub async fn import_addresses_from_reads<T, F>(
-    client: &ElasticsearchStorage,
+    client: ElasticsearchStorage,
```
remi-dupre, line 54: I think we want to rely on a more generic port rather than directly ElasticsearchStorage. We could add a small FIXME as I don't think it takes a lot of time, but could add a bit of noise in the PR.

remi-dupre (approved):

Cargo.toml
crocme10: yes

PR #548, opened 2021-10-05
E2e ntfs2mimir

crocme10:
* Adds end to end tests for stops,
* Adds DownloadBano
* Adding a domain to search for in the gherkin language: 'When the user searches "addr, stop" for "<query>" ...

Note that this PR depends on the ntfs2mimir PR first


features/addresses/limousin.feature
```diff
@@ -22,7 +22,7 @@ Feature: Addresses
     # When using aliases, we should still fetch the query at the top of the
     # result.
     Scenario Outline: Addresses with aliases
-        When the user searches for "<query>"
+        When the user searches "addr" for "<query>"
```
remi-dupre, line 25: What would you think about being even more specific with something like `"addr_limousin"`, the test would become truly independent of what other region has been indexed.

Nevertheless I love this extra syntax, I had spotted the issue but wasn't sure how to tackle it :confused:

src/bin/bano2mimir.rs
```diff
@@ -55,6 +55,9 @@ struct Args {
     /// Elasticsearch parameters.
     #[structopt(short = "c", long, default_value = "http://localhost:9200/munin")]
     connection_string: String,
+    /// Name of the dataset
+    #[structopt(short = "d", long, default_value = "fr")]
+    dataset: String,
```
remi-dupre, line 60: What is the motivation behind this new `--dataset` parameter, compared to `-v container.dataset=[...]`?

features/addresses/limousin.feature
remi-dupre: Well, I think it's a good thing to keep it simple until we eventually need something more complicated. This is not a behavior that would be hard to extend if needed :wink:

src/bin/bano2mimir.rs
crocme10: although there is a slight issue with the naming i feel:
* container is a technical term. it refers to our abstraction of an object containing documents, be it an index or a table.
* dataset is a business logic.
I think its awkward to have this configuration key straddling both domains.

tests/steps/search.rs
```diff
@@ -1,27 +1,33 @@
...
-use places::admin::Admin;
+use mimir2::{
+    adapters::{
+        primary::{
+            common::dsl::build_query, common::filters::Filters, common::settings::QuerySettings,
+        },
+        secondary::elasticsearch::ElasticsearchStorage,
+    },
+    domain::{model::query::Query, ports::primary::search_documents::SearchDocuments},
+};
```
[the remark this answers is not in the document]
crocme10: ok....no pb. So should we have a document describing programming conventions?

PR #603, opened 2021-10-26
Introducing index and components templates

crocme10:
- [x] Correctly loads components and index templates
- [ ] Strategy for breaking down components
- [ ] Update components and index templates after cluster has been created
- [ ] ...

* A new folder is created (`config/elasticsearch/templates`), containing component and index templates.
* The settings and mappings of each place has been removed, because it is now in the templates
* A new primary port `ConfigureBackend`, with a method `configure(directive, config)`: I don't want to have too many methods in that trait, so you just have a directive, which is just a string now, but it says 'what you want to configure', and the config, which is the data you want to pass all the way to the backend.
* A new secondary port, similar to the primary one.
* Two methods in `elasticsearch/internal.rs` one for creating index templates, one for creating component templates.
* A new primary adapter (I'm really not sure this is the right place for that code) `adapters/primary/template.rs`, with a method to import templates
* A new binary `src/bin/ctlmimir.rs` (can't think of a better name, maybe you can), which calls the primary adapter.


libs/mimir2/src/adapters/secondary/elasticsearch/configuration.rs
```diff
@@ -1,24 +1,36 @@
...
 }
 
+// impl From<serde_json::Error> for Error {
+//     fn from(source: serde_json::Error) -> Self {
+//         Self::InvalidConfiguration {
+//             details: source.to_string(),
+//         }
+//     }
+// }
+
```
remi-dupre, line 33: Let's not forget about removing that

libs/mimir2/src/adapters/primary/templates.rs
```diff
@@ -0,0 +1,100 @@
...
+            let client = client.clone();
+            let template_type = template_type.clone();
+            async move {
+                let config = config::Config::builder()
+                    .set_default("elasticsearch.name", template_name)
+                    .unwrap()
+                    .add_source(config::File::new(
+                        template.to_str().unwrap(),
+                        config::FileFormat::Json,
+                    ))
```
remi-dupre, line 57: I think we should set this file as the default to and override it with the full configuration to allow overriding some values, such as `number_of_shards`?

libs/mimir2/src/adapters/primary/templates.rs
```diff
@@ -0,0 +1,100 @@
...
+        .await?
+        .try_for_each(|template| {
+            let template_name = template
+                .file_stem()
+                .expect("file stem")
+                .to_str()
+                .expect("template_name")
+                .to_string();
+            let client = client.clone();
+            let template_type = template_type.clone();
```
remi-dupre, line 49: I think it could help to make this enum `Copy`

remi-dupre (commented):
Maybe we should merge ecefce65b7a92ce722bc47f66fe3f1f19876a606 in a separate PR right away?

config/elasticsearch/templates/components/mimir-base.json
```diff
@@ -0,0 +1,34 @@
+{
```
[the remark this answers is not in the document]
crocme10: je prends le nom du fichier (plus exactement le stem) comme le nom du template (mimir-base.json -> mimir-base).
L'interet c'est qu'on peut facilement enlever tous les templates lie a mimir en utilisant une regex `mimir-*`.

libs/mimir2/src/adapters/primary/templates.rs
crocme10: difficult because we don't know what templates this setting would belong to

remi-dupre (approved):

PR #604, opened 2021-10-27
Add elasticsearch explain feature to bragi

fatal69100:
-> Review requested for the last 2 commits

Add elasticsearch explain endpoint to bragi (https://www.elastic.co/guide/en/elasticsearch/reference/7.14/search-explain.html)


README.md
```diff
@@ -27,7 +27,7 @@ This API provices several services:
 
 | feature            | route      | Parameters |
 | ------------------ | ---------- | ---------- |
-| staus              | `/status`  | None       |
```
remi-dupre, line 30: This made me think that it would be a good idea to update the README to add `/autocomplete-explain` to the list of existing endpoint in previous section.

Also, this is a detail but this has broken the alignment of vertical lines (which is still syntactically ok, html rendering not broken) :smile:

libs/mimir2/src/adapters/primary/bragi/api.rs
```diff
@@ -8,6 +8,16 @@ use crate::adapters::primary::common::filters::Filters;
 use common::document::ContainerDocument;
 use places::{addr::Addr, admin::Admin, poi::Poi, stop::Stop, street::Street};
 
+#[derive(Debug, Default, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ForwardGeocoderExplainQuery {
+    pub id: String,
```
remi-dupre, line 14: Maybe this should be called `doc_id` to make it very clear that this is not homogeneous with other parameters which relate to the query?

What do you think? This may be a bad idea, I usually have bad taste when it comes to naming think.

remi-dupre (changes requested):
:+1:

README.md
fatal69100: yes, btw I was searching something that could generate automatically documentation from API (such as swagger in Rust) because it will be easier to maintained documentation if it is generated from the code.

libs/mimir2/src/adapters/primary/bragi/api.rs
fatal69100: I think you are absolutely right for this one !! It's much more expressive

fatal69100:
Can you review the last 2 commits with the serde default value ? I am unsure about the limit parameter for  the ReverseGeocoderQuery struct

remi-dupre:
> Can you review the last 2 commits with the serde default value please? I am unsure about the limit parameter for the `ReverseGeocoderQuery` struct

I think this is okay :+1:

remi-dupre (approved):

PR #607, opened 2021-10-28
Add limit parameter to reduce the number of result

fatal69100:
This is base over #604

- Default : 10
- add optional limit parameter to Bragi , e.g :  `http://localhost:5000/api/v1/autocomplete?q=Lie&type[]=street&type[]=zone&zone_type[]=city&limit=3`


libs/mimir2/src/adapters/primary/common/dsl.rs
```diff
@@ -30,8 +31,21 @@ pub fn build_query(
                 }
             }
         })
+    } else if filters.is_empty() & limit.is_some() {
+        json!({
+            "from":0,
```
remi-dupre, line 36: Do we want to be this explicit? I feel like the default value (0) is intuitive

libs/mimir2/src/adapters/primary/common/dsl.rs
```diff
@@ -30,8 +31,21 @@ pub fn build_query(
...
+            "size":limit.unwrap_or(10),
```
remi-dupre, line 37: Maybe this default (10) should be moved into a `const`?  (or settings if we want to overthink it, but I don't think we do)

libs/mimir2/src/adapters/primary/common/dsl.rs
```diff
@@ -21,7 +22,7 @@ pub fn build_query(
     let string_query = build_string_query(q, &settings.string_query);
     let boosts = build_boosts(q, settings, coord);
     let filters = build_filters(shape);
-    if filters.is_empty() {
+    if filters.is_empty() & !limit.is_some() {
```
remi-dupre, line 25: That is mostly anterior to your changes, but I feel uncomfortable about how this is structure, we do some kind of manual pattern matching over `(filters, limit)` building essentially the same query in each branch. I think it would be better to build the main body of the query in a mutable variable, and then independently add a `size`/`filter` field if corresponding parameter is set.

remi-dupre (changes requested):
This is base over https://github.com/CanalTP/mimirsbrunn/pull/604.

(sorry I reviewed it in the wrong order)

fatal69100:
It's my bad because I did not mention it

PR #608, opened 2021-10-29
add a `limit` parameter for simple queries in `Query::QueryString`

remi-dupre:
This is a change with very limited impact, it could help making things a bit more explicit in tests here, and to be honest the reason I am making this suggestion is mostly that [it can help us](https://github.com/Qwant/fafnir/pull/63#discussion_r738967912) with the tests for our POI importer, which requires a bit more than the `10` documents limit to fetch all results.


remi-dupre:
Deprecated by #607 as `search_documents` now takes a `limit` parameter.

Issue #619, opened 2021-11-05
Remove log warning


crocme10:
When starting a binary we get

```
 | `mimir=debug` would enable the DEBUG level for the `mimir` target
 = note: the static max level is `info`
 = help: to enable DEBUG logging, remove the `max_level_info` feature
```

remi-dupre:
This can be solved at runtime by specifying the env variable *RUST_LOG* to something like `RUST_LOG=info`.

I enforced it [in the docker images](https://github.com/CanalTP/mimirsbrunn/blob/master/docker/run_with_default_config.sh#L12) but it indeed doesn't seem ideal to have a warning by default. I think we could solve this either by letting tracing use its own defaults are using a [`if cfg!(debug_assertions)`](https://doc.rust-lang.org/reference/conditional-compilation.html#debug_assertions) to specify different defaults in release and debug.

PR #620, opened 2021-11-05
increase default mimir log level

crocme10:
Currently we have a default RUST_LOG string which includes `mimir=debug`. This is not necessary, and this PR changes that to `mimir=default`


crocme10:
Fixes #619

PR #625, opened 2021-11-08
allow to index from streams of any lifetime

remi-dupre:
The constraint was previously enforced by automock, but this is finally not the direction we took for tests.

This relaxing will allow to attach some context to the input stream, an application of this that lead me to this suggestion was the collection of statistics:

```rust
    let mut count_errors = HashMap::new();

    let pois = read_pois(raw_xml, admin_geofinder).filter_map(|poi| {
        future::ready({
            poi.map_err(|err| *count_errors.entry(err).or_insert(0) += 1)
                .ok()
        })
    });

    create_index(
        &mimir_es,
        &raw_config,
        &settings.container_tripadvisor.dataset,
        IndexVisibility::Private,
        pois,
    )
    .await
    .expect("error while indexing POIs");
```

In this example, the stream `pois` only lives as long as `count_errors`, which is not static.

### DIFF
--- a/libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
+++ b/libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
@@ -516,7 +516,7 @@ impl ElasticsearchStorage {
     ) -> Result<InsertStats, Error>
     where
         D: Document + Send + Sync + 'static,
-        S: Stream<Item = D> + Send + Sync + 'static,
+        S: Stream<Item = D> + Send + Sync,
     {
         let stats = Arc::new(Mutex::new(InsertStats::default()));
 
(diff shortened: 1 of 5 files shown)


remi-dupre:
@crocme10 I've removed the changes concerning automock from this PR. I initially thought this was tied because I replaced `'static` with an explicit lifetime, which leads to [a bug in mockall](https://github.com/asomers/mockall/issues/299), but I finally figured out we can left it elided, which is even simpler for automock.

I still think this is not a good idea to keep this dependency as long as we don't use it, but this can be discussed out of this PR :smile:

remi-dupre:
I found a way to to avoid propagating mockall's constraint out of the tests.

The idea is to add a lifetime parameter `'s` to `Storage<'s>` : this trait is implemented for a back-end if it implements storage primitives, with input streams of lifetime `'s`.

So here the `MockStorage` implements `Storage<'static>` like before while `ElasticsearchStorage` implements `Storage<'s>` for any lifetime. This works great but the issue is that there is a bit more of work to do because the attribute macro doesn't support such fined-grain tweaking.

So this is a proof of "we can relax the constraints and still use mocks from mockall crate if we need them someday", I'm not sure what should be our policy here.

PR #630, opened 2021-11-09
typed Container Configuration

### LINKED ISSUE #626: Allow to disable `forcemerge`
This may require some refactoring about how we currently handle container settings.

Also, force merging seems like a very interesting feature but currently has stability issues, do we want to add a ticket in this `launch-2.0` or is it something we should investigate once essential features are stabilized?

remi-dupre:
Container configuration was previously mixed with index mappings and parameters. This had a very specific behavior which differed with other configuration mechanisms:

 - it was loaded at compile time and associated to a document type, then accessed through the static method `default_es_container_config`
 - the container name was updated from main configuration, but the logic was implemented specifically in each binaries

Now the index mappings and parameters are generated by *ctl2mimir*, so all this complexity isn't required anymore. I've changed `generate_document` to take a single `ContainerConfig` struct as parameter with `name`, `dataset` and `force_merge` fields. This has several benefits:

 - this config piece can be loaded from the main configuration at `config.container` and all fields can be overridden at runtime. It is also easier to split it in two for `osm2mimir`: `config.poi_container` / `config.street_container`
 - using a typed configuration for primary ports and elasticsearch internals removes a lot of error handling for the multiple layers where we need to access part of the configuration : the configuration is parsed once, at the initialization of the importer

fixes #626


config/elasticsearch/admin/TODO.md
```diff
@@ -1,16 +0,0 @@
...
- * `"max_gram": 10`
- Was 20. Why not 20 (ex: german names)
-
- * No synonyms. Critical for addresses (bd av etc).
-
- * Analyzers
- 	* No prefix without ellision. (Perhaps it was unused with latest versions ?)
- 	* "autocomplete" seems equivalent to "prefix_elision"
- 	* "autocomplete_search" seems equivalent to "word_elision"
- 	* No "ngram" tokenizer for fuzzy search ?
```
remi-dupre, line 14: Most of this is deprecated but I think this as already been converted into issues

libs/tests/src/admin.rs
```diff
@@ -1,160 +0,0 @@
-use config::Config;
```
remi-dupre, line 1: This has been moved to *libs/tests/src/cosmogony.rs*, but was still tracked by git

libs/mimir/src/adapters/secondary/elasticsearch/mod.rs
```diff
@@ -86,41 +87,19 @@ impl ElasticsearchStorageConfig {
...
-
-        let config = Config::builder().build().expect("build empty config");
-
-        let res = client.create_container(config).await;
-
-        assert!(res
-            .unwrap_err()
-            .to_string()
-            .contains("Container Creation Error: Invalid Configuration"));
-    }
```
remi-dupre, line 122: These tests can now be removed because we enforce to get a valid configuration through the input type.

libs/mimir/src/domain/ports/primary/generate_index.rs
```diff
@@ -14,7 +14,7 @@ use tracing_futures::Instrument;
 pub trait GenerateIndex {
     async fn generate_index<D: ContainerDocument + Send + Sync + 'static>(
         &self,
-        config: Config,
+        config: &ContainerConfig,
```
crocme10, line 17: You are moving to a backend aware configuration (because the ContainerConfig has, and probably will get more, Elasticsearch specific paramaters. So you are sort of breaking the rule 1 of hexagonal architecture, which is to introduce a dependency from the hexagon to something outside of it.

libs/places/src/stop.rs
```diff
@@ -163,7 +163,11 @@ impl Document for Stop {
     }
 }
 
-crate::impl_container_document!(Stop, "stop");
+impl ContainerDocument for Stop {
```
crocme10, line 166: Why this change of heart with macros?

remi-dupre: Previously this macro was in charge of loading ES parameters and mappings and embed them in a `config::Config` object: https://github.com/CanalTP/mimirsbrunn/pull/630/files#diff-53a05293ae5bc8ba07df457d575cc42000b07945a6fce6c35a7286d92032c363L17-L36

Now that this is handled by templates in *ctlmimir*, the implementation of `ContainerDocument` is extremely concise, thus using a macro doesn't doesn't do much more that shadowing Rust's `impl` syntax.

crocme10 (approved):

PR #634, opened 2021-11-10
address importers rationalization

remi-dupre:
I initially just wanted to add extra logs in the address importers which was failing silently, eg. when using an invalid file path, but I think there was a few other issues we wanted to handle.

Here is what changed:

 - added a tracing instrument for each input file, log when there is an issue with deserializing the CSV into a "record" (before calling `into_addr`), log when there is an issue with opening a file
 - de-duplicated addr parsers: there was two separate co-existing implementations, one for bano (close to what was there before async) and one for openaddresses
 - one of the implementations didn't support gzip files, I re-implemented the decoder using `async-compression` (which still uses `flate2` as a back-end)
 - when specifying a directory as input, we will walk recursively instead of just file directly in the input

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -158,7 +158,7 @@ dependencies = [
  "async-graphql-value",
  "async-stream",
  "async-trait",
- "bytes 1.1.0",
+ "bytes",
  "chrono",
  "fnv",
  "futures-util",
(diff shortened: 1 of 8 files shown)


crocme10:
great job!

PR #639, opened 2021-11-16
Update cargos

crocme10:
Change edition 2018 -> 2021
Simplify authors
Update some crates (tokio, ...)


.github/workflows/docker.yml
```diff
@@ -27,6 +27,14 @@ jobs:
...
         with:
           fetch-depth: 0
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+
+      - name: Check versions
+        run: |
+          rustc --version
+          cargo --version
```
remi-dupre, line 37: As this is unrelated to the Rust version used by the docker builder, we probably don't need to add it here?

Cargo.toml
```diff
@@ -1,10 +1,10 @@
 [package]
 name = "mimirsbrunn"
 version = "2.0.0-rc1"
-authors = ["Kisio Digital <team.core@kisio.com>", "Qwant", "Guillaume Pinot <texitoi@texitio.eu>"]
+authors = ["Kisio Digital", "Qwant"]
```
remi-dupre, line 4: Thanks! :smile:

.github/workflows/docker.yml
crocme10: yes, agreed, and removed.

It would be nice to use the version of rust found in rust-toolchain in the dockerfile

remi-dupre (approved):

PR #642, opened 2021-11-17
update to cucumber 0.10 and use global state

remi-dupre:
This new cucumber version comes with a few breaking API changes, but also quite convenient attribute macros to define steps which helps us with having a very minimal amount of code for input parsing / validation.

I also implemented a `GlobalState` which is just a wrapper around a mutex guard, this allows to keep track of the state during the entire execution of end to end tests.

Note that there is probably no simpler way to share this state (from https://docs.rs/cucumber/0.10.2/cucumber/trait.World.html):

> This crate doesn’t provide out-of-box solution for managing state shared across scenarios, because we want some friction there to avoid tests being dependent on each other. If your workflow needs a way to share state between scenarios (ex. database connection pool), we recommend using once_cell crate or organize it other way via shared state.


features/addresses/limousin.feature
```diff
@@ -7,24 +7,26 @@ Feature: Addresses
 
     # With 'Exact Match', we expect the query to be found at the top of the
     # result because the query exactly matches the name / label of the target.
+    @unittest
```
remi-dupre, line 10: The new version of cucumber makes it a bit less natural to explicitly run a list of feature files, so I included the whole *features/* directory and added this tag which selects which scenarios must be included in `cargo test --test end_to_end`.

crocme10 (approved):

PR #643, opened 2021-11-18
update to cucumber 0.10 and use global state

remi-dupre:
Hello ? Could the CI trigger ?


remi-dupre:
Thanks Github, just had to ask politely.

PR #645, opened 2021-11-19
catch elasticsearch bulk errors

remi-dupre:
Parse the `error` field of a bulk insertion, this will help investigating ES errors and makes the error propagation more explicit.

See https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-update for example response formats.


libs/mimir/src/adapters/secondary/elasticsearch/models.rs
```diff
@@ -31,17 +31,138 @@ impl<D> ElasticsearchSearchResponse<D> {
...
+        }
+    })
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use serde_json::json;
+
+    fn sample() -> serde_json::Value {
```
crocme10, line 97: Maybe you should indicate how you got that sample, so that we can reproduce, or get a new sample if the format changes?

crocme10 (approved):

libs/mimir/src/adapters/secondary/elasticsearch/models.rs
remi-dupre: Good idea, I've just updated it, there is a complete example in the documentations from which I was able to pick :)

PR #646, opened 2021-11-19
Add port to update documents before publishing

remi-dupre:
The adds a more flexible API to the `generate_index` ports which allows to update documents before publishing it. This is very targeted for our use-case using mimir as a library, where we need to import data from tripadvisor which are split into different datasets: https://github.com/Qwant/fafnir/pull/71.

This new API can be called with a chain of methods, like that:

```rust
client.init_container(&container_config)
    .await?
    .insert_documents(doc_stream1)
    .await?
    .insert_documents(doc_stream2)
    .await?
    .update_documents(update_stream)
    .await?
    .publish() // there is even a warning if this last call is forgotten
    .await?;
```

The `generate_index` method is still there, I think it is still relevant as a shortcut for the new API as it will group all potential failure points in a single call.

Updates can be defined as a pair of the id of a document together with an instance of the following enum (the only available operation is to set a field, but at least we're explicit):

```rust
pub enum UpdateOperation {
    /// Update a field `ident` with given value
    Set { ident: String, value: String },
}
```

NOTE: This is base on #645 which should be reviewed & merged first (:heavy_check_mark:)

### DIFF
--- a/libs/mimir/src/adapters/secondary/elasticsearch/storage.rs
+++ b/libs/mimir/src/adapters/secondary/elasticsearch/storage.rs
@@ -1,7 +1,9 @@
 use async_trait::async_trait;
 use config::Config;
 use futures::future::TryFutureExt;
-use futures::stream::Stream;
+use futures::stream::{Stream, StreamExt};
+use serde::Serialize;
+use serde_json::json;
 
 use super::configuration::{ComponentTemplateConfiguration, IndexTemplateConfiguration};
 use super::internal;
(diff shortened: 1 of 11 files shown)


libs/mimir/src/adapters/secondary/elasticsearch/storage.rs
```diff
@@ -61,15 +64,12 @@ impl<'s> Storage<'s> for ElasticsearchStorage {
         S: Stream<Item = D> + Send + Sync + 's,
     {
         self.add_pipeline(
-            String::from(include_str!(
-                "../../../../../../config/pipeline/indexed_at.json"
-            )),
-            String::from("indexed_at"),
+            include_str!("../../../../../../config/pipeline/indexed_at.json"),
```
crocme10, line 67: This should probably refer to the CARGO_MANIFEST_DIR

PR #648, opened 2021-11-30
replace structopt with clap-3.0

remi-dupre:
I'm not 100% this is a good idea as clap 3.0 is still in beta, but there are a few benefits:

 - we would update directly to next major version (I don't think structopt will still be maintained as [clap will completely superset its features](https://github.com/clap-rs/clap/blob/master/FAQ.md#how-does-clap-compare-to-structopt))
 - we can fix ambiguity of syntax `command -s "key1=val1" -s "key2=val2" run` (currently doesn't work, there must be another parameter that follows *-s* parameters, this is especially annoying for docker images were `--config-dir` and `--run-mode` [are prepend to command arguments](https://github.com/CanalTP/mimirsbrunn/blob/master/docker/run_with_default_config.sh#L17). A solution is introduced with 3.0 as the behavior for argument lists is split in two: `multiple_values = false` will enforce to put a single argument after `-s` and `multiple_occurrences = true` allows to specify several `-s`.

Any advice ? Maybe we could delay this but also keep the diff accessible (in a draft PR or link it in an issue) to deploy it as soon as clap 3.0 is stable (https://github.com/clap-rs/clap/issues/2869).


remi-dupre:
Any opinion on that @crocme10?

I just noticed that in our custom importer ([fafnir](https://github.com/Qwant/fafnir)) we only have `-s` parameters so there is no trivial workaround when using a docker image with default `--config-dir` and `--run-mode` :confused:

PR #649, opened 2021-12-02
Generate package for jessie

azime:


.github/workflows/release.yml
```diff
@@ -14,13 +14,11 @@ jobs:
...
-            rust: 1.44.1
-          - version: stretch
-            packages: pkg-config libssl1.0-dev
-            rust: 1.44.1
+            rust: 1.54.0
+
           - version: buster
             packages: pkg-config libssl-dev
-            rust: 1.48.0
+            rust: 1.54.0
```
remi-dupre, line 21: Shouldn't we update to latest Rust version (1.56.1) directly?

remi-dupre (commented):
Out of curiosity: why do you need both packages on your end?

.github/workflows/release.yml
azime: :+1:

benoit-bst (approved):

kadhikari (approved):

PR #650, opened 2021-12-02
Rename binaries for migration purpose

benoit-bst:
## Cluster  Migration
:warning:  Remove this step after ES migration 2->7.
We need it to make version 2 and 7 coexist.
Be careful, binaries will not have the same name !

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -25,6 +25,18 @@ jobs:
           apt-get update
           apt-get install -y curl wget build-essential
           apt-get install -y ${{ matrix.packages }}
+      # TODO :
+      # Remove this step after ES migration 2->7.
+      # We need it to make version 2 and 7 coexist.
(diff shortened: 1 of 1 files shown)


kadhikari (approved):

remi-dupre:
I'm not a big fan of this change as we're trying to stabilize the pipelines on our end.

It's pretty hard to suggest a compromise from my end as I don't exactly see what is your use-case. Maybe it would be possible to publish an alternative image with another label? Do you merge images to get all binaries in the same one, maybe it could be renamed at this step too?

benoit-bst:
@remi-dupre 
Finally, we choose to keep the current GH job as such and create a specific duplicate for our migration purpose. 
Hope it will be simpler for you.

kadhikari (approved):

remi-dupre (approved):

PR #651, opened 2021-12-02
Cleanup Dependencies

remi-dupre:
Result of an effort to globally update mimirsbrunn dependencies.

At the time I opened the PR this reduced the number of crates to compile during a `cargo build` from 465 to 404, in addition to maintainability benefits. There may still be some outdated dependencies, I'm waiting for other repos to be merged in order to use `cargo outdated` as it doesn't work with a `[patch.crates-io]` section :grimacing:


crocme10:
Thank you for this tedious job.

bye bye GraphQL, sigh...

remi-dupre:
We still need a deployment of [address-formatter-rs](https://github.com/CanalTP/address-formatter-rs), however I'm not sure who to ping about that.

I'm also not sure why I removed grahpql from dependencies, it is still used in bragi so it probably mean it should be imported elsewhere :o

remi-dupre:
This should be done, the snafu upgrade introduced a big diff as helper struct are now suffixed with `Snafu`, but I think this helps with readability anyway.

`cargo audit` still reports a vulnerability in the `time` crate, we could update it as soon as chrono itself bumped its dependency to 0.3.x (https://github.com/chronotope/chrono/pull/639).

NOTE: the CI is failing but I think it mostly comes from an issue from Github, it runs on an older version of the code, on my fork the CI seems to be working fine though: https://github.com/remi-dupre/mimirsbrunn/runs/4715463120?check_suite_focus=true

PR #652, opened 2021-12-02
Add a parameter to search in a specific index

fatal69100:
Add `override_indices_name` parameter to the autocomplete endpoint in order to search in a not default index (`poi-tripadvisor` in our case)


libs/mimir/src/adapters/primary/bragi/handlers.rs
```diff
@@ -187,16 +187,47 @@ where
...
+        names.to_vec()
     } else {
-        vec![
-            String::from(Admin::static_doc_type()),
-            String::from(Street::static_doc_type()),
-            String::from(Addr::static_doc_type()),
-            String::from(Stop::static_doc_type()),
-            String::from(Poi::static_doc_type()),
-        ]
+        if let Some(ts) = types {
```
remi-dupre, line 198: That comes from clippy, but I think it makes sense that he recommends removing one layer of nest by replacing this with a `else if` :smile:

remi-dupre (approved):
That's very nice that you were able to keep it concise and confined in the bragi adapter, I was afraid we'd have to propagate this change to a lower level :confused: 

Maybe @crocme10 would have some suggestions as he spent the most time working on this component :slightly_smiling_face:

libs/mimir/src/adapters/primary/bragi/api.rs
```diff
@@ -58,6 +59,7 @@ impl From<(ForwardGeocoderQuery, Option<Geometry>)> for Filters {
                 zone_types,
                 poi_types,
                 limit,
+                poi_dataset: _,
```
crocme10, line 62: This change is nice. It is also a bit counter intuitive for me. I would have expected, if you wanted to work with `poi_tripadvisor`, that you'd specify the type = `poi` and the dataset = `tripadvisor`. Instead you specify type = None, and poi_dataset is `poi_tripadvisor`, or am I wrong?

crocme10: And so, in that same line, I would have maybe chang the search primary port: the first argument `doc_types: Vec<String>`, it should probably be something like `container_identifiers: Vec<ContId>`, followed by 
```
enum ContId {
  Alias(String),                          => eg admin, poi, which will be translated to munin_admin, munin_poi, ....
  AliasDataset(String, String)   => eg (poi, tripadvisor), (stop, idfm), ....
}
```

libs/mimir/src/adapters/primary/bragi/handlers.rs
crocme10: I think these default indices should be...... configurable!!! Probably at Qwant you're not thinking too much about Stops, so this should not be here.

fatal69100:
@crocme10 I have given a try to your proposal that seems logical to me with the current behavior that specify the type = poi and the dataset = tripadvisor.  There is another PR, also well done, that implement the previous behavior : I don't mind merging https://github.com/CanalTP/mimirsbrunn/pull/668 first. I just want your though about what I did because I am still not at ease.

fatal69100:
I think this pr has been give up. Closed by #668

PR #665, opened 2021-12-15
Parse timeout in autocomplete query

pbench:
If a "timeout=2000" is present in the query made to bragi, use this timeout value for querying elasticsearch (instead of the default one in the config)

### DIFF
--- a/libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
+++ b/libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
@@ -1029,7 +1029,7 @@ impl ElasticsearchStorage {
         D: DeserializeOwned + Send + Sync + 'static,
     {
         let indices = indices.iter().map(String::as_str).collect::<Vec<_>>();
-        let timeout = timeout.unwrap_or_else(|| self.config.timeout);
+        let timeout = timeout.unwrap_or(self.config.timeout);
         let search = self
             .client
             .search(SearchParts::Index(&indices))
--- a/libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
+++ b/libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
@@ -1023,7 +1023,7 @@ impl ElasticsearchStorage {
         indices: Vec<String>,
         query: Query,
         limit_result: i64,
-        timeout : Option<Duration>, //in milliseconds
+        timeout: Option<Duration>, //in milliseconds
     ) -> Result<Vec<D>, Error>
     where
         D: DeserializeOwned + Send + Sync + 'static,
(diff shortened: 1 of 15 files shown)


antoine-de:
isn't it dangerous to have a way to bypass the timeout on the api? This could lead of an overload on ES no?

remi-dupre:
> isn't it dangerous to have a way to bypass the timeout on the api? This could lead of an overload on ES no?

I kind of agree, the config can be overridden at launch with `-s 'elasticsearch.timeout=30_000'`, but maybe you have a more specific need?

azime:
We need different timeouts depending on the API: feature, reverse or autocomplete.
Also, we will need to easily configure these timeouts without restarting ES

remi-dupre:
I would be fine with this change as we don't expose the API directly.

However, I agree that this is a potential security breach, and that doesn't seem ok for an open-source project that others may want to use? Maybe a good compromise would be to wrap it behind a feature flag and document it well?

pbench:
Hi Antoine ! 

> isn't it dangerous to have a way to bypass the timeout on the api? This could lead of an overload on ES no?

As far as I understand, this "timeout" instruct Elasticsearch to stop looking for results when the specified duration is attained. So it is a way to choose to use more computing time for a more "qualitative" result.

So if this timeout is too small, you risk not getting the "right" result. 
On the other hand setting high timeouts to a lot of queries may overload Elasticsearch.

Anyway, setting the timeout in the query is a feature that was present in ES2, and is used and needed at Kisio.

I you really worry about the risk of overloading Elasticsearch, I can cap the timeout to some (configurable) value. 
But what is a good max timeout ? 
What value do you use today to avoid overloading ?

pbench:
> I'd be okay for a global maximum timeout value (this could be the role of current `elasticsearch.timeout`), as long as this is configurable the value itself isn't really relevant, on our side we will have to fine tune it when everything is set up.

Allright, I'll use `elasticsearch.timeout` as both the maximum timeout, and the default timeout if not specified in the request.
We can always introduce two distincts config parameters later.

Thanks for the feedback !

libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
```diff
@@ -1022,16 +1023,21 @@ impl ElasticsearchStorage {
...
         query: Query,
         limit_result: i64,
+        timeout: Option<Duration>,
     ) -> Result<Vec<D>, Error>
     where
         D: DeserializeOwned + Send + Sync + 'static,
     {
         let indices = indices.iter().map(String::as_str).collect::<Vec<_>>();
+        let timeout = timeout
+            .map(|t| std::cmp::min(t, self.config.timeout)) // let's cap the timeout to self.config.timeout to prevent overloading elasticsearch with long requests
```
remi-dupre, line 1033: Maybe we should add a warning in the cases where `t` is overridden by the config?

remi-dupre (commented):
Thanks :+1:

remi-dupre (approved):
This looks good to me as is, thanks for adapting your PR :+1:

pbench:
Thanks for your swift feedback @remi-dupre !

azime (approved):

PR #666, opened 2021-12-15
activate no strict mode in arguments params

azime:


libs/mimir/src/adapters/primary/bragi/routes.rs
```diff
@@ -2,13 +2,15 @@ use geojson::{GeoJson, Geometry};
 use std::convert::Infallible;
 use tracing::instrument;
 use url::Url;
+
```
remi-dupre, line 5: This new newline makes me wonder if it was introduced by some IDE tooling?

remi-dupre (approved):
As I understand [documentation](https://docs.rs/serde_qs/latest/serde_qs/#strict-vs-non-strict-modes) this will only affect how keys are parsed, which are not likely to contain any `[`/`]`.

libs/mimir/src/adapters/primary/bragi/routes.rs
azime: it is a mistake :)

libs/mimir/src/adapters/primary/bragi/routes.rs
```diff
@@ -407,4 +410,25 @@ mod tests {
...
+            .unwrap()
+            .0
+            .types
+            .as_ref()
+            .unwrap()
+            .iter()
+            .zip([Type::Street, Type::House].iter())
+            .all(|(a, b)| *a == *b));
+        assert_eq!(resp.unwrap().0.q, String::from("rue hictor malot"));
+    }
```
remi-dupre, line 433: Also, this seems more readable :slightly_smiling_face: 

```suggestion
        let filter = forward_geocoder_get();
        let resp = warp::test::request()
            .path("/api/v1/autocomplete?q=rue+hictor+malot&type%5B%5D=street&type%5B%5D=house")
            .filter(&filter)
            .await
            .unwrap();
        assert_eq!(resp.0.types.unwrap(), [Type::Street, Type::House]);
        assert_eq!(resp.0.q, "rue hictor malot");
```

azime: :+1:

libs/mimir/src/adapters/primary/bragi/routes.rs
```diff
@@ -138,7 +139,8 @@ pub fn forward_geocoder_query(
     // warp::query cannot parse array parameters correctly, so we use serde_qs for that:
     warp::filters::query::raw()
         .and_then(|param: String| async move {
-            serde_qs::from_str::<ForwardGeocoderQuery>(&param).map_err(|_| {
+            let config = Config::new(3, false);
```
crocme10, line 142: When you put hardcoded values like that, I suggest you put a comment to explain why those values.... The reader may wonder why you put 3, and not 2, or 42

libs/mimir/src/adapters/primary/bragi/routes.rs
```diff
@@ -155,7 +157,8 @@ pub fn forward_geocoder_query(
 pub fn forward_geocoder_explain_query(
 ) -> impl Filter<Extract = (ForwardGeocoderExplainQuery,), Error = Rejection> + Copy {
     warp::filters::query::raw().and_then(|param: String| async move {
-        serde_qs::from_str::<ForwardGeocoderExplainQuery>(&param).map_err(|_| {
+        let config = Config::new(3, false);
```
crocme10, line 160: ditto

libs/mimir/src/adapters/primary/bragi/routes.rs
```diff
@@ -407,4 +410,16 @@ mod tests {
             "Invalid GeoJSON shape (missing geometry). cannot deserialize body"
         );
     }
+
+    #[tokio::test]
+    async fn should_correctly_extract_query_no_strict_mode() {
+        let filter = forward_geocoder_get();
+        let resp = warp::test::request()
+            .path("/api/v1/autocomplete?q=rue+hictor+malot&type%5B%5D=street&type%5B%5D=house")
```
crocme10, line 418: s/hictor/hector/

libs/mimir/src/adapters/primary/bragi/routes.rs
```diff
@@ -138,7 +139,8 @@ pub fn forward_geocoder_query(
     // warp::query cannot parse array parameters correctly, so we use serde_qs for that:
```
crocme10, line 139: that comment does not seem relevant anymore

crocme10: and if that serde_qs is not needed anymore, perhaps it should be removed from cargo.toml

remi-dupre: That `Config` struct comes from `serde_qs` to override its default behavior, so it still the same library under the hood.

However I would indeed argue that `Config` is too generic and we could explicitly use `serde_qs::Config::new(..., ...)` without the `use serde_qs::Config` to make it clearer, but I suppose one could also argue that it is more consistent to always import everything that is used in the namespace.

pbench (approved):

PR #667, opened 2021-12-17
Add option to define number of threads used by tokio runtime

AzHicham:


config/bragi/default.toml
```diff
@@ -1,4 +1,5 @@
 mode = "default"
+nbthreads = 2
```
remi-dupre, line 2: It'd be nice if this was optional to allow to stick with the default behavior (1 thread / logical core).

config/cosmogony2mimir/default.toml
```diff
@@ -1,4 +1,5 @@
 langs = [ "fr" ]
+nbthreads = 2
```
remi-dupre, line 2: This is not using snake_case because it would conflict with env override behavior, right?

I wonder if we could find a more satisfying solution? Maybe using camel case would be a fair compromise.

remi-dupre (commented):
What is the motivation behind this change ?

Depending on this the parameter may have an ambiguous name or not doing what intended. For example the address imports will spawn up to one extra blocking thread / core : https://github.com/CanalTP/mimirsbrunn/blob/master/src/addr_reader.rs#L84

config/bragi/default.toml
AzHicham: By default, tokio runtime use all available cpu cores, so we prefere to set a defined value to avoid overloading our servers. 
@azime are you ok to set nbthreads to 1 by default ?

pbench (approved):

config/bragi/default.toml
pbench: I guess that @remi-dupre wants to be able to keep the previous behavior : the binary detect the number of cores on the machine and spawn a runtime with that number of worker.
Maybe this can be acheived by having `nbthreads : Option<usize>` and using all cores when `None` is given.
However, I have no idea how an Option<> behave with the Config crate used for parsing config/env/cli settings

AzHicham: As suggested by @remi-dupre, we stick to the default behavior (1t/1c) 
nbthread parameters is now an Option<usize>

azime (approved):

config/bragi/default.toml
remi-dupre: Should we also leave this as default in the default config?

I'm not sure but here we'd have to to specify explicitly `-s nbthreads=""`?

AzHicham: As discussed with @azime , we prefer to have an explicit nbthread in config files for all xxxx2mimir bins, but not for bragi. 
Are you ok for this behavior ? Sorry for the inconvenience

remi-dupre: Well we can compose with that, `MIMIR_NBTHREADS` can be used to specify a global behavior anyway, it just looked as a specific env configuration instead of generic default config ^^'

Also, I'm not sure how you need this to behave but as I mentioned on my first comment some other threads will also be spawned during import process, and tokio's thread pool will actually have little to no impact on how much resources will be consumed as it mostly handles http requests and parsing (on our end ES's writing thread pool is the bottleneck).

AzHicham: Looking at https://github.com/CanalTP/mimirsbrunn/blob/master/src/addr_reader.rs#L84 
I think that N tasks will be spawned, but that's not a problem because the tokio runtime can use up to M threads (defined by the config or env variable). 
So tokio will run theses N tasks on a maximum of M threads

remi-dupre: The issue here is that those tasks are spawned via `spawn_blocking` which spawns a thread independent from tokio's async thread pool. While this is not exactly idiomatic, this is a cheap way to run concurrent CPU-heavy tasks. From https://docs.rs/tokio/1.15.0/tokio/task/fn.spawn_blocking.html:

> In general, issuing a blocking call or performing a lot of compute in a future without yielding is problematic, as it may prevent the executor from driving other futures forward. This function runs the provided closure on a thread dedicated to blocking operations. See the CPU-bound tasks and blocking code section for more information.
> 
> Tokio will spawn more blocking threads when they are requested through this function until the upper limit configured on the Builder is reached. After reaching the upper limit, the tasks are put in a queue. The thread limit is very large by default, because spawn_blocking is often used for various kinds of IO operations that cannot be performed asynchronously. When you run CPU-bound code using spawn_blocking, you should keep this large upper limit in mind. When running many CPU-bound computations, a semaphore or some other synchronization primitive should be used to limit the number of computation executed in parallel. Specialized CPU-bound executors, such as rayon, may also be a good fit.

So there's actually a [`max_blocking_thread`](https://docs.rs/tokio/1.15.0/tokio/runtime/struct.Builder.html#method.max_blocking_threads) function in the runtime configuration but I don't think that it would be a good idea to override it with a small value (as it could block some I/O operations). A solution would be to propagate this nbthread setting or an equivalent to places where we spawn those tasks.

From there comes another question, do we spawn nbthread tasks in addition to tokio's threadpool or do we need granularity? Like an entire `threads` section in settings with `threads.asyncpool`, `threads.addresses`, etc...

AzHicham: Oh I see, I guess for the moment we can merge this PR as it is. 
And in the future we can create a more granular way to control the number of threads as you suggested

PR #671, opened 2021-12-22
Impl /features endpoint

AzHicham:


libs/mimir/src/adapters/primary/bragi/handlers.rs
```diff
@@ -239,13 +298,10 @@ fn build_es_indices_to_search(query: &ForwardGeocoderQuery) -> Vec<String> {
     }
     // no types specified, we search for all objects in all indices
```
crocme10, line 299: that comment does not seem relevant

libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
```diff
@@ -1080,6 +1084,55 @@ impl ElasticsearchStorage {
         }
     }
 
+    pub(super) async fn get_documents_by_id<D>(
```
crocme10, line 1087: It's probably worth documenting that you cannot use the API as it was meant to be (MGetParts) because you want to target multiple indexes, and the API won't let you.

libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
```diff
@@ -1080,6 +1084,55 @@ impl ElasticsearchStorage {
...
+                } else {
+                    t
+                }
+            }) // let's cap the timeout to self.config.timeout to prevent overloading elasticsearch with long requests
+            .unwrap_or(self.config.timeout);
+
+        let get = self.client.mget(MgetParts::None).request_timeout(timeout);
+
+        let response = match query {
+            Query::QueryString(_) => get.send().await.context(ElasticsearchClient {
```
crocme10, line 1112: Isn't it an error if you actually use a QueryString? Do you even want to send that query, or should you just straight return an error?

libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
```diff
@@ -1080,6 +1084,55 @@ impl ElasticsearchStorage {
...
+                details: format!("could not get document by id"),
+            })?,
+            Query::QueryDSL(json) => get.body(json).send().await.context(ElasticsearchClient {
+                details: format!("could not get document by id"),
```
crocme10, line 1116: This is not the correct error msg

libs/mimir/src/adapters/secondary/elasticsearch/models.rs
```diff
@@ -30,6 +30,26 @@ impl<D> ElasticsearchSearchResponse<D> {
...
+}
+
+#[derive(Deserialize)]
+pub struct ElasticsearchDocs<D> {
+    #[serde(rename = "_source")]
+    pub source: D,
+    pub found: bool,
+}
+
+impl<D> ElasticsearchGetResponse<D> {
```
crocme10, line 46: Shouldn't you implement the IntoIterator trait instead (https://doc.rust-lang.org/std/iter/trait.IntoIterator.html)

libs/mimir/src/domain/ports/primary/mod.rs
```diff
@@ -1,6 +1,7 @@
 pub mod configure_backend;
 pub mod explain_query;
 pub mod generate_index;
+pub mod get_documents;
```
crocme10, line 4: little nit pick... but i wonder if opening a new port is necessary:
- we have search 
- we have list
- and now we have get 

maybe this get is a bit like a list_by_id, or a search_by_id ?

libs/mimir/src/domain/ports/secondary/get.rs
```diff
@@ -0,0 +1,26 @@
...
+
+#[derive(Debug, Clone)]
+pub struct Parameters {
+    pub query: Query,
+    pub timeout: Option<Duration>,
+}
+
+#[derive(Debug, Snafu)]
+pub enum Error {
+    #[snafu(display("Document Retrieval Error: {}", source))]
```
crocme10, line 18: Maybe make the error message different

libs/mimir/src/domain/ports/primary/mod.rs
AzHicham: well .. using /get /mget API should be faster than doing a search

libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
AzHicham: any suggestions ?

libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
AzHicham: In fact /mget let you target multiple indexes, so I don't really understand what you mean ..

libs/mimir/src/adapters/secondary/elasticsearch/models.rs
AzHicham: I guess we could but I tried to be coherent with ElasticsearchSearchResponse -> into_hits

libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
crocme10: I just say you should document (write a comment) to explain why you use the API differently from the expected way it is intendede to be used

libs/mimir/src/adapters/secondary/elasticsearch/models.rs
crocme10: You could try to change SearchResponse as well,... It's only a suggestion to make the code more idiomatic. But then again into_hits is very readable.

libs/mimir/src/domain/ports/primary/mod.rs
crocme10: it's not so much a matter of speed, but to make it intuitive for someone reading the code, or writing new code. Do as you wish

azime (approved):

PR #672, opened 2021-12-22
Reproduce bragi prefix autocomplete search

fatal69100:
Reproduce in ElasticSearch7 what Bragi, with ES2, was doing for the `prefix` autocomplete part.

### DIFF
--- a/libs/mimir/src/adapters/primary/bragi/autocomplete.rs
+++ b/libs/mimir/src/adapters/primary/bragi/autocomplete.rs
@@ -39,7 +39,7 @@ pub fn build_query(
     })
 }
 
-fn build_string_query(q: &str, settings: &StringQuery) -> serde_json::Value {
+fn build_autocomplete_should_query(q: &str, _lang: &str, settings: &StringQuery) -> serde_json::Value {
     json!({
         "bool": {
             "boost": settings.global,
@@ -47,23 +47,39 @@ fn build_string_query(q: &str, settings: &StringQuery) -> serde_json::Value {
                 {
                     "multi_match": {
                         "query": q,
-                        "fields": ["name"],
+                        "fields": ["name", format!("names.{}", lang)],
                         "boost": settings.boosts.name
                     }
                 },
                 {
                     "multi_match": {
                         "query": q,
-                        "fields": ["label"],
+                        "fields": ["label", format!("label.{}", lang)],
                         "boost": settings.boosts.label
                     }
                 },
                 {
                     "multi_match": {
                         "query": q,
-                        "fields": ["label.prefix"],
+                        "fields": ["label.prefix", format!("label.prefix.{}", lang)],
                         "boost": settings.boosts.label_prefix
                     }
+                },
+                {
+                    "match": {
+                        "zip_codes": {
+                        "query": q,
+                    }
+                    }
+                    "boost": settings.boosts.zip_codes
+                },
+                {
+                    "match": {
+                        "house_number": {,
+                         "query": q,
+                    }
+                    }
+                   "boost": settings.boosts.house_number
                 }
             ]
         }
--- a/libs/mimir/src/adapters/primary/bragi/autocomplete.rs
+++ b/libs/mimir/src/adapters/primary/bragi/autocomplete.rs
@@ -25,8 +25,6 @@ pub struct Filters {
 
 pub fn build_query(
     q: &str,
-    _filters: Filters,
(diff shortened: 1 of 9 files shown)


crocme10:
I can't seem to be able to comment in the code.... But, I see sometimes the parameter '_lang' is being used, instead of 'lang'

libs/mimir/src/adapters/primary/bragi/handlers.rs
```diff
@@ -41,8 +41,9 @@ where
     let timeout = params.timeout;
     let es_indices_to_search_in =
         build_es_indices_to_search(&params.types, &params.pt_dataset, &params.poi_dataset);
+    let lang = params.lang.clone().unwrap_or_else(|| "fr".to_string());
```
AzHicham, line 44: Maybe instead of setting a default value here, we could use #[serde(default = "default_lang")] in api.rs

libs/mimir/src/adapters/primary/bragi/autocomplete.rs
```diff
@@ -39,31 +37,47 @@ pub fn build_query(
...
+fn build_autocomplete_should_query(q: &str, lang: &str, settings: &StringQuery) -> serde_json::Value {
     json!({
         "bool": {
             "boost": settings.global,
             "should": [
                 {
                     "multi_match": {
                         "query": q,
-                        "fields": ["name"],
+                        "fields": ["name", format!("names.{}", lang)],
```
azime, line 48: You don't need to change the mapping ?

azime (approved):

libs/mimir/src/adapters/primary/bragi/autocomplete.rs
fatal69100: https://github.com/CanalTP/mimirsbrunn/blob/da27c5c35f3e8f3d41bdef5aec3298c1b6ae2fc7/config/elasticsearch/templates/components/mimir-dynamic-mappings.json#L7 : I think this mapping can handle names field with different languages

azime: :+1:

libs/mimir/src/adapters/primary/common/dsl.rs
```diff
@@ -94,12 +104,117 @@ fn build_boosts(
...
+    zone_types: Option<Vec<String>>,
+) -> Vec<serde_json::Value> {
+    let mut filters: Vec<Option<serde_json::Value>> = Vec::new();
+    let geoshape_filter = shape.map(|(geometry, scope)| build_shape_query(geometry, scope));
+    filters.push(geoshape_filter);
+    let poi_types_filter = poi_types.map(build_poi_types_filter);
+    filters.push(poi_types_filter);
+    let zone_types_filter = zone_types.map(build_zone_types_filter);
+    filters.push(zone_types_filter);
+    filters.into_iter().flatten().collect()
```
pbench, line 122: I think you are creating two vectors here : the first one is `filters` inside the function, and the second one is created by `collect` and returned.
I think you can avoid creating a temporary vector inside the function as follows :

```suggestion
    let mut filters: Vec<serde_json::Value> = Vec::new();
    if let Some(geoshape_filter) = shape.map(|(geometry, scope)| build_shape_query(geometry, scope)) {
        filters.push(geoshape_filter);
    };
    if let Some(poi_types_filter) = poi_types.map(build_poi_types_filter) {
        filters.push(poi_types_filter);
    }
    if let Some(zone_types_filter) = zone_types.map(build_zone_types_filter) {
        filters.push(zone_types_filter);
    }
    filters
```

libs/mimir/src/adapters/primary/common/dsl.rs
```diff
@@ -8,80 +9,89 @@ use super::{filters, settings};
 pub fn build_query(
     q: &str,
     filters: filters::Filters,
-    _langs: &[&str],
+    lang: String,
```
pbench, line 12: why not a `&str` here ?

pbench (approved):
Minor suggestions, otherwise it looks good !

azime (approved):

libs/mimir/src/adapters/primary/common/dsl.rs
fatal69100: I was due to the `lang` parameter type in `QueryParams`, but I take your suggestion into account, It's more corfortable

PR #674, opened 2021-12-23
Fix shape & shape_scope filters

azime:


libs/mimir/src/adapters/primary/bragi/api.rs
```diff
@@ -32,7 +32,7 @@ pub struct ForwardGeocoderQuery {
     pub q: String,
     pub lat: Option<f32>,
     pub lon: Option<f32>,
-    pub shape_scope: Option<Vec<Type>>,
+    pub shape_scope: Option<Vec<PlaceDocType>>,
```
crocme10, line 35: why use placedoctype here? The line below you use type? So you risk having multilple ways to name the same thing

libs/mimir/src/adapters/primary/common/dsl.rs
```diff
@@ -216,14 +215,14 @@ pub fn build_shape_query(shape: Geometry, scope: Vec<String>) -> serde_json::Val
...
                     "must": {
                         "terms": {
-                            "_source.type": scope
+                            "type": scope
                         }
                     },
                     "filter": {
                         "geo_shape": {
-                            "location": {
+                            "approx_coord": {
```
crocme10, line 223: why approx_coord and not coord?

libs/places/src/lib.rs
```diff
@@ -164,13 +164,18 @@ impl Place {
 }
 
 // This is a bit of a kludge to a get a string version for the doc_type.
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(PartialEq, Copy, Clone, Debug, Deserialize, Serialize)]
 #[serde(rename_all = "lowercase")]
 pub enum PlaceDocType {
+    #[serde(rename = "admin")]
```
crocme10, line 170: don't understand why you have to rename everything when there is a rename_all = lowercase?

config/elasticsearch/templates/components/mimir-base.json
```diff
@@ -65,7 +65,8 @@
             "enabled": false
           },
           "approx_coord": {
-            "enabled": false
+            "type": "geo_shape",
```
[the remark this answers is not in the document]
azime: I did not rename coord to approx_coord

libs/mimir/src/adapters/primary/bragi/api.rs
azime: Same [Bragi2](https://github.com/CanalTP/mimirsbrunn/blob/da72c2fac5887d3c6fbbc2b4d69201ecb9a95319/libs/bragi/src/routes/autocomplete.rs#L94)

AzHicham (approved):

PR #683, opened 2021-12-27
Display more informations bulk insert

azime:
Example:

**PUT test_coord**
``` json
{
  "mappings": {
    "properties": {
          "approx_coord": {
            "type": "geo_shape",
            "precision": "5m"
          }
    }
  }
}
```
Post invalid geo_shape

**POST _bulk**
``` json
{ "index" : { "_index" : "test_coord", "_id" : "1" } }
{"text": "Geopoint as an object","approx_coord": {    "type" : "point",    "coordinates" : [-7703653, 38.897676]}}
```

result:

``` json
{
  "took" : 0,
  "errors" : true,
  "items" : [
    {
      "index" : {
        "_index" : "test_coord",
        "_type" : "_doc",
        "_id" : "1",
        "status" : 400,
        "error" : {
          "type" : "mapper_parsing_exception",
          "reason" : "failed to parse",
          "caused_by" : {
            "type" : "invalid_shape_exception",
            "reason" : "Bad X value -7703653.0 is not in boundary Rect(minX=-180.0,maxX=180.0,minY=-90.0,maxY=90.0)"
          }
        }
      }
    }
  ]
}
```

Result in log file:

image

### DIFF
--- a/libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
+++ b/libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
@@ -603,12 +603,13 @@ impl ElasticsearchStorage {
             let es_response: ElasticsearchBulkResponse =
                 resp.json().await.context(ElasticsearchDeserialization)?;
             es_response.items.into_iter().try_for_each(|item| {
-                let result = item.inner().result.map_err(|err| {
+                let inner = item.inner();
+                let result = inner.result.map_err(|err| {
                     let reason = err
                         .caused_by
                         .map_or("".to_string(), |caused_by| caused_by.reason);
                     Error::NotCreated {
-                        details: format!("{}, {}", err.reason, reason),
+                        details: format!("Object id {}, Error: {}, {}", inner.id, err.reason, reason),
                     }
                 })?;
 
--- a/libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
+++ b/libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
@@ -602,10 +602,15 @@ impl ElasticsearchStorage {
         } else {
             let es_response: ElasticsearchBulkResponse =
                 resp.json().await.context(ElasticsearchDeserialization)?;
-
             es_response.items.into_iter().try_for_each(|item| {
-                let result = item.inner().result.map_err(|err| Error::NotCreated {
-                    details: err.reason,
+                let result = item.inner().result.map_err(|err| {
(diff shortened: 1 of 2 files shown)


libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
```diff
@@ -602,10 +602,15 @@ impl ElasticsearchStorage {
...
                 resp.json().await.context(ElasticsearchDeserialization)?;
-
             es_response.items.into_iter().try_for_each(|item| {
-                let result = item.inner().result.map_err(|err| Error::NotCreated {
-                    details: err.reason,
+                let result = item.inner().result.map_err(|err| {
+                    let reason = match err.caused_by {
+                        Some(caused_by) => caused_by.reason,
+                        _ => "".to_string(),
+                    };
```
AzHicham, line 610:
```suggestion
                    let reason = err.caused_by.map_or("".to_string(), |caused_by| caused_by.reason);
```

AzHicham (approved):

pbench (approved):

libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
azime: :+1:

libs/mimir/src/adapters/secondary/elasticsearch/models.rs
```diff
@@ -86,6 +86,13 @@ pub enum ElasticsearchBulkResult {
     Deleted,
 }
 
+#[derive(Debug, Eq, PartialEq, Deserialize)]
+pub struct ElasticSearchBulkErrorCausedBy {
```
crocme10, line 90: s/ElasticSearch/Elasticsearch/

libs/mimir/src/adapters/secondary/elasticsearch/models.rs
```diff
@@ -86,6 +86,13 @@ pub enum ElasticsearchBulkResult {
...
+    #[serde(rename = "type")]
+    pub caused_by_type: String,
```
crocme10, line 92: the name of the variable is ....type, so I would expect something other than a String, like an enum for example.

azime: like type [here](https://github.com/CanalTP/mimirsbrunn/blob/master/libs/mimir/src/adapters/secondary/elasticsearch/models.rs#L92)

PR #685, opened 2021-12-28
Remove "boundary" field from Elasticsearch's responses to bragi

pbench:
Since the "boundary" field is quite heavy, and not used in bragi anyway, we ask Elasticsearch to *not* include it in its responses.

The  "boundary" field is still included by Elasticsearch in the queries made in *2mimir

### DIFF
--- a/libs/mimir/src/adapters/primary/bragi/handlers.rs
+++ b/libs/mimir/src/adapters/primary/bragi/handlers.rs
@@ -200,6 +200,12 @@ where
         build_es_indices_to_search(&None, &params.pt_dataset, &params.poi_dataset);
     let dsl = dsl::build_features_query(&es_indices_to_search_in, &doc_id);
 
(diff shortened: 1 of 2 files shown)


azime (approved):

AzHicham (changes requested):
I thinks you could also add "_source_excludes" : ["boundary] in build_features_query fn

pbench:
> I thinks you could also add "_source_excludes" : ["boundary] in build_features_query fn eg /_mget API https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-multi-get.html
Thanks ! 
Done in https://github.com/CanalTP/mimirsbrunn/pull/685/commits/d073245207445c3bc62d21ed108aa54eb9b60dd5

AzHicham (approved):

fatal69100 (approved):

PR #686, opened 2021-12-28
Add fuzzy query

fatal69100:
To reproduce the previous behavior : -> Request elasticsearch in `PREFIX` mode first (fast and confident)
-> If no results found: request in `FUZZY` mode (with ngrams)

### DIFF
--- a/libs/mimir/src/adapters/primary/bragi/handlers.rs
+++ b/libs/mimir/src/adapters/primary/bragi/handlers.rs
@@ -72,7 +72,7 @@ where
 
     match client
         .search_documents(
-            es_indices_to_search_in,
+            es_indices_to_search_in.clone(),
             Query::QueryDSL(dsl),
             filters.limit,
             timeout,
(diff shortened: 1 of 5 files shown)


libs/mimir/src/adapters/primary/bragi/handlers.rs
```diff
@@ -44,30 +48,89 @@ pub struct InternalError {
 impl Reject for InternalError {}
 
 #[instrument(skip(client, settings))]
-pub async fn forward_geocoder<S>(
+pub async fn forward_geocoder(
     params: ForwardGeocoderQuery,
     geometry: Option<Geometry>,
-    client: S,
+    client: ElasticsearchStorage,
```
crocme10, line 54: Why do you use ElasticsearchStorage instead of a generic?

fatal69100: It was easier for me to understand and to develop this feature with explicit code. I am not sure, but do we use another client than ElasticsearchStorage ?

libs/mimir/src/adapters/primary/bragi/handlers.rs
```diff
@@ -44,30 +48,89 @@ pub struct InternalError {
...
+                }
+            } else {
+                Ok(with_status(json(&resp), StatusCode::OK))
+            }
+        }
+        Err(err) => Err(warp::reject::custom(InternalError {
+            reason: InternalErrorReason::ObjectNotFoundError,
+            info: err.to_string(),
+        })),
+    }
```
pbench, line 121: You don't run a fuzzy query when :

- the prefix query failed
- the prefix query succeeded but found nothing.

Is that on purpose ? I would expect a fuzzy query to be launched when one of these things happen.

pbench:
Related to https://github.com/CanalTP/mimirsbrunn/pull/688 
I guess we will need to have different timeouts for prefix/fuzzy queries.
This can be done in a subsequent PR

libs/mimir/src/adapters/primary/bragi/handlers.rs
pbench: Ok, this is the same behavior as with es 2 https://github.com/CanalTP/mimirsbrunn/blob/da72c2fac5887d3c6fbbc2b4d69201ecb9a95319/libs/bragi/src/query.rs#L690

PR #690, opened 2021-12-31
Fuzzy query

AzHicham:
Based on PR #686

### DIFF
--- a/libs/mimir/src/adapters/primary/common/dsl.rs
+++ b/libs/mimir/src/adapters/primary/common/dsl.rs
@@ -1,7 +1,6 @@
 use crate::adapters::primary::common::settings::{BuildWeight, ImportanceQueryBoosts, StringQuery};
 use geojson::Geometry;
 use serde_json::json;
-use std::collections::BTreeMap;
 
 use super::coord::Coord;
 use super::{filters, settings};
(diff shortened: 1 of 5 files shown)


libs/mimir/src/adapters/primary/common/dsl.rs
```diff
@@ -199,19 +254,44 @@ fn build_house_number_condition(q: &str) -> serde_json::Value {
...
+        // it's supposed to be able to manage cases BOTH missspelt one-word
+        // www.elastic.co/guide/en/elasticsearch/guide/current/match-multi-word.html#match-precision
+        // requests AND very long requests.
+        // Missspelt one-word request:
+        //     Vaureaaal (instead of Vaureal)
+        // Very long requests:
+        //     Caisse Primaire d'Assurance Maladie de Haute Garonne, 33 Rue du Lot, 31100 Toulouse
+        QueryType::FUZZY => json!({
+            "match": {
+                "label": {
```
fatal69100, line 288: I take a look to the previous version to understand the difference between `label` and `full_label` fields in ES2 mapping : `full_label` = `label` + `zip_codes`. I think only label field is great for a first version, and we just keep in mind that we can improve the fuzzy query here ?

fatal69100 (approved):
Thank you

PR #693, opened 2022-01-03
Fix fuzzy queries

fatal69100:
- [x] Remove standard analyzer from search_as_you_type
- [x] Filter with MinimumShouldMatch condition on a ngram subfield
- [  ] Test this PR


libs/mimir/src/adapters/primary/common/dsl.rs
```diff
@@ -71,16 +71,17 @@ fn build_string_query(
     ));
 
     if let QueryType::FUZZY = query_type {
+        string_should = Vec::new();
```
AzHicham, line 74: Here you are discarding all previous push into string_should, may be you could refactor this code with if let FUZZY and else if let PREFIX ?

fatal69100: yes, sorry I forget to put the PR in draft

AzHicham (approved):

fatal69100:
As discussed, `search_as-you_type` is give up. Closed by https://github.com/CanalTP/mimirsbrunn/pull/698

PR #697, opened 2022-01-04
Fix empty answer autocomplete

fatal69100:
On autocomplete endpoint, when no answer is found, return :
```
type	"FeatureCollection"
geocoding	
version	"0.1.0"
query	""
features	[]
```
Reverse endpoint should already return this answer


remi-dupre (approved):
That's indeed more consistent with mimir1's behavior.

AzHicham (approved):

pbench (approved):

PR #702, opened 2022-01-05
Reproduce Bragi2 autocomplete

fatal69100:
- [x] Filter on `full_label` field 
- [x] Add build weight function
- [x] Update function score to ES7 format

Geocodeur test works better + performance enhanced compared to search_as_you_type + autocomplete working better (on very few manual tests)

### DIFF
--- a/libs/mimir/src/adapters/primary/common/dsl.rs
+++ b/libs/mimir/src/adapters/primary/common/dsl.rs
@@ -254,7 +254,7 @@ fn build_matching_condition(q: &str, query_type: QueryType) -> serde_json::Value
             "match": {
                 "full_label.ngram": {
                     "query": q,
-                    "minimum_should_match": "1<-1 3<-2 9<-4 20<25"
+                    "minimum_should_match": "1<-1 3<-2 9<-4 20<25%"
                 }
             }
         }),
--- a/libs/mimir/src/adapters/primary/common/dsl.rs
+++ b/libs/mimir/src/adapters/primary/common/dsl.rs
@@ -1,4 +1,6 @@
-use crate::adapters::primary::common::settings::{BuildWeight, ImportanceQueryBoosts, StringQuery};
+use crate::adapters::primary::common::settings::{
+    BuildWeight, ImportanceQueryBoosts, StringQuery, Types,
+};
 use geojson::Geometry;
 use serde_json::json;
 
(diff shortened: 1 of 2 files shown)


azime (approved):

libs/mimir/src/adapters/primary/common/dsl.rs
```diff
@@ -246,7 +252,7 @@ fn build_matching_condition(q: &str, query_type: QueryType) -> serde_json::Value
         //     Caisse Primaire d'Assurance Maladie de Haute Garonne, 33 Rue du Lot, 31100 Toulouse
         QueryType::FUZZY => json!({
             "match": {
-                "label": {
+                "full_label.ngram": {
                     "query": q,
                     "minimum_should_match": "1<-1 3<-2 9<-4 20<25"
```
pbench, line 257: Not really related, but
```suggestion
                    "minimum_should_match": "1<-1 3<-2 9<-4 20<25%"
```
I think this is a 25% and not 25. A 25 does not really make sense when I read [the doc](https://www.elastic.co/guide/en/elasticsearch/reference/7.16/query-dsl-minimum-should-match.html).
And there is a subtle f64 [in bragi2](https://github.com/CanalTP/mimirsbrunn/blob/da72c2fac5887d3c6fbbc2b4d69201ecb9a95319/libs/bragi/src/query.rs#L373)

pbench (approved):

libs/mimir/src/adapters/primary/common/dsl.rs
fatal69100: thank you, indeed it was a nonsense

libs/mimir/src/adapters/primary/common/dsl.rs
```diff
@@ -246,36 +252,31 @@ fn build_matching_condition(q: &str, query_type: QueryType) -> serde_json::Value
         //     Caisse Primaire d'Assurance Maladie de Haute Garonne, 33 Rue du Lot, 31100 Toulouse
         QueryType::FUZZY => json!({
             "match": {
-                "label": {
+                "full_label.ngram": {
                     "query": q,
-                    "minimum_should_match": "1<-1 3<-2 9<-4 20<25"
+                    "minimum_should_match": "1<-1 3<-2 9<-4 20<25%"
```
remi-dupre, line 257: If I understand the documentation `1<-1` means that if there is a single close, then it is optional, right?

If I'm not wrong `3<-2` seems very permissive too as it allows 67% of mismatches?

pbench: > If I understand the documentation 1<-1 means that if there is a single close, then it is optional, right?

I understand the same thing.
It seems strange, for me it means that "if there is only one clause, then I accept everything"

However, I do not really understand how (and how many) clauses are constructed from the query.

remi-dupre: I thought this was related to how the 3grams were handled but it actually defines the behaviors over words, my bad :grimacing: 

We had the same behavior with previous version of mimir with a little piece of explaination: https://github.com/CanalTP/mimirsbrunn/blob/v1.24.0/libs/bragi/src/query.rs#L358-L367

PR #706, opened 2022-01-06
Fix autocomplete with coord

azime:
Reproduce the consideration of coordinates as in [Bragi2](https://github.com/CanalTP/mimirsbrunn/blob/da72c2fac5887d3c6fbbc2b4d69201ecb9a95319/libs/bragi/src/query.rs#L117)

### DIFF
--- a/libs/mimir/src/adapters/primary/common/dsl.rs
+++ b/libs/mimir/src/adapters/primary/common/dsl.rs
@@ -121,10 +121,15 @@ fn build_boosts(
         decay.offset = proximity.offset;
         decay.decay = proximity.decay;
     }
+    let weight_boost = match query_type {
+        QueryType::PREFIX => settings.importance_query.proximity.weight,
+        _ => settings.importance_query.proximity.weight_fuzzy,
+    };
+
     let proximity_boost = filters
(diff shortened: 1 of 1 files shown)


libs/mimir/src/adapters/primary/common/dsl.rs
```diff
@@ -121,10 +121,15 @@ fn build_boosts(
         decay.offset = proximity.offset;
         decay.decay = proximity.decay;
     }
+    let weight_boost = match query_type {
+        QueryType::PREFIX => settings.importance_query.proximity.weight,
+        _ => settings.importance_query.proximity.weight_fuzzy,
```
pbench, line 126:
```suggestion
        QueryType::FUZZY => settings.importance_query.proximity.weight_fuzzy,
```
So we won't forget about this if a new QueryType is added

pbench (approved):

AzHicham (approved):

libs/mimir/src/adapters/primary/common/dsl.rs
azime: :+1:

PR #708, opened 2022-01-06
Fix nb_threads xxx2mimir

AzHicham:
We just removed all confusing 'buffered' & 'buffer_unordered' functions because tasks were not running asynchronously !!! 
To run async tasks with  'buffered' & 'buffer_unordered' functions  please use tokio::spawn. 
The number of function running in // is up to tokio runtime (cf config nb_threads)

### DIFF
--- a/src/bin/osm2mimir.rs
+++ b/src/bin/osm2mimir.rs
@@ -70,8 +70,6 @@ fn main() -> Result<(), Error> {
     }
 }
 
-const POI_REVERSE_GEOCODING_CONCURRENCY: usize = 8;
-
 async fn run(
     opts: settings::Opts,
     settings: settings::Settings,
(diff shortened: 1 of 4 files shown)


src/bin/osm2mimir.rs
```diff
@@ -153,13 +151,14 @@ async fn import_pois(
     client: &ElasticsearchStorage,
     config: &ContainerConfig,
 ) -> Result<(), Error> {
+    // This function rely on AdminGeoFinder::get_objs_and_deps
+    // which use all available cpu/cores to decode osm file and cannot be limited by tokio runtime
```
pbench, line 155: :cry:

remi-dupre:
> To run async tasks with 'buffered' & 'buffer_unordered' functions please use tokio::spawn.

I initially did it that way to ensure we won't affect asynchronous tasks (mainly interaction with ES), have you been able to test that it will not be an issue for performance, ie. by blocking async threads for long time (a few ms may already be an issue :/)

remi-dupre (approved):

PR #717, opened 2022-01-13
Add physical mode weights for stops objects

azime:


src/stops.rs
```diff
@@ -93,6 +93,50 @@ pub fn initialize_weights<'a, It, S: ::std::hash::BuildHasher>(
...
+    admin_weight = admin_weight * 1024.0 + 1.0;
+    admin_weight = admin_weight.log10();
+
+    let mut result = Vec::new();
+    if let Some(ph_weight) = physical_mode_weight {
+        result = stop
+            .physical_modes
+            .iter()
+            .map(|mode| {
+                let pm_w = ph_weight.iter().find(|&md| md.id == mode.id);
```
pbench, line 118: Instead of iterating through the vector `physical_mode_weight` for ALL stops, it may be more efficient to :

- transform the vector `physical_mode_weight` into an HashMap<String, f64> which maps the physical_mode_id to the weight 
- pass this hashmap to `make_weight` instead of the vector
- use `hashmap.get(mode.id)` here

azime: :+1:

src/stops.rs
```diff
@@ -183,35 +228,34 @@ pub async fn index_ntfs(
...
+    info!("Build stops weight by physical modes and city population");
+    let md_weight_hash_map: Option<HashMap<String, f64>> = match physical_mode_weight {
+        Some(modes) => Some(
+            modes
+                .iter()
+                .map(|mode| (mode.id.to_string(), mode.weight as f64))
+                .collect::<HashMap<String, f64>>(),
+        ),
+        _ => None,
+    };
```
pbench, line 254:
```suggestion
    let md_weight_hash_map: HashMap<String, f64> = match physical_mode_weight {
        Some(modes) => Some(
            modes
                .iter()
                .map(|mode| (mode.id.to_string(), mode.weight as f64))
                .collect::<HashMap<String, f64>>(),
        ),
        _ => HashMap::new(),
    };
```

src/stops.rs
```diff
@@ -93,6 +93,50 @@ pub fn initialize_weights<'a, It, S: ::std::hash::BuildHasher>(
...
+                            "Physical mode, id: {} name: {}, not found in mimir config.",
+                            mode.id, mode.name
+                        );
+                        0.0
+                    }
+                }
+            })
+            .filter(|weight| !weight.is_nan())
+            .collect();
+    }
```
pbench, line 132:
```suggestion
pub fn make_weight(stop: &mut Stop, physical_mode_weights: &HashMap<String, f64>) {
    // Admin weight
    let mut admin_weight = stop
        .administrative_regions
        .iter()
        .filter(|adm| adm.is_city())
        .map(|adm| adm.weight)
        .next()
        .unwrap_or(0.0);
    // FIXME: 1024, automagic!
    // It's a factor used to bring the stop weight and the admin weight in the same order of
    // magnitude...
    // We then use a log to compress the distance between low admin weight and high ones.
    admin_weight = admin_weight * 1024.0 + 1.0;
    admin_weight = admin_weight.log10();

    let mut result = Vec::new();
    
        result = stop
            .physical_modes
            .iter()
            .map(|mode| {
                let pm_w = physical_mode_weights.get(&*mode.id);
                match pm_w {
                    Some(value) => *value,
                    _ => {
                        warn!(
                            "Physical mode, id: {} name: {}, not found in mimir config.",
                            mode.id, mode.name
                        );
                        0.0
                    }
                }
            })
            .filter(|weight| !weight.is_nan())
            .collect();
    
```

pbench (approved):
Some suggestion to replace the Option<HashMap> with a plain HashMap (which becomes empty when no data is provided).

PR #732, opened 2022-03-08
forcemerge default behavior and timeout

remi-dupre:
Performing a force merge after the creation of an index seems to be a free performance improvement, however with its current defaults the operation can be very long and can lead to have the http client hanging forever.

With this change, this will leave the default ES behavior (merges indexes with more than 10% deleted documents I think) and allows to tune the timeout of the ES request while still being able to completely disable the forcemerge or tune the `max_number_segments`.


libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
```diff
@@ -828,27 +828,44 @@ impl ElasticsearchStorage {
...
+        let response = request.send().await;
+
+        // The forcemerge operation can be very long if a large number of segments have to be
+        // merged, in such a case the user may set `allow_timeout` to true in order to let the
+        // operation run in background.
+        if allow_timeout && matches!(&response, Err(err) if err.is_timeout()) {
+            warn!(
+                "forcemerge query timeout after {:?} on indices {}, it will continue running in background",
+                timeout,
+                indices.iter().copied().collect::<Vec<_>>().join(", ")
```
woshilapin, line 858: How about this instead (and the two other occurences below)?
```suggestion
                indices.join(", ")
```

woshilapin (approved):
I'm pretty new to the project, so please take everything with a pinch of salt :wink:

libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
remi-dupre: :exploding_head: 
I don't know how I ended up with this, probably started modifying the iterator and then managed to forget that I was operating on a slice in a first place.

libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
```diff
@@ -828,27 +828,44 @@ impl ElasticsearchStorage {
     pub(super) async fn force_merge(
         &self,
         indices: &[&str],
-        max_num_segments: i64,
+        max_num_segments: Option<i64>,
+        timeout: Duration,
+        allow_timeout: bool,
```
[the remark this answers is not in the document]
remi-dupre: That's somewhat simpler by passing `&ElasticsearchStorageForceMergeConfig`, I was afraid it wouldn't fit the original ambitions of strict hexagonal architecture for mimirsbrunn but I don't have any strong opinion either. So let's get this quick simplification :)

PR #733, opened 2022-03-10
elasticsearch internal: implement a backoff for bulk operations

remi-dupre:
Add a naïve exponential backoff to bulk operations (inserts / updates) in
elasticsearch to improve robustness of such operations. This could help
tweaking parameters of import jobs without crashing when they put too
much load on the ES and will ease conflict with ES performing
maintenance operations on its nodes.


libs/mimir/src/utils/futures.rs
```diff
@@ -0,0 +1,39 @@
...
+    mut wait: Duration,
+) -> Result<T, E>
+where
+    F: Future<Output = Result<T, E>>,
+    T: fmt::Debug,
+    E: fmt::Display,
+{
+    let mut res = task().await;
+
+    while retries > 0 && res.is_err() {
```
woshilapin, line 25: Sadly, [we can't yet do the following](https://github.com/rust-lang/rust/issues/53667)...

```rust
    while retries > 0 && let Err(err) = res {
```

That would avoid the `.unwrap_err()` below... Another way would be something like the following (not necessarily shorter though).

```rust
while retries > 0 {
    if let Err(err) = res {
        // impl
    } else {
        break;
    }
}
```

(sorry, I'm always trying ways to avoid the `.unwrap` everywhere :smile:)

woshilapin (approved):

libs/mimir/src/utils/futures.rs
remi-dupre: Yeah I really missed the chained if assignment here.

I not a big fan of this extra nesting which makes it quite harder to understand what's happening, but I'll go with your solution, it also allows to removes the unecessary `T: Debug` requirement in the function signature.

libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
```diff
@@ -593,17 +594,22 @@ impl ElasticsearchStorage {
...
-            .bulk(BulkParts::Index(index.as_str()))
-            .request_timeout(self.config.timeout)
-            .body(chunk)
-            .send()
-            .await
-            .and_then(|res| res.error_for_status_code())
-            .context(ElasticsearchClientSnafu {
-                details: "cannot bulk insert",
-            })?;
+        let resp = with_backoff(
```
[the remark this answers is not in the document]
remi-dupre: Yeah, I also personally don't like how rustfmt doesn't put any indentation in "dot chains" after a multi-line expression :shrug:

PR #735, opened 2022-03-15
re-implement metric `bragi_elasticsearch_request_duration_seconds`

remi-dupre:
This was a metric that was computed for the ES2 branch and will be useful to debug were the time is spent during computation.


remi-dupre:
Replaced by https://github.com/CanalTP/mimirsbrunn/pull/739

PR #739, opened 2022-03-23
re-implement metric `bragi_elasticsearch_request_duration_seconds`

remi-dupre:
This was a metric that was computed for the ES2 branch and will be useful to debug were the time is spent during computation.

### DIFF
--- a/libs/mimir/src/adapters/primary/bragi/handlers.rs
+++ b/libs/mimir/src/adapters/primary/bragi/handlers.rs
@@ -2,12 +2,10 @@ use crate::adapters::primary::bragi::prometheus_handler;
 use geo::algorithm::haversine_distance::HaversineDistance;
 use geojson::Geometry;
 use std::time::Duration;
-use tracing::{debug, instrument};
-use warp::{
-    http::StatusCode,
-    reject::Reject,
-    reply::{json, with_status},
-};
+use tracing::instrument;
+use warp::http::StatusCode;
+use warp::reject::Reject;
+use warp::reply::{json, with_status};
 
 use crate::{
     adapters::primary::{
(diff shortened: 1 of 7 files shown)


libs/mimir/src/adapters/primary/bragi/handlers.rs
```diff
@@ -1,14 +1,15 @@
...
 use geo::algorithm::haversine_distance::HaversineDistance;
 use geojson::Geometry;
-use std::time::Duration;
-use tracing::{debug, instrument};
-use warp::{
-    http::StatusCode,
-    reject::Reject,
-    reply::{json, with_status},
-};
+use prometheus::{exponential_buckets, register_histogram_vec, HistogramVec};
```
woshilapin, line 5: This will cause problems since `promotheus` is an optional dependency. You can see the issue if you try to compile with the following command.

```
cargo check --package mimir --no-default-features
```

Maybe we can also try to add something in the CI to cover this kind of case?

woshilapin (changes requested):

libs/mimir/src/adapters/primary/bragi/handlers.rs
remi-dupre: Oh thanks, I forgot about that :+1:

remi-dupre: That should be fixed now :wink:

azime (approved):

libs/mimir/src/adapters/primary/bragi/handlers.rs
```diff
@@ -38,8 +37,12 @@ use crate::{
 use common::document::ContainerDocument;
 use places::{addr::Addr, admin::Admin, poi::Poi, stop::Stop, street::Street, Place};
 
+#[cfg(metrics)]
```
woshilapin, line 40: I don't believe this is correct. Did you try that metrics worked since you made this change? On my side, I tried to add the following to see if this syntax worked and had no compilation error.

```rust
#[cfg(metrics)]
std::compile_error("panic on metrics");
```

and nothing happened. I think you meant to write `#[cfg(feature = "metrics")]`.

woshilapin (approved):

PR #740, opened 2022-03-23
If configured, each indexing should update all components and templates

kadhikari:
In ES7, the function to update components and templates for each type of data (poi, ntfs, bano...) is separated and is ctlmimir.
Instead of executing ctlmimir each time to update components and templates if there is any modifications, this function is also called if configured while indexing any type of data.

To activate update of components and templates, add 'update_templates = true' in /config/xx2mimir/default.toml

Without any modification of configuration, it should work as before.

### DIFF
--- a/config/ntfs2mimir/default.toml
+++ b/config/ntfs2mimir/default.toml
@@ -1,4 +1,5 @@
 nb_threads = 2
+update_templates = true
 
 [container]
   name = "stop"
--- a/src/bin/bano2mimir.rs
+++ b/src/bin/bano2mimir.rs
@@ -95,7 +95,7 @@ async fn run(
 
     tracing::info!("Connected to elasticsearch.");
 
-    // Update all the template coponents and indexes
+    // Update all the template components and indexes
     if settings.update_templates {
         update_templates(&client, opts.config_dir).await?;
     }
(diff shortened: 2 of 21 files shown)


src/bin/bano2mimir.rs
```diff
@@ -92,6 +93,13 @@ async fn run(
         .context(ElasticsearchConnectionSnafu)
         .map_err(Box::new)?;
 
+    tracing::info!("Connected to elasticsearch.");
+
+    // Update all the template coponents and indexes
```
remi-dupre, line 98: A small typo that is repeated in most files :stuck_out_tongue: 

```suggestion
    // Update all the template components and indexes
```

kadhikari: :+1:

config/ntfs2mimir/default.toml
```diff
@@ -1,4 +1,5 @@
 nb_threads = 2
+update_templates = true
```
azime, line 2: This parameter is to be added in each config

kadhikari: :+1:

kadhikari: Done in all files

pbench (approved):

AzHicham (approved):

azime (approved):

PR #742, opened 2022-03-28
[Clippy] Code maintenance

AzHicham:
🤖 This is an automatic PR 🤖

Clippy suggests the following changes

### DIFF
--- a/libs/mimir/src/adapters/primary/bragi/handlers.rs
+++ b/libs/mimir/src/adapters/primary/bragi/handlers.rs
@@ -4,16 +4,20 @@ use geo::algorithm::haversine_distance::HaversineDistance;
 use geojson::Geometry;
 use serde::{Deserialize, Serialize};
 use tracing::instrument;
-use warp::http::StatusCode;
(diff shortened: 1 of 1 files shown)


fatal69100 (approved):

remi-dupre (approved):
Out of curiosity what version of clippy are you using ?

I remember finding related nightly lints related to modules imports, suggestion either factorization like here or either to expand everything to a single line (only grouping at the max depth).

I personally really don't like those big import trees, it reads too differently from what I'm used from other languages :'(

remi-dupre (approved):
Out of curiosity what version of clippy are you using ?

I remember finding related nightly lints related to modules imports, suggestion either factorization like here or either to expand everything to a single line (only grouping at the max depth).

I personally really don't like those big import trees, it reads too differently from what I'm used from other languages :'(

AzHicham:
> 

Hello, for import/use we are using a nightly feature of clippy, but if you prefer we can disable it !

azime (approved):

PR #746, opened 2022-04-05
doc updates: better support for dot-separated keys

remi-dupre:
For some reason when using a dot-sparated key chain to update a document
value, elasticsearch will attempt to update the mapping. Which conflicts
with `properties` which is flattened.

For example the following update operation fails:

```json
{
   "script": {
       "source": "ctx._source['properties.image'] = \"test\""
    }
}
```

> Could not dynamically add mapping for field [properties.image]. Existing mapping for [properties] must be of type object but found [flattened].

And the following one succeeds

```json
{
   "script": {
       "source": "ctx._source['properties']['image'] = \"test\""
    }
}
```

So the approach of this PR is to manually split key and subkeys and
generate the subtree for ES to update:

```json
{
    "doc": {
        "properties": {
            "image": "test"
        }
    }
}
```

EDIT: And it also gets rid of the use of `script` which could be problematic when there are hardcoded values inside (formerly the key)

### DIFF
--- a/libs/mimir/src/domain/model/update.rs
+++ b/libs/mimir/src/domain/model/update.rs
@@ -16,7 +16,7 @@ pub fn generate_document_parts(ops: Vec<UpdateOperation>) -> serde_json::Value {
             //   { "properties": { "review": "excellent" } }
             // And `ident` = "properties.image", `value` = "https://foo.jpg", then this
             // will update `result` with this value:
-            //   { "properties": { "review", "excellent", "image": "https://foo.jpg" } }
+            //   { "properties": { "review": "excellent", "image": "https://foo.jpg" } }
             UpdateOperation::Set { ident, value } => {
                 // Get a reference to the position in the JSON where the value must be
                 // inserted.
--- a/libs/mimir/src/domain/model/update.rs
+++ b/libs/mimir/src/domain/model/update.rs
@@ -1,5 +1,81 @@
+use serde_json::json;
+
 #[derive(Clone, Debug)]
 pub enum UpdateOperation {
     /// Update a field `ident` with given value
     Set { ident: String, value: String },
+}
+
+/// Given the list of operations to perform, generate the parts of the documents that must be
+/// updated.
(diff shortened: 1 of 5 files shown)


woshilapin (approved):

remi-dupre:
We need to update different values for a same document, and it turns out that sending separate updates operations for this doesn't seem to be a good idea. The different operations may be executed concurrently which results in an error:

> thread 'main' panicked at 'Error inserting chunk: Elasticsearch Response: Not Created: Object id ta:poi:1700060, Error: [ta:poi:1700060]: version conflict, required seqNo [2998], primary term [1]. current
 document has seqNo [2999] and primary term [1], ', /usr/local/cargo/git/checkouts/mimirsbrunn-a571f9500814a748/06c4934/libs/mimir/src/adapters/secondary/elasticsearch/internal.rs:579:47
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: exit status 101

I've update my PR to allow specifying multiple update operations to a single document and added a small extra unit test as this part of the code starts being quite complicated.

libs/mimir/src/domain/model/update.rs
```diff
@@ -1,5 +1,81 @@
...
+/// updated.
+pub fn generate_document_parts(ops: Vec<UpdateOperation>) -> serde_json::Value {
+    ops.into_iter().fold(json!({}), |mut result, op| {
+        match op {
+            // Adds the part of the document that must be updated to `result`. For
+            // example if at current iteration `result` has this value:
+            //   { "properties": { "review": "excellent" } }
+            // And `ident` = "properties.image", `value` = "https://foo.jpg", then this
+            // will update `result` with this value:
+            //   { "properties": { "review", "excellent", "image": "https://foo.jpg" } }
```
woshilapin, line 19:
```suggestion
            //   { "properties": { "review": "excellent", "image": "https://foo.jpg" } }
```

woshilapin (approved):

azime (approved):

PR #748, opened 2022-04-29
update to Rust 1.60 and new production build profile

remi-dupre:
Update to latest Rust version (which includes some convenient features
such as better "string interpolation").

I also took this as a chance to clean up Dockerfile:

 - removed stretch support, which was not really supported: no image had
   been released for 2 years, ie. since rust 1.45

 - based images of slim versions, which should spare a few bytes and
   reduce the risk for vulnerabilities

 - use a new *production* compilation profile (ft. Rust 1.57), which
   enables aggressive LTO and symbols stripping. This only makes a small
   difference and makes the images longer to build, but can definitely be
   worth it with our long-running binaries.


AzHicham (approved):

Cargo.toml
```diff
@@ -90,6 +90,12 @@ zip = { version = "0.5.13", default_features = false }
 [build-dependencies]
 json = "0.12"
 
+[profile.production]
+inherits = "release"
+lto = "fat"
+codegen-units = 1
+strip = "symbols"
```
pbench, line 97: In case of panic, does stripping symbols also remove backtrace information ? Like file/line/function where the panic occurs.
 
If this does remove this information (and stripping does not significantly reduce the binary size) I would prefer keeping the symbols.

remi-dupre: 😱 you're right

Screenshot_2022-06-20_17-34-36

I wasn't aware of that, and there isn't so much clue about it when looking for it, even https://github.com/rust-lang/cargo/issues/4122 is not exactly clear about it.

I'll reverse it, the inconvenience is not worth a few megabytes :/

remi-dupre:
This PR is opened from my personal fork, which I'll delete once this is merged, which explains why the CI can't connect to docker.

Cargo.toml
pbench: allright, this sounds good to me :+1:

docker/bragi/Dockerfile
```diff
@@ -1,27 +1,18 @@
-ARG RUST_VERSION="1.56"
+ARG RUST_VERSION="1.60"
 ARG DEBIAN_VERSION="buster"
 
-FROM rust:${RUST_VERSION}-${DEBIAN_VERSION} as builder
+FROM rust:${RUST_VERSION}-slim-${DEBIAN_VERSION} as builder
```
pbench, line 4: Note that this is the "builder" container, not the final one in which bragi is shipped.
The final one [is already "slim"](https://github.com/hove-io/mimirsbrunn/blob/804aeaa71ec670d216e8d9d0dfccb88d2e1ff850/docker/bragi/Dockerfile#L52).

remi-dupre: Indeed, I'll revert this for future convenience

docker/bragi/Dockerfile
```diff
@@ -1,27 +1,18 @@
...
-  apt-get update \
-    && apt-get install -y make libssl1.0-dev git \
+RUN apt-get update \
+    && apt-get install -y pkg-config make libsqlite3-dev libssl-dev git \
     && apt-get clean \
-    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*; \
-else \
-  echo "Unsupported debian version '$DEBIAN_VERSION'"; \
-fi
+    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*;
```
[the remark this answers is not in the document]
remi-dupre: I agree, simpler is better :+1:

docker/bragi/Dockerfile
```diff
@@ -39,37 +28,27 @@ COPY ./benches ./benches
...
-elif [ "${DEBIAN_VERSION}" = "stretch" ]; then \
-  apt-get update \
-    && apt-get install -y curl libcurl3 sqlite3 \
-    && apt-get clean \
-    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*; \
-else \
-  echo "Unsupported debian version '$DEBIAN_VERSION'"; \
-fi
+RUN apt-get update
+RUN apt-get install -y libcurl4 sqlite3
```
pbench, line 49:
```suggestion
RUN apt-get install -y curl libcurl4 sqlite3
```
we do need curl in the docker for implementing an healthcheck

docker/mimirsbrunn/Dockerfile
```diff
@@ -39,42 +28,32 @@ COPY ./src ./src
...
-elif [ "${DEBIAN_VERSION}" = "stretch" ]; then \
-  apt-get update \
-    && apt-get install -y libcurl3 sqlite3 \
-    && apt-get clean \
-    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*; \
-else \
-  echo "Unsupported debian version '$DEBIAN_VERSION'"; \
-fi
+RUN apt-get update
+RUN apt-get install -y libcurl4 sqlite3
```
pbench, line 54:
```suggestion
RUN apt-get install -y curl libcurl4 sqlite3
```

docker/bragi/Dockerfile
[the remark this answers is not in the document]
pbench: great, thanks !

pbench:
A last request to put back curl inside the docker, and I'll be happy to merge !

remi-dupre:
> A last request to put back curl inside the docker, and I'll be happy to merge !

Should be done ! :)

PR #750, opened 2022-05-03
WIP: simplification of Bragi

remi-dupre:
The main goal here is to make Bragi more maintainable, mostly through a
more straightforward use of Warp that does not require to maintain several
macros per endpoint.

I've also changed the `SearchDocuments` trait so that it doesn't enforce
a return type for the JSON result. This resolves several cases where we
had to convert to a `Value` and then convert to another type (which
neither performant nor optimal).


libs/mimir/src/adapters/primary/bragi/api.rs
```diff
@@ -236,82 +267,6 @@ pub struct StatusResponseBody {
     pub elasticsearch: ElasticsearchStatus,
 }
 
-/// This macro is used to define the forward_geocoder route.
```
fatal69100, line 239: :heart:

remi-dupre:
Closed in favor of https://github.com/hove-io/mimirsbrunn/pull/768 (simpler to maintain from Qwant/mimirsbrunn)

PR #757, opened 2022-06-08
feat: always refresh after force_merge

woshilapin:
One of the breaking change going from forward after ElasticSearch version 7 is that `force_merge` doesn't automatically runs `refresh`. Since our `force_merge` only happens on the publication of a new index, we can explicitly runs the `refresh` like advised.


libs/mimir/src/adapters/secondary/elasticsearch/storage.rs
```diff
@@ -128,6 +128,7 @@ impl<'s> Storage<'s> for ElasticsearchStorage {
         index: Index,
         visibility: ContainerVisibility,
     ) -> Result<(), StorageError> {
+        // FIXME Why do we need to refresh BEFORE publish?
```
remi-dupre, line 131: We refresh the index and the update aliases pointing to it, so I think this is fine?

libs/mimir/src/adapters/secondary/elasticsearch/storage.rs
```diff
@@ -182,6 +183,12 @@ impl<'s> Storage<'s> for ElasticsearchStorage {
                 .map_err(|err| StorageError::ForceMergeError {
                     source: Box::new(err),
                 })?;
+            // https://www.elastic.co/guide/en/elasticsearch/reference/7.14/breaking-changes-7.0.html#flush-force-merge-no-longer-refresh
+            self.refresh_index(index.name.clone())
+                .await
+                .map_err(|err| StorageError::IndexPublicationError {
+                    source: Box::new(err),
+                })?;
```
remi-dupre, line 191: Note that if `elasticsearch.force_merge.allow_timeout` is set to true, the `force_merge` call above might return *before* the operation actually ends in elasticsearch. This is the default behavior which helps with not having to keep connection alive when the operation takes several hours.

remi-dupre (commented):
Did you have issues with this behavior ? What I understand from the documentation is that the `refresh` ensures that all newly indexed documents are exposed,. As performing a force_merge doesn't create new document I'm surprised that it is required to perform a refresh for a second time :thinking:

libs/mimir/src/adapters/secondary/elasticsearch/storage.rs
woshilapin: OK, I'll try to make a comment to explain it then.

libs/mimir/src/adapters/secondary/elasticsearch/storage.rs
woshilapin: You are totally right, I completely missed that. The problem seems to be that it is then not possible to check for completion before launching the `refresh` in version 7 (see [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/7.14/indices-forcemerge.html)). It seems possible to do it in version 8 thought (see [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-forcemerge.html)).

> ### Running force merge asynchronously
> 
> If the request contains wait_for_completion=false, Elasticsearch performs some preflight checks, launches the request, and returns a [task](https://www.elastic.co/guide/en/elasticsearch/reference/current/tasks.html) you can use to get the status of the task. However, you can not cancel this task as the force merge task is not cancelable. Elasticsearch creates a record of this task as a document at _tasks/<task_id>. When you are done with a task, you should delete the task document so Elasticsearch can reclaim the space.

I've noticed that the `_task` API was actually [in beta for version 7](https://www.elastic.co/guide/en/elasticsearch/reference/7.14/tasks.html), I'm wondering if there is still a possibility to use it in some way :shrug:

woshilapin:
@remi-dupre I completely rewrote the PR based on the following information.

`force_merge` is part of preparing the index, not publishing it. Therefore, I moved the `force_merge` from `publish_index` to `insert_documents`.

Before Elastic Search version 7, `refresh` was part of `force_merge`. After Elastic Search version 8, there is the API `_tasks` to asynchronously check that the `force_merge` is finished. But on version 7, the `refresh` is not automatically done with `force_merge` and I did not find a mechanism to detect the end of `force_merge`.

Moving `force_merge` before `refresh` should be relatively ok, especially since we do have a configurable `timeout` that might help us to let enough time for the `force_merge` to resume.

azime (approved):

libs/mimir/src/adapters/secondary/elasticsearch/storage.rs
```diff
@@ -84,12 +84,32 @@ impl<'s> Storage<'s> for ElasticsearchStorage {
...
             .map_err(|err| StorageError::DocumentInsertionError {
                 source: Box::new(err),
-            })
+            })?;
+
+        if self.config.force_merge.enabled {
+            // A `force_merge` needs an explicit `refresh` since ElastiSearch 7
+            // https://www.elastic.co/guide/en/elasticsearch/reference/7.14/breaking-changes-7.0.html#flush-force-merge-no-longer-refresh
+            // `refresh` will be executed during `publish_index`.
+            // WARN: `force_merge` is interrupted after a timeout
```
pbench, line 99: From what I understand from [the docs](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/indices-forcemerge.html), you cannot specify a timeout for ElasticSearch for this operation.
The timeout in https://github.com/hove-io/mimirsbrunn/blob/d2c7ce06619e0ae28a9616d8ea3f71b18c15da84/libs/mimir/src/adapters/secondary/elasticsearch/internal.rs#L845 is just a timeout on the http request. 
So if we obtained a timeout in mimir, we just known that the force_merge has not finished during the allowed time. But I think the force_merge keeps running on ElasticSearch, even if the timeout occurs.

remi-dupre (approved):

pbench (approved):

woshilapin:
I added some more logs in the [ed34414](https://github.com/hove-io/mimirsbrunn/pull/757/commits/ed3441402b39c1b3947010afc211a95c7841cf37).

PR #758, opened 2022-06-13
feat: add 'curl in bragi's docker image

woshilapin:
In order to be able to do some `healthcheck` inside Docker, we need some way to query the `/status` API from inside the docker image. Note that this solution is not ideal since `curl` is not necessary for `bragi` and therefore, it adds a dependency inside that image (and therefore, possibly some attack surface). [But one of the alternative is not necessarily easier](https://blog.sixeyed.com/docker-healthchecks-why-not-to-use-curl-or-iwr/).

ref: NAV-1358


azime (approved):

pbench:
You don't define the HEALTHCHECK command in the Dockerfile ?

PR #760, opened 2022-06-16
internals: avoid crash when ES bulk returns NoOp status

remi-dupre:
On some corner cases we can get the following error while updating documents:

```
thread 'main' panicked at 'Error inserting chunk: JSON Elasticsearch Deserialization Error: error decoding response body: unknown variant `noop`, expected one of `created`, `updated
```

### DIFF
--- a/libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
+++ b/libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
@@ -624,6 +624,7 @@ impl ElasticsearchStorage {
                 .json()
                 .await
                 .context(ElasticsearchDeserializationSnafu)?;
+
             es_response.items.into_iter().try_for_each(|item| {
                 let inner = item.inner();
                 let result = inner.result.map_err(|err| {
@@ -641,7 +642,10 @@ impl ElasticsearchStorage {
                 match result {
                     ElasticsearchBulkResult::Created => stats.created += 1,
                     ElasticsearchBulkResult::Updated => stats.updated += 1,
-                    _ => unreachable!("no port implements document deletion"),
+                    ElasticsearchBulkResult::NoOp => stats.skipped += 1,
+                    ElasticsearchBulkResult::Deleted => {
+                        unreachable!("no port implements document deletion")
+                    }
                 }
 
                 Ok::<_, Error>(())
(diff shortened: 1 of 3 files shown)


woshilapin (approved):

libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
```diff
@@ -641,7 +642,10 @@ impl ElasticsearchStorage {
                 match result {
                     ElasticsearchBulkResult::Created => stats.created += 1,
                     ElasticsearchBulkResult::Updated => stats.updated += 1,
-                    _ => unreachable!("no port implements document deletion"),
+                    ElasticsearchBulkResult::NoOp => stats.skipped += 1,
+                    ElasticsearchBulkResult::Deleted => {
+                        unreachable!("no port implements document deletion")
```
pbench, line 647: Maybe return an error here instead of panicking ?

remi-dupre: While I was on it I just added a counter for deletion too, I should always be zero but it is not worth keeping a potential point of failure for such a tiny check.

woshilapin (commented):
I'm not sure I understand why you added the `deleted` field since it doesn't seem to be incremented :thinking:

remi-dupre:
> I'm not sure I understand why you added the `deleted` field since it doesn't seem to be incremented thinking

I'm trying to handle any possible output from elasticsearch, deleted *should* always be 0 but ES is sometimes surprising, for example performing updates during insertions (I suppose when there are two documents with the same ID :shrug:)

Would you want me to remove this deleted field and ignoring it from ES's result? (maybe just show a warning)

libs/mimir/src/adapters/secondary/elasticsearch/internal.rs
pbench, line 647: I think @woshilapin was expecting something like : 
```suggestion
                    ElasticsearchBulkResult::Deleted => {
                       stats.deleted +=1
```
Since this should not happen, I would also log a warning when it does

remi-dupre: Oh crap I was sure I did perform this increment, that's why I didn't understand the command :confused:

remi-dupre: That should be fixed [here](https://github.com/hove-io/mimirsbrunn/pull/760/commits/594176c04b60364cf351071a19ded98dbbefdd5d), it really didn't make any sense without it, sorry :sweat_smile:

pbench (approved):

PR #762, opened 2022-06-16
Update docker CI to be compatible with forks

remi-dupre:
We are currently experimenting working on a fork of mimirsbrunn, which we would keep up to date with mimirsbrunn but in which we could operate more freely (eg. merging before you had time to review in order to deploy).

This PR proposes a change to make the CI work on several repos:
 -  replacing the hardcoded "navitia" org with a secret value (ie. you would need to add DOCKER_ORG=navitia in the secrets before merging this PR)
 - push docker images for branches and use the branch name as label, just for convenience

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -11,7 +11,6 @@ on:
 
 env:
   DOCKER_BUILDKIT: 1
-  DOCKER_ORG: navitia
 
 jobs:
   build:
@@ -41,7 +40,6 @@ jobs:
           echo "${{ steps.rust_version.outputs.rustupToolchain }}"
 
       - name: Login to DockerHub
-        if: github.event_name == 'push'
         uses: docker/login-action@v1
         with:
           username: ${{ secrets.DOCKER_USER }}
@@ -51,13 +49,13 @@ jobs:
         run: |
           VERSION=${GITHUB_REF#refs/*/}
           if [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then
-            IMAGE_TAG=$GITHUB_SHA
+            IMAGE_TAG=$GITHUB_HEAD_REF
           elif [ "$VERSION" == "master" ]; then
             IMAGE_TAG=latest
           else
             IMAGE_TAG=$VERSION
           fi
-          echo "DOCKER_IMAGE=$DOCKER_ORG/${{ matrix['image'] }}:$IMAGE_TAG" >> $GITHUB_ENV
+          echo "DOCKER_IMAGE=${{ secrets.DOCKER_ORG }}/${{ matrix['image'] }}:$IMAGE_TAG" >> $GITHUB_ENV
 
       - name: Build docker image
         run: >-
@@ -68,5 +66,4 @@ jobs:
           --tag $DOCKER_IMAGE
           .
 
-      - if: github.event_name == 'push'
-        run: docker push $DOCKER_IMAGE
+      - run: docker push $DOCKER_IMAGE
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -31,6 +31,7 @@ jobs:
         if: "${{ env.SONAR_TOKEN != '' }}"
         run: cargo-sonar --features clippy --clippy-path clippy.json
       - name: Run sonar-scanner
+        uses: SonarSource/sonarcloud-github-action@master
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -26,14 +26,15 @@ jobs:
       - name: Install 'cargo-sonar'
         run: cargo install cargo-sonar --version 0.9.0 --locked
       - name: Convert into Sonar compatible format
-        if: ${{ env.DOCKER_ORG == 'navitia' }}
+        env:
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+        if: "${{ env.SONAR_TOKEN != '' }}"
         run: cargo-sonar --features clippy --clippy-path clippy.json
       - name: Run sonar-scanner
-        if: ${{ env.DOCKER_ORG == 'navitia' }}
-        uses: SonarSource/sonarcloud-github-action@master
         env:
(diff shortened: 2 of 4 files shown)


.github/workflows/docker.yml
```diff
@@ -51,13 +49,13 @@ jobs:
         run: |
           VERSION=${GITHUB_REF#refs/*/}
           if [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then
-            IMAGE_TAG=$GITHUB_SHA
+            IMAGE_TAG=$GITHUB_HEAD_REF
```
woshilapin, line 52: I'm not sure this change is a good idea. The SHA will help to know exactly what version of the code to debug. `HEAD_REF` will not since you don't know which version of the branch was used. Is there a reason you made this change?

sonarcloud.yml
```diff
@@ -11,7 +11,7 @@ env:
 
 jobs:
   clippy:
-    if: ${{ github.event_name == 'push' || (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository && github.actor != 'dependabot[bot]' }}
+    if: ${{ github.event_name == 'push' || (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository && github.actor != 'dependabot[bot]') }}
```
woshilapin, line 14: oops :sweat_smile: Thank you for the fix!

sonarcloud.yml
```diff
@@ -26,8 +26,10 @@ jobs:
       - name: Install 'cargo-sonar'
         run: cargo install cargo-sonar --version 0.9.0 --locked
       - name: Convert into Sonar compatible format
+        if: ${{ env.DOCKER_ORG == 'navitia' }}
```
woshilapin, line 29: That will help contribution indeed. However, I'm wondering if this would not be more correct to use the following condition (see list of [environment variables](https://docs.github.com/en/actions/learn-github-actions/environment-variables)).

```suggestion
        if: ${{ env.GITHUB_REPOSITORY_OWNER == 'hove-io' }}
```

.github/workflows/docker.yml
remi-dupre: Would it help if we put this reference as a label ? ` --label "org.label-schema.vcs-ref=$GITHUB_SHA"`?

To me using the commit ref as the image tag is not very convenient as the branch name is usually more expressive of what the image/branch is meaning to solve.

sonarcloud.yml
remi-dupre: That seems better indeed :+1: 

While we are discussion on it I was wondering if could be a good idea to use a `DISABLE_SONARCLOUD` env variable instead, it would make it even less specific to hove-io as any fork could choose to opt-out without any change.

.github/workflows/docker.yml
woshilapin: At first, I thought you could just do something like this so you get both.

```suggestion
             IMAGE_TAG=${GITHUB_HEAD_REF}_${GITHUB_SHA}
```

I didn't know about labels. Seems good too, I'm just worry people won't know about it and therefore, won't find it. So if you choose this solution, can you document it somewhere in the developer's documentation?

sonarcloud.yml
woshilapin: I would go the other way around. Let's make `hove-io/mimirsbrunn` have a `ENABLE_SONARCLOUD` so it's less a bother for contributors that might not know about it, what do you think? I feel the default behavior for a new contributor should just work with minimal setup.

woshilapin: Thinking about it, we might just need to test for the presence of `SONAR_TOKEN`?

remi-dupre: That's a great idea, just did that :+1:

.github/workflows/docker.yml
remi-dupre: I've added some documentation here, never too inspired about this but I think it was a shame we never mentioned there was some docker images somewhere :stuck_out_tongue:  https://github.com/hove-io/mimirsbrunn/pull/762/commits/56861baca86df1d38195bcbc72a6b1322e0d332c

README.md
```diff
@@ -68,10 +68,28 @@ For running end to end or unit tests, you need the docker engine available.
...
+You can also fetch official images from DockerHub here:
+
+ - [navitia/mimirsbrunn:latest](https://hub.docker.com/r/navitia/mimirsbrunn)
+ - [navitia/bragi:latest](https://hub.docker.com/r/navitia/bragi)
+
+
+If you want to get the commit ref used to build the latest image you can run the following command:
+
+```bash
+docker inspect --format='{{index .Config.Labels "org.label-schema.vcs-ref"}}' qwantresearch/mimirsbrunn:latest
```
woshilapin, line 87:
```suggestion
docker inspect --format='{{index .Config.Labels "org.label-schema.vcs-ref"}}' navitia/mimirsbrunn:latest
```

.github/workflows/sonarcloud.yml
```diff
@@ -26,12 +26,15 @@ jobs:
       - name: Install 'cargo-sonar'
         run: cargo install cargo-sonar --version 0.9.0 --locked
       - name: Convert into Sonar compatible format
+        env:
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+        if: "${{ env.SONAR_TOKEN != '' }}"
         run: cargo-sonar --features clippy --clippy-path clippy.json
       - name: Run sonar-scanner
-        uses: SonarSource/sonarcloud-github-action@master
```
woshilapin, line 31: Are you sure about removing this line?

remi-dupre: That's a mistake :+1:

pbench:
Looks good to me, thanks !

The docker worflows are failing on this PR, but I think it is because the PR is from the `Qwant` fork, which does not have access to the secrets DOCKER_USER and DOCKER_PASSWORD.

@remi-dupre maybe if you add these secrets to your fork, the CI will turn green again for this workflow ?

I am merging anyway.

remi-dupre:
> Looks good to me, thanks !
> 
> The docker worflows are failing on this PR, but I think it is because the PR is from the `Qwant` fork, which does not have access to the secrets DOCKER_USER and DOCKER_PASSWORD.
> 
> @remi-dupre maybe if you add these secrets to your fork, the CI will turn green again for this workflow ?
> 
> I am merging anyway.

Well it managed to login on both repos, maybe it was using the configuration from master :thinking: 

Thanks anyway :clap:

PR #765, opened 2022-06-17
Update all deps

remi-dupre:
Updates most dependencies with the help of `cargo update` and `cargo outdated`. The only remaining outdated direct dependency is cucumber, which would probably need more work.

:warning: There is a bit of breaking changes:

 - `MIMIR_` and `BRAGI_` env prefixes are now `MIMIR__` and `BRAGI__`, I don't think there is a way to tweak that :/ (I hate the *config* crate)
 - newer cosmogony version **needs** a newly generated cosmogony file, with the `loc_name` and `alt_name` fields, see https://github.com/osm-without-borders/cosmogony/pull/137


libs/common/src/config.rs
```diff
@@ -53,41 +54,38 @@ pub fn config_from<
...
-            Ok(config)
-        })
-        .context(ConfigCompilationSnafu)?;
+            builder.add_source(File::from(local_path).required(false))
+        });
 
     // Add in settings from the environment
     // Eg.. `<prefix>_DEBUG=1 ./target/app` would set the `debug` key
     if let Some(prefix) = prefix.into() {
         let prefix = Environment::with_prefix(prefix).separator("__");
```
woshilapin, line 70: What about changing this to keep compatibility about the environment variables? (I actually don't really understand why we do have a double-underscore here, maybe you can enlighten me?).

src/bin/cosmogony2mimir.rs
```diff
@@ -400,7 +400,7 @@ mod tests {
         let brittany = admins.iter().find(|a| a.name == "Bretagne").unwrap();
         assert_eq!(brittany.id, "admin:osm:relation:102740");
         assert_eq!(brittany.zone_type, Some(cosmogony::ZoneType::State));
-        assert_relative_eq!(brittany.weight, 0.002_298, epsilon = 1e-6);
+        assert_relative_eq!(brittany.weight, 0.002_396, epsilon = 1e-6);
```
woshilapin, line 403: Do you know the reason of this change, I can't understand it.

woshilapin (commented):
Thank you for the work.

libs/common/src/config.rs
remi-dupre: The double underscore is here to remove the ambiguity with snake case keys, for example `ELASTICSEARCH_FORCE_MERGE` would be interpreted as `elasticsearch.force.merge` instead of `elasticsearch.force_merge`.

It would be great if `config` was aware of the target configuration and would be able to deal with this kind of ambiguity, I have to concede that it isn't the use-case it tries to solve :/

src/bin/cosmogony2mimir.rs
remi-dupre: I re-built the whole file instead of just adding the missing keys (mostly for convenience), this weight is proportional to the population of Bretagne, which must have been updated since last time the dataset was built.

woshilapin:
I'm starting to wonder if the `config` update should be contained in its own PR since it will create some impact on the deployment (environment variables). We might want to have more control on when we want to merge it and independently of the other updates. What do you think @azime?

fatal69100:
@woshilapin We can close this PR because we split it into two ohter PRs to be easier to merge :  https://github.com/hove-io/mimirsbrunn/pull/778 and https://github.com/hove-io/mimirsbrunn/pull/779.
By the way, we can't create branch on this repository and I think these PRs are not based on each other's (even with https://github.com/hove-io/mimirsbrunn/pull/768) that make them not readable.

woshilapin:
Yes, I saw that your PR are based on `simplify bragi`. I'm checking if we can merge it, that should makes it easier for #778 and #779.

PR #768, opened 2022-06-20
Simplify bragi

remi-dupre:
The main goal here is to make Bragi more maintainable, mostly through a
more straightforward use of Warp that does not require to maintain several
macros per endpoint.

I've also changed the `SearchDocuments` trait so that it doesn't enforce
a return type for the JSON result. This resolves several cases where we
had to convert to a `Value` and then convert to another type (which
neither performant nor optimal).

My main concern about this PR is that it removes a big chunk of unit tests in
*bragi/handlers.rs* and *bragi/routes.rs*. I'm not sure how to get them back
without making the code complicated again but I would argue that most of
these tests were validating the behavior of external libs (eg. param parsing
from warp) anyway.


libs/mimir/src/adapters/primary/bragi/api.rs
```diff
@@ -83,6 +97,23 @@ impl From<ForwardGeocoderExplainQuery> for ForwardGeocoderQuery {
     }
 }
 
+impl Validate for ForwardGeocoderExplainQuery {
+    fn filter(&self) -> Result<(), warp::Rejection> {
+        ensure! {
+            !self.q.is_empty();
+
+            self.lat.is_some() == self.lon.is_some(),
```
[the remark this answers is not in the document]
remi-dupre: This is intended (the position must not be provided or fully provided), but your comment highlights that it deserves a comment, or at least that the error message bellow is not clear enough).

src/bragi/server.rs
```diff
@@ -71,43 +70,79 @@ pub async fn run_server(settings: Settings) -> Result<(), Error> {
...
+            .context(ElasticsearchConnectionSnafu)?;
+
+        let settings = settings.clone();
+        let ctx = Arc::new(handlers::Context { client, settings });
+
+        move || {
+            let ctx = ctx.clone();
+            move || ctx.clone()
+        }
+    };
```
pbench, line 88: I am surprised you need to put the client inside and `Arc` because : 
- it was not needed before your PR
- the client struct satisfy the[ ConnectionPool trait](https://docs.rs/elasticsearch/7.14.0-alpha.1/elasticsearch/http/transport/trait.ConnectionPool.html) and thus is Sync and can safely be shared between threads
- the Settings struct also seems to be Sync

libs/mimir/src/adapters/primary/bragi/api.rs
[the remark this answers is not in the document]
pbench: Ok ! yes, I did not understood that having no lat and no lon was allowed.
I can suggest for the message
```
        "lat and lon parameters must either be both present or both absent
```

remi-dupre:
That's really weird that this docker login works on the master branch of both repository but doesn't when opening a PR :disappointed:

remi-dupre:
> That's really weird that this docker login works on the master branch of both repository but doesn't when opening a PR disappointed

Nah, got it! :)

> Note: With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

From https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow

remi-dupre:
So what I've done is I changed the CI to trigger a build for pushes to any branch, which will be triggered from the fork and will allow access to its secrets.

On the other hand it is still interesting to see if the build succeeds in the PR, so I've kept PR builds but disabled dockerhub upload in that context.

libs/mimir/src/adapters/primary/bragi/routes.rs
```diff
@@ -158,181 +33,100 @@ pub enum InvalidRequestReason {
...
+    fn filter(&self) -> Result<(), Rejection> {
+        Ok(())
+    }
+}
+
+/// Extract and validate input parameter from the query
+pub fn validate_query<T>() -> impl Filter<Extract = (T,), Error = Rejection> + Copy
+where
+    T: DeserializeOwned + Validate + Send + Sync,
+{
```
pbench, line 61: Since this PR is about simplifying the code: do you really need to introduce yet another trait here ?

This trait is used only for the `validate_query<T>()` function, and implemented for the 3 structs `ReverseGeocoderQuery`, `ForwardGeocoderQuery` and `ForwardGeocoderExplainQuery`.

I think it would be simpler to have 3 functions 
`validate_reverse_query()`, `validate_forward_query()` and `validate_forward_explain_query()`
and call them explicitely.

It would make more explicit what happens in the route definition.

For now, it is really hard to understand that [the](https://github.com/hove-io/mimirsbrunn/blob/c4d93ef4eab0256ebec096b0c2f3ffc815d15626/src/bragi/server.rs#L94) [three](https://github.com/hove-io/mimirsbrunn/blob/c4d93ef4eab0256ebec096b0c2f3ffc815d15626/src/bragi/server.rs#L110) [calls](https://github.com/hove-io/mimirsbrunn/blob/c4d93ef4eab0256ebec096b0c2f3ffc815d15626/src/bragi/server.rs#L117) to validate_query() does three different things

libs/mimir/src/adapters/primary/bragi/routes.rs
```diff
@@ -407,233 +201,3 @@ where
...
-            .path("/api/v1/autocomplete")
-            .filter(&filter)
-            .await;
-        assert!(
-            resp.unwrap_err()
-                .find::<warp::reject::InvalidQuery>()
-                .is_some(),
-            "Empty query parameter not allowed"
-        );
-    }
```
pbench, line 429: I agree that these tests are not the most useful, but they do provide examples of api calls, which I think is a good way to understand how the code is supposed to work.

So I would prefer if we could keep them somewhere.

I managed to add a test on this branch by adding the following to src/bragi/server.rs

```
#[cfg(test)]
mod tests {
    use super::*;
    use mimir::adapters::primary::bragi::routes::validate_query;
    #[tokio::test]
    async fn should_report_invalid_query_with_no_query() {
        let filter = validate_query::<ForwardGeocoderQuery>();
        let resp = warp::test::request()
            .path("/api/v1/autocomplete")
            .filter(&filter)
            .await;
        assert!(
            resp.unwrap_err()
                .find::<warp::reject::InvalidQuery>()
                .is_some(),
            "Empty query parameter not allowed"
        );
    }

}
```

libs/mimir/src/adapters/primary/bragi/handlers.rs
```diff
@@ -54,6 +58,38 @@ lazy_static::lazy_static! {
...
+    #[serde(deserialize_with = "deserialize_duration")]
+    pub autocomplete_timeout: Duration,
+    #[serde(deserialize_with = "deserialize_duration")]
+    pub reverse_timeout: Duration,
+    #[serde(deserialize_with = "deserialize_duration")]
+    pub features_timeout: Duration,
+}
+
+#[derive(Clone)]
+pub struct Context<C> {
```
pbench, line 88: Do we ever use something else than  `ElasticsearcStorage` for `C` ?

If not, we may as well remove the genericity here (and all the traits everywhere).

This may be a big change (?), so it can be postponed to another PR.

remi-dupre: Yeah I agree that all this genericity is overkill and makes the code more complicated to maintain / creates pairs of traits for which it is hard of handle the role. This came from an effort from Matthieu to push for an hexagonal architecture which required to expose as little "implementation-specific" types as possible but I have to say I am not a huge fan of how it operates in Rust and for a rather little project :/

However I also think that it may required a big diff to remove this pattern so it would be better to put it in another PR.

pbench: Allright, let's keep this in mind for a future PR.

libs/mimir/src/adapters/primary/bragi/api.rs
```diff
@@ -83,6 +97,23 @@ impl From<ForwardGeocoderExplainQuery> for ForwardGeocoderQuery {
     }
 }
 
+impl Validate for ForwardGeocoderExplainQuery {
+    fn filter(&self) -> Result<(), warp::Rejection> {
+        ensure! {
```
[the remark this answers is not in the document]
remi-dupre: Indeed, plus it's only used twice

libs/mimir/src/adapters/primary/bragi/routes.rs
remi-dupre: Yeah you may be right: it is already hard to understand how warp works without hiding many types, so the more hints we leave the better.

I've done something in the same fashion as your suggestion, the three functions were added as methods of the constructed structs. I think it makes it even clearer that we build a `ReverseGeocoderQuery` when calling `ReverseGeocoderQuery::validate`.

pbench: very nice ! thanks

libs/mimir/src/adapters/primary/bragi/routes.rs
remi-dupre: I've added all tests back :+1:

pbench: thanks !

pbench (approved):
Looks good to me, thanks a lot for the hard work ! 
Since this is quite a big refacto, I'll wait for a review by another pair of eyes before merging

libs/mimir/src/domain/ports/secondary/search.rs
```diff
@@ -7,6 +7,9 @@ use snafu::Snafu;
 
 use crate::domain::model::query::Query;
 
+// TODO: this trait seems like bloat: it is the exact same interface as primary port
+//       search_documents
```
woshilapin, line 11: Probably a good reason to duplicate from the hexagonal architecture point of view... 

```
                                   _______________
                                  /               \
                                /                   \
                              /                       \
External (probably HTTP) ---------> Business Logic ---------> External (probably ES)
                         Parameters                Parameters
                         [primary]                 [secondary]
                              \                       /
                                \                   /
                                  \               /
                                   _______________
```
so they are two different interface semantically.

But as you said, for such a small project, hexagonal architecture is probably over-architectured, especially when most of the patterns above are essentially just forwarding things... explaining the similarity between the `trait`.

src/bragi/settings.rs
```diff
@@ -106,21 +74,19 @@ pub enum Command {
...
-            "BRAGI",
-            opts.settings.clone(),
-        )
-        .context(ConfigCompilationSnafu)?
-        .try_into()
-        .context(ConfigMergeSnafu {
-            msg: "cannot merge bragi settings",
-        })
-    }
+pub fn build_settings(opts: &Opts) -> Result<Settings, Error> {
```
woshilapin, line 77: Just feels like having this function as some sort of constructor of `Settings` would makes it clearer. Since it's faillible, maybe you can implement `impl TryFrom<&Opts> for Settings` instead of having a free function?

libs/mimir/src/adapters/primary/bragi/routes.rs
```diff
@@ -209,130 +73,36 @@ pub fn forward_geocoder_explain_query(
...
-        Err(warp::reject::custom(InvalidRequest {
-            reason: InvalidRequestReason::InconsistentZoneRequest,
-            info: "'zone_type' must be specified when you query with 'type' parameter 'zone'"
-                .to_string(),
-        }))
-    } else {
-        Ok(params)
-    }
+            .map(|zone_types| !zone_types.is_empty())
+            .unwrap_or(false)
```
woshilapin, line 88: Just for readibility, how about extracting both part of the `<a> || <b>` into variables? Because, weirdly enough, having `||` at the start of the line, it took me a minute to realize it was not a closure :sweat_smile:

woshilapin (approved):
:clap: :clap: :clap: Much less code, much clearer: thank you!

remi-dupre:
Thanks for the review, these should be fixed now :)

woshilapin (approved):

woshilapin:
This PR is being reverted in #801 because it does remove important endpoints (that's a shame that tests didn't catch that, schedule actually didn't allow us to realize that before). We might reintroduce these simplifications later.

PR #769, opened 2022-06-20
add search endpoint and alt names

remi-dupre:
:warning: this PR is based on https://github.com/hove-io/mimirsbrunn/pull/768, we should review and merge it first.

There is two thing in this PR and it seems more complicated to take them appart :/

### Search endpoint

We added a */search* endpoint, which doesn't basically the same thing as */autocomplete* without matching the prefix mappings. In our case it is used to get more relevant results out of the scope of our auto-completion module, for example when looking up the result for a Qwant request (which is usually a "finished" query) in https://www.qwant.com/?q=paris&t=web.

### Alt names

This also adds mappings for alternative and local names. A sadly know relevant example which is addressed with this change is french people looking for "Kiev" instead of "Kyïv" (the former name being specified as alt_name [in OSM](https://www.openstreetmap.org/relation/421866)).

Also see https://github.com/osm-without-borders/cosmogony/pull/144


fatal69100:
Outdated, can be closed by https://github.com/hove-io/mimirsbrunn/pull/780

PR #771, opened 2022-06-27
[ntfs2mimir] attach admins from file

pbench:
Ntfs2mimir needs data about the administrative regions to enrich the stops before putting them into elasticsearch.
The data about the administrative regions  was collected from elasticsearch. 
This collection method causes a non-negligible load on elasticsearch, resulting in various timeouts and errors in bragi.

This PR adds the option to collect the data about administrative region from a local cosmogony file (the same file used to build the "admin" index in cosmogony2mimir).

If a "cosmogony_file" is configured in ntfs2mimir config file, then this file will be used. 
If such configuration is not present, ntfs2mimir will gather administrative regions from elasticsearch, as before.

The same thing is done for the binaries osm2mimir, openaddresses2mimir and poi2mimir.

For more details: https://navitia.atlassian.net/browse/NAV-1339

### DIFF
--- a/config/bano2mimir/default.toml
+++ b/config/bano2mimir/default.toml
@@ -1,14 +1,15 @@
 nb_threads = 2
 update_templates = true
 
-# If a cosmogony_file is configured, ntfs2mimir will read the admins from this file.
-# If such a file is not configured, ntfs2mimir will read the admins from elasticsearch
 
+
+# If the admins sections is present, the admins will be read in the cosmogony file.
+# Otherwise, admins will be fetched  from elasticsearch
+# [admins]
 # cosmogony_file = "/home/pascal/mimirsbrunn/cosmogony-europe.jsonl.gz"
-
 # configuration for reading the cosmogony file
-french_id_retrocompatibility = true
-langs = [ "fr" ]
+# french_id_retrocompatibility = true
+# langs = [ "fr" ]
 
 [container]
   name = "addr"
--- a/config/bano2mimir/default.toml
+++ b/config/bano2mimir/default.toml
@@ -1,6 +1,15 @@
 nb_threads = 2
 update_templates = true
 
+# If a cosmogony_file is configured, ntfs2mimir will read the admins from this file.
+# If such a file is not configured, ntfs2mimir will read the admins from elasticsearch
+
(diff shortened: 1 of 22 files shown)


woshilapin (approved):

azime (approved):

config/bano2mimir/default.toml
```diff
@@ -1,6 +1,15 @@
...
 update_templates = true
 
+# If a cosmogony_file is configured, ntfs2mimir will read the admins from this file.
+# If such a file is not configured, ntfs2mimir will read the admins from elasticsearch
+
+# cosmogony_file = "/home/pascal/mimirsbrunn/cosmogony-europe.jsonl.gz"
+
+# configuration for reading the cosmogony file
+french_id_retrocompatibility = true
+langs = [ "fr" ]
```
remi-dupre, line 11: Maybe I'm a bit late to the party and it's not very important, but wouldn't it be better to have this part in its own cosmogony-related subsection? (and same for other binaries)

```suggestion
  [cosmogony]
  # If a cosmogony_file is configured, ntfs2mimir will read the admins from this file.
  # If such a file is not configured, ntfs2mimir will read the admins from elasticsearch
  # cosmogony_file = "/home/pascal/mimirsbrunn/cosmogony-europe.jsonl.gz"

  # configuration for reading the cosmogony file
  french_id_retrocompatibility = true
  langs = [ "fr" ]
```

src/pois.rs
```diff
@@ -116,21 +121,34 @@ pub async fn index_pois(
...
+                        .await
+                }
+                Err(err) => {
+                    warn!("administratives regions not found in es db. {:?}", err);
+                    return Err(Error::AdminRetrieval {
+                        details: err.to_string(),
+                    });
+                }
+            }
+        };
```
remi-dupre, line 150: :warning:  That may be a suggest for another PR since the issues extends to other parts of the settings.

This if/else is duplicated in several places, maybe grouping cosmogony settings is a good way to abstract things ot be able to call something like `cosmogony_settings.load_admins()?`.

config/bano2mimir/default.toml
pbench: I've put everything in a [admins] subsection

src/pois.rs
pbench: done

remi-dupre (approved):

PR #775, opened 2022-07-25
Switch to alpine

sinaure:
vulnerabilities improvements switching to alpine


SGrenet:
> No description provided. 
Hello
32 files changes and only a title, it's not really kind for a QA who does not read rust

SGrenet:
and I con continue as long as necessary

PR #776, opened 2022-07-26
feat: use an integer gauge for counting request in flight

woshilapin:
A minima, we should increment the number, not decrement it.

And let's use an integer gauge that is susceptible to get better performance (see [documentation](https://docs.rs/prometheus/latest/prometheus/type.IntGauge.html)).

Also, `Gauge` uses `AtomicF64` that implement increment with [a fallible addition](https://docs.rs/prometheus/0.13.0/src/prometheus/atomic64.rs.html#112).

`IntGauge` uses `AtomicI64` which itself use [`std::sync::atomic::AtomicI64::fetch_add`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicI64.html#method.fetch_add): `fetch_add` does not panic but [wraps on overflow](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicI64.html#method.fetch_add).

### DIFF
--- a/libs/mimir/src/adapters/primary/bragi/prometheus_handler.rs
+++ b/libs/mimir/src/adapters/primary/bragi/prometheus_handler.rs
@@ -51,7 +51,7 @@ lazy_static::lazy_static! {
     )
     .unwrap();
 
-    static ref HTTP_IN_FLIGHT: prometheus::Gauge = prometheus::register_gauge!(
+    static ref HTTP_IN_FLIGHT: prometheus::IntGauge = prometheus::register_int_gauge!(
         "bragi_http_requests_in_flight",
         "current number of http request being served"
     )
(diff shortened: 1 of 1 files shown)


pbench (commented):
Seems like a good idea to use an IntGauge for counting requests :+1:

remi-dupre:
Looks like a good idea indeed :sweat_smile: 

> IntGauge uses AtomicI64 which itself use [std::sync::atomic::AtomicI64::fetch_add](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicI64.html#method.fetch_add): fetch_add does not panic but [wraps on overflow](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicI64.html#method.fetch_add).

Plus I'm pretty sure none of us will ever have to deal with more than $2^{63} \approx 10^{19} = 10^{10}~\text{billion}$ requests in flight :stuck_out_tongue:

remi-dupre (approved):

pbench (approved):

PR #783, opened 2022-10-14
ci: trigger external job when new Docker image is produced

woshilapin:
Correlated with https://github.com/hove-io/bragi-aws-assets/pull/77 which should be merged before.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -65,6 +65,17 @@ jobs:
         with:
           app_id: ${{ secrets.GA_OS_WORKFLOW_TRIGGER_APP_ID }}
           private_key: ${{ secrets.GA_OS_WORKFLOW_TRIGGER_APP_PEM }}
+      
+      - name: Init variables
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -57,3 +57,31 @@ jobs:
...
+      - name: Generate token
+        if: github.event_name == 'push'
+        id: app-token
+        uses: getsentry/action-github-app-token@v1
+        with:
+          app_id: ${{ secrets.GA_OS_WORKFLOW_TRIGGER_APP_ID }}
+          private_key: ${{ secrets.GA_OS_WORKFLOW_TRIGGER_APP_PEM }}
+      
+      - name: Init variables
+        run: |
```
prhod, line 70: Maybe you can also add 
```suggestion
if: github.event_name == 'push'
```

woshilapin: Done in 1f714462.

PR #800, opened 2023-03-09
feat: add possibility to deactivate ANSI formatting for logs

woshilapin:
Will give the ability to deactivate ANSI formatting (give some coloring and bold for example) when necessary (see screenshot below with all the `[2m` and other `[32m`).

image

When `NO_COLOR` variable is defined (whatever the value you put in it), then ANSI will be deactivated and it's just going to be raw text.

Note: also modifying level of log for some errors to `WARN` (they are not unrecoverable errors from `bragi` point of view so `WARN` is enough).

### DIFF
--- a/src/utils/logger.rs
+++ b/src/utils/logger.rs
@@ -27,6 +27,7 @@ pub fn logger_init() -> Result<tracing_appender::non_blocking::WorkerGuard, Erro
     let (non_blocking, guard) = tracing_appender::non_blocking(std::io::stdout());
 
     let event_format = tracing_subscriber::fmt::format()
+        .with_ansi(!std::option_env!("NO_COLOR").is_some())
         .with_target(false) // Don't include event targets.
         .compact();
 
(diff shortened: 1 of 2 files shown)


pbench (approved):

src/utils/logger.rs
```diff
@@ -27,6 +27,7 @@ pub fn logger_init() -> Result<tracing_appender::non_blocking::WorkerGuard, Erro
     let (non_blocking, guard) = tracing_appender::non_blocking(std::io::stdout());
 
     let event_format = tracing_subscriber::fmt::format()
+        .with_ansi(!std::option_env!("NO_COLOR").is_some())
```
pbench, line 30: from sonalcloud
```suggestion
        .with_ansi(std::option_env!("NO_COLOR").is_none())
```

woshilapin: :facepalm: Obviously better. I also changed the name of the variable to `NO_ANSI` since it's not only about coloring, but about formatting in general (bold and italic, for example).

pbench (approved):

PR #801, opened 2023-03-13
Revert "Simplify bragi #768"

woshilapin:
We're reverting #768 which happened to remove 2 of the endpoints of `bragi` (`/features` and `/autocomplete-explain`). The simplification in #768 is actually worth reconsidering later, but time doesn't allow addressing that right now.

Note 3 commits were not reverted which all concerns CI (ca27e648213d2d2f6281007dc720e980a681087d, 29c1baf874b098e8b11b12d6b16c69ccc35daea4, and c4d93ef4eab0256ebec096b0c2f3ffc815d15626). The actual CI is working pretty well and those commits were an improvement, so let's keep them.

Another modification has not been reverted: we kept the upgrade to Rust `1.60.0`.


rust-toolchain
```diff
@@ -1 +1 @@
-1.60
+1.56
```
pbench, line 1: I think this is the cause of your CI fail.
I guess bumping to a newer version will fix it
```suggestion
1.68
```

pbench (approved):

PR #803, opened 2023-03-15
tests: clean-up unused or broken tests

woshilapin:
I'm removing all the tests, planning to add more tests in another form and in another PR. In any case, I don't really remove tests since they were failing anyway.

Note however, I'm having errors on my local machine running the tests, notably on tests for `cosmogony2mimir` binary. It's randomly failing on one of the test with the following error

```
---- tests::should_correctly_index_a_small_cosmogony_file stdout ----
thread 'tests::should_correctly_index_a_small_cosmogony_file' panicked at 'Error inserting chunk: Elasticsearch Response: Not Created: Object id ad
min:osm:relation:7398, Error: index [munin_admin_fr_20230315_132709_498930314] blocked by: [TOO_MANY_REQUESTS/12/disk usage exceeded flood-stage wa
termark, index has read-only-allow-delete block];, ', /home/woshilapin/projects/hove/mimirsbrunn/libs/mimir/src/adapters/secondary/elasticsearch/in
ternal.rs:593:47
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I tried increasing memory of the docker container for ElasticSearch in `config/docker/default.toml` without any success. I also tried to reduce concurrency when executing tests `cargo test --jobs 1 --bin cosmogony2mimir`, but not better. Can any one of you try (I'd say more than once ideally since it's a bit random... but at least one failed 100% of the time for me) on his machine and say if they can reproduce?


pbench:
`cargo test --bin cosmogony2mimir` worked fine for me twice in a row :thinking:

pbench (approved):

azime (approved):

PR #807, opened 2023-03-24
feat: parse addresses without async

woshilapin:


pbench (approved):
that's really cleaner ! :sparkles:

azime (approved):

